PR #1, opened 2017-02-01
Writing clean Javascript

li-kai:
As per title, perhaps a little more towards "testable" now but I'll move towards ES6 and patterns in the future.

### DIFF
--- a/contents/javascript/javascript.md
+++ b/contents/javascript/javascript.md
@@ -91,7 +91,7 @@ function bindStudentPhotoLink(selector) {
 }
 ```
 
-Of course, this may be a problem if two Javascript functions have the same name, as they are all in the global scope. This is where the module pattern can be used.
+Of course, this may be a problem if two Javascript functions have the same name, as they are all in the global scope. This is where the module pattern (see: [Namespacing in Javascript]()) can be used.
 
 ```js
 var myApp = (function() {
@@ -195,7 +195,7 @@ createPopUp('Warning', 'This will delete everything!', dangerStatus, { bodyColor
 
 ### Understand method chaining
 
-Method chaining is syntax such as `array.concat([1, 2]).push(1).filter(isEven)`. It is also sometimes referred to as the fluent interface.
+Method chaining is syntax such as `array.concat([1, 2]).push(1).filter(isEven)`. It is also sometimes referred to as the fluent interface. (see: [Method Chaining in JavaScript]())
 
 It is achieved by returning the object itself in the call.
 ```js
@@ -248,4 +248,4 @@ The title says it all. From the most common to obscure patterns, this book cover
 ## References
 [Namespacing in Javascript](https://javascriptweblog.wordpress.com/2010/12/07/namespacing-in-javascript/)
 
-[MMethod Chaining in JavaScript](https://schier.co/blog/2013/11/14/method-chaining-in-javascript.html)
+[Method Chaining in JavaScript](https://schier.co/blog/2013/11/14/method-chaining-in-javascript.html)
--- a/contents/javascript/javascript.md
+++ b/contents/javascript/javascript.md
@@ -112,5 +112,44 @@ var myApp = (function() {
 
 As demonstrated above, only `myApp` is declared in the global scope, and we can access its methods through the object notation, e.g. `myApp.next()`. This is also especially useful to declare private variables that be used among functions. `id` is not accessible outside of scope in this example.
 
+### Purity is worth persuing
+
+There are tons of literature about functional programming. I will heavily recommend reading the [Mostly adequate guide to Functional Programming](https://github.com/MostlyAdequate/mostly-adequate-guide). It explains functional concepts extremely well.
+
+The benefit of pure functions is simple, there is no need to keep track of state. Given an input, the output is guaranteed to be the same every time. This allows us to write extremely simple unit tests, instead of having to maintian the state while testing.
+
+```js
+var counter = 0;
+var todos = [];
+
+function getTodo() {
+    counter++;
+    if (counter < todos.length) {
+        return 'NIL';
+    }
+    return todos[counter];
+}
+...
+getTodo();
+```
+
+The above function `getTodo` is not stateless as it depends on counter's value. In order to write the tests, we would need to ensure counter is reset to the same value at the end of each test. A better way would be to do this:
+
+```js
+var counter = 0;
+var todos = [];
+
+function getTodo(counter, todos) {
+    if (counter < todos.length) {
+        return 'NIL';
+    }
+    return todos[counter + 1];
+}
+...
+getTodo(counter, todos);
+```
+
+Now, not only that the person who writes the unit test can write in fewer lines of code, you can also use the function for some other state other than the global values.
+
 ## References
 [Namespacing in Javascript](https://javascriptweblog.wordpress.com/2010/12/07/namespacing-in-javascript/)
--- a/contents/javascript/javascript.md
+++ b/contents/javascript/javascript.md
@@ -20,7 +20,7 @@ function populateStatusMessageDiv(message, status) {
 }
 ```
 
-In this piece of code we see that the selector is intrinsically tied to the function. This means that the function is tightly tied with the selector and is not easily testable as the test code has to generate the same markup in the test suite for the function to hook up to. In order to prevent such tight coupling, it is advised to leave the selector as a parameter, or sometimes, simply pass in the element itself.
+In this piece of code, we see that the selector is intrinsically tied to the function. This means that the function is tightly tied with the selector and is not easily testable as the test code has to generate the same markup in the test suite for the function to hook up to. In order to prevent such tight coupling, it is advised to leave the selector as a parameter, or sometimes, simply pass in the element itself.
 
 ```js
 function populateStatusMessage(selector, message, status) {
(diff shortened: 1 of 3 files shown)


contents/javascript/javascript.md
```diff
@@ -0,0 +1,114 @@
...
+    error: function() {
+        $(button).html('An Error Occurred, Please Retry');
+    },
+    success: function(data) {
+        ...
+    }
+});
+```
+
+It order to test such a function, we would now have to has both logic and also the mock-up generated. Splitting the logic and markup in two separate functions will both make it easier to test and composable because now you can reuse code that generates the markup in multiple places.
```
Skaty: `now have to has`

I guess this is a typo?

Also, `test and composable` doesn't sound quite right.

contents/javascript/javascript.md
```diff
@@ -0,0 +1,114 @@
...
+$.ajax({
+    type: 'POST',
+    url: '/admin/adminStudentGoogleIdReset?' + params,
+    beforeSend: setLoadingImage,
+    error: setErrorText,
+    success: displayResults
+});
+```
+
+Already, we are seeing something traces leading to the MVC, albeit in a very small scale.
```
Skaty: Some traces?

contents/javascript/javascript.md
```diff
@@ -0,0 +1,114 @@
...
+        },
+ 
+        reset: function() {
+            id = 0;     
+        }
+    };  
+})();
+```
+
+In the above, only `myApp` is declared in the global scope, and we can access its methods through the object notation, e.g. `myApp.next()`. This is also especially useful to declare private variables that be used among functions. `id` is not accessible outside of scope in this example.
```
Skaty: Instead of `In the above`, why not something like, `As shown/demonstrated above`.

contents/javascript/javascript.md
```diff
@@ -0,0 +1,114 @@
+# Writing Testable Javascript
+
+JavaScript is a powerful language, and because it is powerful, there are multiple ways to go about doing the same thing. The end result is that multiple collaborators working on a single project can produce different code that does the same thing. That is why there is need for a standard to writing JavaScript code, especially on the front end.
```
Skaty: > JavaScript is a powerful language, and because it is powerful, there are multiple ways to go about doing the same thing.

Reword this, it just sounds awkward.

Skaty (changes requested):
Just a lot of nitpicking.

damithc (commented):
An overall comment after a quick skim:
1. It sounds like you came up with all these guidelines. If that's not the case, you should cite the sources. 
2. I'm sure there are more of such guidelines out there. Any guidance as to where the reader can go from here?
On a related note, we have js coding standard at oss-generic and we want to have a js best practices doc to go with it. If you want, you can add your work in those two docs instead of here.

li-kai:
> I'm sure there are more of such guidelines out there. Any guidance as to where the reader can go from here?

There is indeed. I'll put up a few links.

damithc:
@li-kai Request a review from me after approved by peer reviewer.

li-kai:
@Skaty could you review again? thanks!

Skaty (approved):
LGTM

li-kai:
@damithc Should the resources be in discrete chunks or continuous? The content is ready for review, but the entire thing isn't finished, so it'll read like it got cut off at the end haha.

damithc:
> The content is ready for review, but the entire thing isn't finished, so it'll read like it got cut off at the end haha.

Perhaps a brief explanation at the end to say what will be (or should be) added in future to make it a complete picture?

contents/javascript/javascript.md
```diff
@@ -0,0 +1,272 @@
+# Writing Testable Javascript
```
damithc, line 1: Link from ToC.md

Need to add `Author:`

contents/javascript/javascript.md
```diff
@@ -0,0 +1,272 @@
...
+
+JavaScript is a powerful language. However, its flexibility leads to multiple ways for people to go about doing the same thing. The end result is that multiple collaborators working on a single project can produce different code that does the same thing.
+
+That is why there is a need to follow a standard way of writing JavaScript - it allows for more maintainable cleaner and more beautiful code.
+
+Good JavaScript code should be testable, composable and reusable.
```
damithc: 'composable'  may not be self explanatory.

damithc: In any case, it's not addressed in the rest of the document. Only testable and reusable are addressed.

contents/javascript/javascript.md
```diff
@@ -0,0 +1,272 @@
...
+
+## Writing testable JavaScript
+
+### Avoid coupling with selectors
```
damithc: Perhaps phrase all headings at this level the same way? 
e.g. these two are phrased differently

### Avoid coupling with selectors
### Avoiding big anonymous functions

contents/javascript/javascript.md
```diff
@@ -0,0 +1,272 @@
...
+
+However, like any design pattern, this is not to be abused. Only use method chaining when methods are related to the object at hand.
+
+### Don't reinvent the wheel
+
+The limitation to Javascript is its limited utility library. That is why libraries like [lodash](https://lodash.com/) is such a popular module. There are a lot of functions being provided, which would help a lot in making shorter, more coherent syntax. Lodash reduces the need to write and maintain your own `_.assign` or `_.range` and even provides powerful functions such as `_.merge` and `_.groupBy`.
+
+As a beginner, there may be some hesistation to use lodash or external utility libraries. The big advantage is the amount of benefit it brings to the team. Whether its the overhead of having to load the `lodash` library, or having to learn a large library's syntax, not having to maintain your own utility functions makes the project more manageable, maintainable and that alone makes it worth its cost.
+
+## Testing Libraries
```
damithc: Feels like this sections belongs in a different document. This document is about writing clean JS code.

contents/javascript/javascript.md
```diff
@@ -0,0 +1,272 @@
...
+- [Mocha](https://mochajs.org/)
+- [Jasmine](https://jasmine.github.io/)
+- [QUnitJs](https://qunitjs.com)
+- [Jest](https://facebook.github.io/jest/)
+- [AVA](https://github.com/avajs/ava)
+- [Tape](https://github.com/substack/tape)
+
+So which one do you choose? Among the above, I have had the opportunity of write in all of the above except Tape. 
+
+My general advice is to go for Mocha if you are new, due to its [large community](http://stateofjs.com/2016/testing/) and therefore overall amount of help available. When you are familiar with Mocha and understand its pitfalls (speed and organisation), you can explore other libraries such as Jest.
```
damithc: This doc will have multiple authors in future. Avoid `My`.

contents/javascript/javascript.md
```diff
@@ -0,0 +1,272 @@
...
+
+Jest, due to its speed, support for React.js, support for asynchronous testing and helpful terminal outputs, seems to be the upcoming major player. The fact that it is being made by Facebook is icing to the cake.
+
+The odd test framework in the list above is QUnit, which is a test framework that runs on the browser instead of node.js like the others. QUnit has been around longer. Unless you are running JQuery and need to test for browser or UI related bugs like in [TEAMMATES](https://github.com/TEAMMATES/teammates/), it is no longer a conventional choice.
```
damithc: TEAMMATES is not a good example for a general audience.

contents/javascript/javascript.md
```diff
@@ -0,0 +1,272 @@
...
+[Airbnb Javascript Style](https://github.com/airbnb/javascript)
+It is one thing to follow the style guide, and another to understand why it is that way. Understanding why Airbnb chose certain constructs and syntax reveals ways to write code that is clean, understandable and maintainable.
+
+[JavaScript Design Patterns](https://addyosmani.com/resources/essentialjsdesignpatterns/book/)
+The title says it all. From the most common to obscure patterns, this book covers design patterns and explains trade offs. Although it specifically caters to Javascript, it's recommended reading for all prospective software engineers.
+
+## References
+[Namespacing in Javascript](https://javascriptweblog.wordpress.com/2010/12/07/namespacing-in-javascript/)
+
+[MMethod Chaining in JavaScript](https://schier.co/blog/2013/11/14/method-chaining-in-javascript.html)
```
damithc: Technically, _references_ should be referenced explicitly in the main text. For example, something like this:

> The tool X is currently the most popular tool in the marked [source: [X Dev Guide]()]. However, tool Y is also ...
> ...
> ### References
> * [X Dev Guide] [_A Guide to Developing on X_]() - This is the official Developer Guide for X

damithc (commented):
Good work  @li-kai 
Some comments added. Revise as you see fit.
Adding more content and/or getting more peer reviews is welcome but not required. 
I will be merging end of week 12

contents/javascript/javascript.md
mauris: If there's going to be a document on testing for web apps, there's [Karma](https://github.com/karma-runner/karma) that launches multiple browsers automatically for testing (something similar to what the automated GUI testing does in Address Book)

contents/javascript/javascript.md
```diff
@@ -0,0 +1,272 @@
...
+There are a large amount of Javascript testing libraries. The below is a non-exhaustive list.
+
+- [Mocha](https://mochajs.org/)
+- [Jasmine](https://jasmine.github.io/)
+- [QUnitJs](https://qunitjs.com)
+- [Jest](https://facebook.github.io/jest/)
+- [AVA](https://github.com/avajs/ava)
+- [Tape](https://github.com/substack/tape)
+
+So which one do you choose? Among the above, I have had the opportunity of write in all of the above except Tape.
```
acjh: > So which one do you choose? Among the above, I have had the opportunity ~`of`~`to` write in all ~`of the above`~ except Tape.

Also consider ordering the list alphabetically :)

PR #7, opened 2017-02-10
Java 8 Streams - An introduction

leeyimin:
## Outline  ##

The contents will be targeted at students who have some familiarity with Java but have no experience in functional-style programming.

- Benefits of using Streams
- What are Functional interfaces?
- How to use streams?
- Common mistakes to look out for.

### DIFF
--- a/contents/java/streams-an-introduction.md
+++ b/contents/java/streams-an-introduction.md
@@ -84,9 +84,9 @@ If used in the same context, these two lambda expressions are actually equivalen
 
 The parameters types of lambda expressions can be inferred by the compiler and are optional. It is actually recommended that parameters types are omitted when writing lambda expressions so as to keep the code concise.
 
-When an expression is used for the `<body>`, the result of the expression is returned. The expression can also result in nothing (eg. `(String s) -> System.out.println(s)`), which means that the method expressed by the lambda returns `void`.
+When an expression is used for the body, the result of the expression is returned. The expression can also result in nothing (eg. `(String s) -> System.out.println(s)`), which means that the method expressed by the lambda returns `void`.
 
-When a block is used for the `<body>`, the same rules for using or omitting the `return` statement for a normal method applies.
+When a block is used for the body, the same rules for using or omitting the `return` statement for a normal method applies.
 
 When the lambda has a single parameter, the parentheses surrounding the parameter can also be removed,
 ```java
@@ -117,12 +117,27 @@ students.sort(new Comparator<Student>() {
 ```
 Let's face it. Declaring classes (anonymous or not) is quite a pain. Declaring a named class adds to the number of classes you need to maintain while the syntax of a anonymous class is quite an eyesore.
 
-In Java 8, this interface has become a functional interface (surprise, surprise), which allows you to write the code like this instead!
+In Java 8, this interface has become a functional interface (surprise, surprise), so you can write a comparator with a lambda expression.
 ```java
 // trying to sort students by height
 students.sort((s1, s2) -> Double.compare(s1.getHeight(), s2.getHeight()));
 ```
-The compiler is able to infer that a object of type `Comparator<Student>` is expected and that the lambda expression fits into the definition for `compare` (the single abstract method), thus creating an instance of type `Comparator<Student>` with the lambda expression implemented as the `compare` method.
+The compiler is able to infer that an object of type `Comparator<Student>` is expected and that the lambda expression fits into the definition for `compare` (the single abstract method), thus creating an instance of type `Comparator<Student>` with the `Double.compare(s1.getHeight(), s2.getHeight())` returned in the implemeneted `compare` method.
+
+By using an lambda expression, the code is much more simplified, and can be now written on a single line. However, the expression in the body is slightly complicated and it may not be easily understood by everyone.
+
+Suppose this comparison is used over and over again in the code. To improve cohesion in the code, you may want to add instance method `compareToByHeight(Student other)` in the `Student` class.
+```java
+public int compareToByHeight(Student other){
+    return Double.compare(height, other.height);
+}
+```
+And you can use a method reference as the functional interface instead.
+```java
+// trying to sort students by height
+students.sort(Student::compareToByHeight);
+```
+Notice that the code is very easy to understand at a high level and what the author is trying to achieve can be easily read through the code.
 
 ### The Stream API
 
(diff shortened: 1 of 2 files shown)


damithc:
@leeyimin can you get this peer reviewed by someone? Request a review from me after approved by peer reviewer.

contents/java/streams-an-introduction.md
```diff
@@ -0,0 +1,75 @@
+# Java 8 Streams - An Introduction
+
+Authors: Lee Yi Min
```
yamgent: You can add a link to your resume/website for your name.

contents/java/streams-an-introduction.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Overview
+
+In Java 8, we were introduced to new features such as lambda expressions and streams. If you weren't familiar with the concept of functional programming, you might be (like me,) silently screaming in your head as you stare at a chunk of code infused with lambda expressions and stream operations.
```
yamgent: The comma in `(like me,)` feels out of place, I think it can be removed.

contents/java/streams-an-introduction.md
```diff
@@ -0,0 +1,75 @@
...
+double meanMaleHeight = students.toStream()
+                                .filter(student -> student.isMale())
+                                .mapToDouble(student -> student.getHeight())
+                                .average()
+                                .orElse(Double.NaN);
+```
+
+Not only is the numbers of lines reduced (almost by half!), the code utilising streams is also rather intuitive. First, we filter the students who are male, using a lambda expression (`student -> student.isMale()`). Then, we get the heights of these students and followed by the average value of these heights. If the average value does not exist (which can happen when there are no male students), we store NaN in `meanMaleHeight` instead.
+
+Besides its brevity, another cool feature of streams is that code can be executed in parallel using your multicore processor, which leads to faster performance. And the best part is this can be done just by using a simple method call in the Stream API, you don't have implement multithreading or worry about how to go about splitting the work for it to work in parallel.
```
yamgent: Perhaps can give a brief hint on how it executes in parallel -> the students are processed all at once, rather than one-by-one.

contents/java/streams-an-introduction.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Getting Started
+
+Before we can get started on Stream, it is imperative to first have some understanding on the concept of functional interface and lambda expressions. Feel free to skip right ahead if you are already familiar with such concepts!
+
+### Functional Interface and Lambda Expressions
+
+If you have tried using the Stream library and looked at its [API page](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html), you would see that many of its methods accept objects of new types such as `Predicate`, `Collector` and more. If you navigate into the APIs of these new types, you would have realised they are merely interfaces (and the APIs themselves aren't giving much help on how you are supposed to make use of these interfaces).
+
+These interfaces are actually functional interfaces, which are simply __interfaces declared with a single abstract method__ (, excluding abstract methods which overrides public methods of `java.lang.Object`). These interfaces provide an easy way for you to somehow reference a method, which could be a existing class method, class constructor or a lambda expression.
```
yamgent: The comma may not be necessary in `(, excluding abstract...`

yamgent:
```
These interfaces provide an easy way for you to somehow reference a method, which
could be a existing class method, class constructor or a lambda expression.
```

Actually, they allow the compiler to self-infer the methods, since such functional interfaces must have only one method, so smarter compilers allows us to write more concise code (EDIT: I wrote this comment because the word 'somehow' is used :P).

contents/java/streams-an-introduction.md
```diff
@@ -0,0 +1,75 @@
...
+
+Let's see this through an example.
+
+`Comparator<T>` may be a familiar interface to you. In previous Java editions, to create an object extending this interface, you could either instantiate the object from a named class or an anonymous class, both of which need to override the abstract methods in `Comparator<T>`. So the code may look something like this:
+```
```
yamgent: You can use ` ```java ` so that the code snippet will be syntax highlighted as Java code (otherwise no color is rendered).

contents/java/streams-an-introduction.md
```diff
@@ -0,0 +1,75 @@
...
+The compiler is able to infer that a object of type `Comparator<Student>` is expected and that the lambda expression fits into the definition for `compare` (the single abstract method), thus creating an instance of type `Comparator<Student>` with the lambda expression implemented as the `compare` method.
+
+### The Stream API
+
+## Using Java 8 Streams
+
+### Drawbacks and Pitfalls
+
+## Resources
```
yamgent: You can consider linking any reference materials used to write the book chapter, or additional materials that introduce other features of Stream.

yamgent (commented):
The book chapter demonstrates clearly the advantages of using Streams, and allows readers who are not completely familiar with functional languages to still utilise it by briefly describing about functional interface in an easy-to-understand manner.

damithc:
@leeyimin Let me know when ready for my review

contents/java/streams-an-introduction.md
```diff
@@ -0,0 +1,212 @@
...
+  });
+```
+Let's face it. Declaring classes (anonymous or not) is quite a pain. Declaring a named class adds to the number of classes you need to maintain while the syntax of a anonymous class is quite an eyesore.
+
+In Java 8, this interface has become a functional interface (surprise, surprise), which allows you to write the code like this instead!
+```java
+// trying to sort students by height
+students.sort((s1, s2) -> Double.compare(s1.getHeight(), s2.getHeight()));
+```
+The compiler is able to infer that a object of type `Comparator<Student>` is expected and that the lambda expression fits into the definition for `compare` (the single abstract method), thus creating an instance of type `Comparator<Student>` with the lambda expression implemented as the `compare` method.
```
yamgent: `a object` -> `an object`

contents/java/streams-an-introduction.md
```diff
@@ -0,0 +1,212 @@
...
+
+### The Stream API
+
+// source of elements
```
yamgent: I assume you will be adding these by today? Otherwise the TODOs will look weird if merged.

yamgent (commented):
Nice improvements made on the book chapter to include more explanations about the different aspects of Java 8 Streams.

yamgent (approved):

PR #13, opened 2017-02-13
Writing testable code

chao1995:


contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,110 @@
+# Writing Testable Code
+
+Below is a collection of 10 rules that you can follow to write testable code.
```
IanTeo: Just my opinion, saying it this way makes me think of those top 10 click bait articles. Overall, the article seems to assume that the reader has knowledge of testing. Is that intended?

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Don't mix object construction with application logic
+
+Because when you are writing tests, you mostly want to isolate some part of the application.
```
IanTeo: Your header should not link in to your sentence. It feels weird to me. You could just remove the Because.

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,110 @@
...
+
+In order to instantiate the class in isolation you have to make sure that the class itself does not instantiate other objects.
```
IanTeo: Maybe you could explain why this is so?

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,110 @@
...
+
+So the first rule for writing testable code is do not mix object construction with application logic, instead, you should have two kinds of classes.
+
+First, the **factories**, where all `new` operators reside in, are in charge of constructing objects, and nothing else.
```
IanTeo: You mention **factories**, but there is no explanation of what that is.

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,110 @@
+# Writing Testable Code
+
+Below is a collection of 10 rules that you can follow to write testable code.
+
+## Don't mix object construction with application logic
```
IanTeo: Maybe you can number the rules?

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Avoid Global States
+
+Global states may be tolerable at runtime if you only have one instance of your application.
+
+However, each test is a small instantiation of your application. But the global states persists from one test to another (e.g. a global configuration file).
+
+When tests fail, there is possibility it cannot be reproduced in isolation due to the global states. In other words, order of tests matter, which is bad.
+
+## Avoid Singletons
```
IanTeo: If this is going to be a point that is different from "Avoid Global States", maybe you could elaborate more?

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Don't look for things. Ask for things.
+
+Aka Dependency Injection / Law of Demeter.
+
+The [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) can be highlighted as:
+
+* Only talk to your **immediate** friends.
+
+> If you are an Authenticator class don't pass in a path of the configuration information which you read inside the constructor to configure yourself, just ask for the configuration object and let some other class worry about reading the object from the disk.
```
IanTeo: I can't really understand what you are trying to say here. Maybe you could clarify it for me?

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Avoid Static Methods
+
+Because if you build your application only with static methods, even if you are using Java and Class, it is not an OOP application but a procedural application.
+
+And there is *no way* to unit test procedure application. Why?
+
+> The key to testing is the presence of seams (places where you can divert the normal execution flow). Seams are essentially polymorphism. Seams are needed so that you can isolate the unit of test.
+
+However,
```
IanTeo: Weird formatting. Why just one word? Maybe you can omit it completely?

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,110 @@
...
+
+> A leaf method such as Math.abs() is not a problem since the execution call graph ends there. But if you pick a method in a core of your application logic than everything behind the method becomes hard to test, since there is no way to insert test doubles (and there are no seams). Additionally it is really easy for a leaf method to stop being a leaf and than a method which was OK as static no longer is.
+
+## Favor composition over inheritance
+
+Because
```
IanTeo: Weird formatting here too

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,110 @@
...
+
+> At run-time you can not chose a different inheritance, but you can chose a different composition. This is important for tests as we want to test thing in isolation.
+
+Wrongly used inheritance hurts clutter the focus of test because you need to mock out the your parent classes' irrelevant dependencies, too.
+
+> Inheriting from AuthenticatedServlet will make your sub-class very hard to test since every test will have to mock out the authentication. And you have to do so to successfully navigate the super class. But what if AuthenticatedServlet inherits from DbTransactionServlet? (that gets so much harder)
+
+## Favor polymorphism over conditionals
```
IanTeo: Usually polymorphism has inheritance. Since your previous point was "Favor composition over inheritance", does this mean we should "Favor composition over conditionals"? Maybe you could be abit more specific for these 2 points

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,110 @@
...
+
+If a class has `switch` or the same `if` conditions in many places, you need to break down your complex class into several smaller simpler classes.
+
+Because testing smaller and simpler classes is easier, polymorphism can help you do that.
+
+## Mixing Service Objects with Value Objects
+
+Values objects are your **model** objects, like `User`, `Email`, `CreditCard`.
```
IanTeo: Value~~s~~ objects

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,110 @@
...
+
+> Inheriting from AuthenticatedServlet will make your sub-class very hard to test since every test will have to mock out the authentication. And you have to do so to successfully navigate the super class. But what if AuthenticatedServlet inherits from DbTransactionServlet? (that gets so much harder)
+
+## Favor polymorphism over conditionals
+
+If a class has `switch` or the same `if` conditions in many places, you need to break down your complex class into several smaller simpler classes.
+
+Because testing smaller and simpler classes is easier, polymorphism can help you do that.
+
+## Mixing Service Objects with Value Objects
```
IanTeo: Do you mean "Do not mix Service Objects with Value Objects"?

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,110 @@
...
+1. Service objects are harder to construct. And their constructors ask for lots of other objects for collaboration.
+2. Service objects are harder to construct and as a result are never constructed with a new operator in-line, (instead use factory / DI-framework) for the object graph construction.
+**Note**: service objects don't take value-objects in their constructors since DI-frameworks tend to be unaware about the how to create a value-object.
+3. Service objects are harder to test since they are all about collaboration and as a result we are forced to use mocking, something which we want to minimize.
+
+Clearly, they are very different in their nature and usage.
+
+> Mixing the two creates a hybrid which has no advantages of value-objects and all the baggage of service-object.
+
+## Mixing of Concerns
```
IanTeo: Do you mean "Do not Mix Concerns"? or something along those lines?

IanTeo (commented):
Towards the end of your document, you start to lose focus on what your document is about. Maybe you could look at the last 2 points again. Also, I don't think you should treat your headers as part of your content.

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,110 @@
...
+
+Global states may be tolerable at runtime if you only have one instance of your application.
+
+However, each test is a small instantiation of your application. But the global states persists from one test to another (e.g. a global configuration file).
+
+When tests fail, there is possibility it cannot be reproduced in isolation due to the global states. In other words, order of tests matter, which is bad.
+
+## Avoid Singletons
+
+Because singletons are global states in sheep's clothing! It's easy to understand because Singletons enforce their singletoness through private constructor and a global instance variable.
```
Neurrone: "Its easy to understand" - what's the its?

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Avoid Static Methods
+
+Because if you build your application only with static methods, even if you are using Java and Class, it is not an OOP application but a procedural application.
```
Neurrone: Suggest not mentioning a specific programming language for this, as this applies equally well to any OOP language.

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,110 @@
...
+
+However,
+
+> A leaf method such as Math.abs() is not a problem since the execution call graph ends there. But if you pick a method in a core of your application logic than everything behind the method becomes hard to test, since there is no way to insert test doubles (and there are no seams). Additionally it is really easy for a leaf method to stop being a leaf and than a method which was OK as static no longer is.
+
+## Favor composition over inheritance
+
+Because
+
+> At run-time you can not chose a different inheritance, but you can chose a different composition. This is important for tests as we want to test thing in isolation.
```
Neurrone: chose -> choose

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,110 @@
...
+
+Wrongly used inheritance hurts clutter the focus of test because you need to mock out the your parent classes' irrelevant dependencies, too.
+
+> Inheriting from AuthenticatedServlet will make your sub-class very hard to test since every test will have to mock out the authentication. And you have to do so to successfully navigate the super class. But what if AuthenticatedServlet inherits from DbTransactionServlet? (that gets so much harder)
+
+## Favor polymorphism over conditionals
+
+If a class has `switch` or the same `if` conditions in many places, you need to break down your complex class into several smaller simpler classes.
+
+Because testing smaller and simpler classes is easier, polymorphism can help you do that.
```
Neurrone: Provide an example?

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Mixing Service Objects with Value Objects
+
+Values objects are your **model** objects, like `User`, `Email`, `CreditCard`.
+
+1. They are (or should be) very easy to construct. And they should never take a service object in its constructor (since otherwise it is not easy to construct).
+2. You can freely create value objects with the "new" operator directly in line with your business logic (exception to [point 1](#Don’t-mix-object-construction-with-application-logic) since they are leafs of your application graph).
+3. They are **never mocked**.
+
+Service objects, in the other way, are your **logic** objects, like `UserAthenticator`, `MailServer`, `CreditCardProcessor`. Compared to value objects,
```
Neurrone: in the other way -> on the other hand?
UserAthenticator -> UserAuthenticator

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Mixing of Concerns
+
+Single Responsibility Principle also requires you to have classes with one concern, because these classes are simpler to understand what they are doing and easier to unit test.
+
+Symptoms of mixing of concerns:
+
+* summing up what the class does includes the word "and"
+* class has fields that are only used in some methods
+* class has static methods that only operate on parameters
```
Neurrone: Why is this a sign of mixing of concerns?

Neurrone (changes requested):

contents/testing/writing-testable-code.md
chao1995: Thanks for the detailed review! I'll reformat the 10 rules under four categories (i.e the four reasons I used in lightningtalks)

contents/testing/writing-testable-code.md
chao1995: Ok!

contents/testing/writing-testable-code.md
chao1995: Ok!

contents/testing/writing-testable-code.md
chao1995: Ok! Basically for unit testing, ideally there is only one class under testing, that's why we mock out the dependencies. But if you instantiate other objects directly using `new`, that's a dependency we cannot mock out (how to mock out `new`?)

contents/testing/writing-testable-code.md
chao1995: I will add a link to wikipedia.

contents/testing/writing-testable-code.md
chao1995: Ok! Say there is a `Config` class. And you can get a `Config` by reading a configuration file stored in disk. Then there is a class A needs a `Config`, do you inject a configuration file path, or do you pass in a `Config` directly?

Why injecting a file path is not good? Think about what you need to at testing, you will need to write some configuration to a file first, then pass in the file path to A, and let A read a `Config` itself. Versus you can directly create a `Config` object on the fly and pass it to A.

In practice, the reason many people complain about Java is because it's too verbose. It's often you need to write `a.getB().getC().getD().doSomething()`. This rule forbids this. Actually you should ask for a D in constructor, and write `d.doSomething()`.

contents/testing/writing-testable-code.md
chao1995: Ok!

contents/testing/writing-testable-code.md
chao1995: Ok!

contents/testing/writing-testable-code.md
chao1995: polymorphism can also be done via interfaces. this is encouraged.

The thing about favor composition over polymorphism is that, by inheritance, you need to instantiate your parent, and parent's parent....first. If each parent takes another 5 dependencies. There is a lot of work to mock out them. But composition is very easy, if you use a mocking framework, it is as simple as `mock(ParentClass)`.

contents/testing/writing-testable-code.md
chao1995: Yes! Will fix it.

contents/testing/writing-testable-code.md
chao1995: Ok!

contents/testing/writing-testable-code.md
chao1995: Yes. Will fix it too.

contents/testing/writing-testable-code.md
chao1995: Thanks for the review! Will elaborate on this.

contents/testing/writing-testable-code.md
chao1995: Ok!

contents/testing/writing-testable-code.md
chao1995: Ok!

contents/testing/writing-testable-code.md
chao1995: Ok!

contents/testing/writing-testable-code.md
chao1995: Ok! Thanks for noticing the typo!

contents/testing/writing-testable-code.md
chao1995: Because you can usually move the static methods to instance methods of one of these parameters.

Neurrone: Hm, how's having static vs instanced methods related to mixing of concerns? Is it because static methods may access global state?

chao1995: If the static method can be moved to another class, doesn't that mean it should not be placed in this class? (this class should not concern about this method)

Neurrone: Not necessarily, for example, pure functions like java's String.valueOf methods.

chao1995: This is a good point! Similar things like `Math.abs()`. They are **acceptable** because they are at the end of call graph. But even though they are acceptable, why write Math.abs(-5) if you can write -5.abs(). (similarly, why writing String.valueOf(3), if you can 3.toString()) Java got this wrong. Ruby got this right. :)

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,146 @@
+# Writing Testable Code
+
+When the project is big enough and needs to be maintainable in the long run, it has to rely on automated tests to keep its quality. Especially, the idea of unit testing becomes very important for Object-Oriented programs. Compared to system testing where you test the program altogether, unit testing has its benefit for being fast (because it only instantiates a small piece of the program) and stable (because it usually mocks out the unstable dependency, e.g. network connection, database connection).
```
IanTeo: Don't we rely on automated tests to prevent regression, not keep quality?
Especially is not the right word to use here. Maybe you can use "Because of this,". But I think there are better ways to write it. I can't think of any at the moment.

damithc:
@chao1995 let me know after approved by peer reviewer

IanTeo (approved):
Other than the minor suggestion for improvement, LGTM

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,146 @@
...
+
+#### Warning Signs
+
+* `new` keyword in a constructor or at field declaration
+* Code does complex object graph construction inside a constructor rather than using a factory or builder
+* Anything more than field assignment in constructors
+
+> Any work you do in a constructor needs to successfully navigate through in every test (not just the direct test, but also any related test which tries to instantiate your class indirectly as part of some larger object graph). - [Google Testing Blog](https://testing.googleblog.com/2008/08/by-miko-hevery-so-you-decided-to.html)
+
+In short, the constructor's job is to assign the dependencies to fields. And that's all.
```
Neurrone: What if in the constructor you're checking that invariants hold?

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Rule #1: Avoid Global States
+
+#### Warning Signs
+
+* Adding or using static fields or static methods
+* Adding or using registries
+* Using `System.currentTimeMillis()`, `new Date()` or `Math.random()`
+
+Global states are a bad idea because they make the code hard to understand and reason.
```
Neurrone: reason. -> reason about.

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,146 @@
...
+
+To extreme, it becomes an anti-pattern, [God Object](https://en.wikipedia.org/wiki/God_object).
+
+### Rule #1: Avoid mixing of concerns
+
+#### Warning Signs:
+
+* summing up what the class does includes the word "and"
+* class has fields that are only used in some methods (objects hiding inside)
+* class has static methods that only operate on parameters (methods should be instance methods of one of the parameters)
```
Neurrone, line 140: In some languages, it isn't possible to create methods that aren't associated with a class, e.g c# or java, so making them static is the only choice, even if they only use their parameters and don't touch other state.

Neurrone (changes requested):

contents/testing/writing-testable-code.md
chao1995: > Don't we rely on automated tests to prevent regression, not keep quality?

Having automation helps to prevent regression. As a result, we also keep up the code quality, right?





> Especially is not the right word to use here. Maybe you can use "Because of this,".

Thanks for the suggestion. Decide to use "Because of this", but moving the sentence to the end.

contents/testing/writing-testable-code.md
chao1995: The rule is Don't do actual work in constructors.

checking invariants hold is actually not considered as "actual work".

contents/testing/writing-testable-code.md
chao1995: ok.

contents/testing/writing-testable-code.md
chao1995: > In some languages, it isn't possible to create methods that aren't associated with a class, e.g c# or java, so making them static is the only choice

I would argue that such methods inherently disagree with OO principle.

Neurrone (approved):

chao1995:
@damithc Prof, all peer reviewers have approved the changes.

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,146 @@
...
+
+#### Warning Signs
+
+* Create objects using `new` keyword freely anywhere in your code
+
+> You should have two kinds of classes in your application. - [Google Testing Blog](http://misko.hevery.com/2008/09/10/where-have-all-the-new-operators-gone/)
+
+First, **[factories](https://en.wikipedia.org/wiki/Factory_(object-oriented_programming))**, where all `new` operators reside in, are in charge of constructing objects, and nothing else.
+
+The other, **application classes**, are devoid of `new` operators. Instead of creating object, they simply ask for them. This makes it easier to test the application logic by replacing the real classes for test doubles.
```
damithc: These two sentences are from the Google Testing Blog too? If they are, they should be inside the block quote.

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Rule #3: Don't mix Service Objects with Value Objects
+
+Value objects are your **model** objects, like `User`, `Email`, `CreditCard`.
+
+1. They are (or should be) very easy to construct. And they should never take a service object in its constructor (since otherwise it is not easy to construct).
+2. You can freely create value objects with the "new" operator directly in line with your business logic (exception to previous rule) since they are leafs of your application graph).
+3. They are **never mocked**.
```
damithc, line 51: Where did you find these rules? Need to credit the original source.

contents/testing/writing-testable-code.md
```diff
@@ -0,0 +1,146 @@
...
+
+By using polymorphism, you get the same functionality and because a new case is a new class you don't have to search your code for things that need to be updated. It is all isolated for each class.
+
+This closely follows with [Open Closed Principle](https://en.wikipedia.org/wiki/Open/closed_principle) because you can ship this abstract parent class (**closed for modification**) as part of your binary library, but people can still extend the functionality by adding new child classes (**open for extension**).
+
+# Concluding Notes
+
+This article includes ten rules (5 + 2 + 1 + 2) in total in a flavor of theory explanation. Nevertheless I hope this article can help you understand some key concepts, such as seams, dependency injection, global states, singletons and Singleton. Also, I hope you can apply these rules into practice, like writing a program with these rules in mind or reviewing some code your wrote before and see whether you can improve its testability, so you can benefit from the things you learn in this article.
```
damithc: > This article includes ten rules (5 + 2 + 1 + 2) in total in a flavor of theory explanation. Nevertheless I hope this article can ...

This article includes ten rules that can ...

Avoid `I` / `My` because this article might have multiple authors in a future version.

damithc (commented):
Good work @chao1995 
Some comments added. Revise as you see fit.
Adding more content and/or getting more peer reviews is welcome but not required. 
I will be merging end of week 12

contents/testing/writing-testable-code.md
chao1995: Yes they are. I added links to the original posts at the end of every rule.

PR #24, opened 2017-04-24
ToC: Fix JavaScript link

paradite:
fix upper case letter of JavaScript link

Original link was upper case and leads to 404 page:

https://github.com/se-edu/learningresources/blob/master/contents/javascript/Javascript.md


damithc:
Thanks for the fix @paradite 👍

PR #28, opened 2018-02-14
Add more content to Findbugs

tshradheya:
1. Added about some types of bugs that FindBugs can find
1. Mentioned about SpotBugs as successor of dead project FindBugs
1. Fixed a minor bug in link of Checkstyle

### DIFF
--- a/contents/staticAnalysis/FindBugs.md
+++ b/contents/staticAnalysis/FindBugs.md
@@ -31,7 +31,7 @@ FindBugs analyses bytecode in compiled Java `.class` file and checks multiple fi
 
 ## Examples of Bugs that can be found using FindBugs
 
-### Find hash-equals mismatch
+### Incorrectly overriding methods
 
 Consider the following code:
 
@@ -47,9 +47,11 @@ class Foo {
 }
 ```
 
-1.  In the above code, if `foo.equals()` method is called, the `equals()` method of `Object` class rather than `Foo` class will be called. This is due to the way the Java code resolves overloaded methods at compile-time. FindBugs warns the developer of possible cases when a class defines a co-variant version of the `equals()` or `compareTo()` method.
+In the above code, if `foo.equals()` method is called, the `equals()` method of `Object` class rather than `Foo` class will be called. This is due to the way the Java code resolves overloaded methods at compile-time. FindBugs warns the developer of possible cases when a class defines a co-variant version of the `equals()` or `compareTo()` method.
 
-1. The hashCode() and equals() method are called by many `Collection` based classes like - List, Maps, Sets, etc. FindBugs helps in finding problems when a class **overrides the `equals()` but not the `hashCode()` method or vice-versa**. Overriding only one of the `equals()` or `hashCode()` method can cause methods of Collection based classes to fail and hence FindBugs helps in reporting these errors at an early stage
+### Find hash-equals mismatch
+
+The hashCode() and equals() method are called by many `Collection` based classes like - List, Maps, Sets, etc. FindBugs helps in finding problems when a class **overrides the `equals()` but not the `hashCode()` method or vice-versa**. Overriding only one of the `equals()` or `hashCode()` method can cause methods of Collection based classes to fail and hence FindBugs helps in reporting these errors at an early stage
 
 ### Return value of method ignored
 
--- a/contents/staticAnalysis/FindBugs.md
+++ b/contents/staticAnalysis/FindBugs.md
@@ -1,16 +1,73 @@
 # FindBugs
 
-Author: [Xiao Pu](https://nus-oss.github.io/cs3281-website/students/AY1617S2/xiaoPu/xiaoPu-Resume.html)
+Author: [Xiao Pu](https://nus-oss.github.io/cs3281-website/students/AY1617S2/xiaoPu/xiaoPu-Resume.html), [Shradheya Thakre](https://github.com/tshradheya)
 
 ## Overview
 
-FindBugs is a static analysis tool to find "bugs" in **Java** programme. It looks for "bug patterns" in the code and signals possible violations. For example, the bug "null pointer dereference" has the pattern — A programme declares a non-nullable variable but assigns `null` to the variable somewhere and uses it later.
+FindBugs is a static analysis tool to find "bugs" in **Java** programme. It looks for "bug patterns" in the code and signals possible violations. Potential errors are classified in four ranks:
+1. scariest,
+1. scary,
+1. troubling,
+1. of concern.
+
+This is a hint to the developer about their possible impact or severity For example, the bug "null pointer dereference" has the pattern — A programme declares a non-nullable variable but assigns `null` to the variable somewhere and uses it later.
 
 ## Features
 
 The "bug patterns" can be divided into nine groups: Bad practice, Correctness, Experimental, Internationalization, Malicious code vulnerability, Multithreaded correctness, Performance, Security and Dodgy code. [A comprehensive list](http://findbugs.sourceforge.net/bugDescriptions.html) of bugs is provided to explain the meaning of each bug.
 
 FindBugs analyses bytecode in compiled Java `.class` file and checks multiple files at the same time. These features overcome some limitations in [CheckStyle](CheckStyle.md) and [PMD](PMD.md) (The two tools can only check files one by one and analyse Java source code). Therefore, FindBugs can spot some errors that CheckStyle and PMD cannot find. For example, one of the bug patterns in FindBugs is `RCN: Redundant nullcheck of value known to be non-null`. FindBugs will analyse all the assignments to a particular variable in the code base and then check whether the `nullcheck` for the variable is redundant or not.
+
+## Examples of Bugs that can be found using FindBugs
+
+### Find hash-equals mismatch
+
+Consider the following code:
+
+``` java
+class Foo {
+    //... data members ...
+    //... methods ...
+
+    //Overriding equals method - the wrong way
+    public boolean equals(Foo foo) {
+        //... logic ...
+    }
+}
+
+```
+
+1.  In the above code, if `foo.equals()` method is called, the `equals()` method of `Object` class rather than `Foo` class will be called. This is due to the way the Java code resolves overloaded methods at compile-time. FindBugs warns the developer of possible cases when a class defines a co-variant version of the `equals()` or `compareTo()` method.
+
+1. The hashCode() and equals() method are called by many `Collection` based classes like - List, Maps, Sets, etc. FindBugs helps in finding problems when a class **overrides the `equals()` but not the `hashCode()` method or vice-versa**. Overriding only one of the `equals()` or `hashCode()` method can cause methods of Collection based classes to fail and hence FindBugs helps in reporting these errors at an early stage
+
+### Return value of method ignored
+
+* FindBugs helps in finding places where your code has ignored the return value of method when it shouldn't have been
+
+* ``` Java
+    1 String s = "bob";
+    2 s.replace('b', 'p');
+    3 boolean isCorrect = s.equals("pop"); //isCorrect is `false`
+    ```
+    In the above examples, one would assume that the variable `isCorrect` is assigned `true` because the `line 2` replaces `b` with `p`. However since strings are immutable, the `replace()` function actually returns a new string with updated value rather than updating the string the method is called on.
+    Hence, `line 2` should be `String newString  = s.replace('b', 'p'); //newString ="pop"`
+
+### Null pointer dereference
+
+* FindBugs looks for cases where a code path will or could cause a null pointer exception.
+
+* ``` Java
+
+    1  Person person = aMap.get("bob");
+    2  if (person != null) {
+    3      person.updateAccessTime();
+    4  }
+    5  String name = person.getName();
+    ```
+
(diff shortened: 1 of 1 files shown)


contents/staticAnalysis/FindBugs.md
```diff
@@ -1,17 +1,74 @@
...
+class Foo {
+    //... data members ...
+    //... methods ...
+
+    //Overriding equals method - the wrong way
+    public boolean equals(Foo foo) {
+        //... logic ...
+    }
+}
+
```
crphang: Might want to delete this empty line.

contents/staticAnalysis/FindBugs.md
```diff
@@ -1,17 +1,74 @@
...
+Author: [Xiao Pu](https://nus-oss.github.io/cs3281-website/students/AY1617S2/xiaoPu/xiaoPu-Resume.html), [Shradheya Thakre](https://github.com/tshradheya)
 
 ## Overview
 
-FindBugs is a static analysis tool to find "bugs" in **Java** programme. It looks for "bug patterns" in the code and signals possible violations. For example, the bug "null pointer dereference" has the pattern — A programme declares a non-nullable variable but assigns `null` to the variable somewhere and uses it later.
+FindBugs is a static analysis tool to find "bugs" in **Java** programme. It looks for "bug patterns" in the code and signals possible violations. Potential errors are classified in four ranks:
+1. scariest,
+1. scary,
+1. troubling,
+1. of concern.
```
crphang: Would it be clearer if this is written as 1. 2. 3. 4. even though it is rendered similarly? I went through some other book chapters and some authors wrote it as 1. 1. 1. 1. while others wrote in numerically, so I am not sure whether there is a particular standard to this.

contents/staticAnalysis/FindBugs.md
```diff
@@ -1,17 +1,74 @@
...
+
+* FindBugs helps in finding places where your code has ignored the return value of method when it shouldn't have been
+
+* ``` Java
+    1 String s = "bob";
+    2 s.replace('b', 'p');
+    3 boolean isCorrect = s.equals("pop"); //isCorrect is `false`
+    ```
+    In the above examples, one would assume that the variable `isCorrect` is assigned `true` because the `line 2` replaces `b` with `p`. However since strings are immutable, the `replace()` function actually returns a new string with updated value rather than updating the string the method is called on.
+    Hence, `line 2` should be `String newString  = s.replace('b', 'p'); //newString ="pop"`
```
crphang: I think you wanted to add a newline to Hence, but I think you need to add another newline in order to achieve that.

image

contents/staticAnalysis/FindBugs.md
tshradheya: I thought of following the MarkDown syntax and hence 1. 1. 1. ...

tshradheya:
Thanks @Crphang for the review. 🎉  I have made the necessary changes. I personally feel going with ` 1. 1. 1.` is preferrable cause it follows the markdown syntax

contents/staticAnalysis/FindBugs.md
yamgent: Writing it as 1. 1. 1. has an additional benefit: should you remove/add items in the middle of the list, you need not have to renumber everything else.

crphang: I see. Thanks for pointing out.

tshradheya:
@xpdavid Requesting a review( Can ignore if you are busy with other work :P)

contents/staticAnalysis/FindBugs.md
junkiattan: By numbering the points, you can remove the commas or full-stops in each point

contents/staticAnalysis/FindBugs.md
```diff
@@ -1,16 +1,71 @@
...
 ## Overview
 
-FindBugs is a static analysis tool to find "bugs" in **Java** programme. It looks for "bug patterns" in the code and signals possible violations. For example, the bug "null pointer dereference" has the pattern — A programme declares a non-nullable variable but assigns `null` to the variable somewhere and uses it later.
+FindBugs is a static analysis tool to find "bugs" in **Java** programme. It looks for "bug patterns" in the code and signals possible violations. Potential errors are classified in four ranks:
+1. scariest,
+1. scary,
+1. troubling,
+1. of concern.
+
+This is a hint to the developer about their possible impact or severity For example, the bug "null pointer dereference" has the pattern — A programme declares a non-nullable variable but assigns `null` to the variable somewhere and uses it later.
```
junkiattan: 1. You missed a fullstop after "severity"
2. Also, it would be good to highlight the pattern in a different format from the text. Perhaps put them in quotes or something to differentiate the text from the message

contents/staticAnalysis/FindBugs.md
```diff
@@ -1,16 +1,71 @@
...
 
 ## Features
 
 The "bug patterns" can be divided into nine groups: Bad practice, Correctness, Experimental, Internationalization, Malicious code vulnerability, Multithreaded correctness, Performance, Security and Dodgy code. [A comprehensive list](http://findbugs.sourceforge.net/bugDescriptions.html) of bugs is provided to explain the meaning of each bug.
```
junkiattan: Perhaps can consider putting into a numbered list to improve readability

contents/staticAnalysis/FindBugs.md
```diff
@@ -1,16 +1,71 @@
...
+FindBugs analyses bytecode in compiled Java `.class` file and checks multiple files at the same time. These features overcome some limitations in [CheckStyle](checkStyle.md) and [PMD](PMD.md) (The two tools can only check files one by one and analyse Java source code). Therefore, FindBugs can spot some errors that CheckStyle and PMD cannot find. For example, one of the bug patterns in FindBugs is `RCN: Redundant nullcheck of value known to be non-null`. FindBugs will analyse all the assignments to a particular variable in the code base and then check whether the `nullcheck` for the variable is redundant or not.
```
junkiattan: 1. After saying Findbugs can check multiple files simultaneously, 

> This is unlike Checkstyle or PMD which can only check files one by one and analyze Java source code, allowing FindBugs to spot errors that would have been missed by Checkstyle or PMD.

2. The example about the bug pattern does not seem to show the difference in capability of FindBugs versus other analysis tools.

contents/staticAnalysis/FindBugs.md
```diff
@@ -1,16 +1,71 @@
...
+}
+```
+
+1.  In the above code, if `foo.equals()` method is called, the `equals()` method of `Object` class rather than `Foo` class will be called. This is due to the way the Java code resolves overloaded methods at compile-time. FindBugs warns the developer of possible cases when a class defines a co-variant version of the `equals()` or `compareTo()` method.
+
+1. The hashCode() and equals() method are called by many `Collection` based classes like - List, Maps, Sets, etc. FindBugs helps in finding problems when a class **overrides the `equals()` but not the `hashCode()` method or vice-versa**. Overriding only one of the `equals()` or `hashCode()` method can cause methods of Collection based classes to fail and hence FindBugs helps in reporting these errors at an early stage
+
+### Return value of method ignored
+
+* FindBugs helps in finding places where your code has ignored the return value of method when it shouldn't have been
```
junkiattan: Can consider removing the bullet points, since it isn't really needed and you did not include bullet points in your earlier section on "Find has-equals mismatch"

contents/staticAnalysis/FindBugs.md
```diff
@@ -1,16 +1,71 @@
...
+    2 s.replace('b', 'p');
+    3 boolean isCorrect = s.equals("pop"); //isCorrect is `false`
+    ```
+    In the above examples, one would assume that the variable `isCorrect` is assigned `true` because the `line 2` replaces `b` with `p`. However since strings are immutable, the `replace()` function actually returns a new string with updated value rather than updating the string the method is called on.
+
+    Hence, `line 2` should be `String newString  = s.replace('b', 'p'); //newString ="pop"`
+
+### Null pointer dereference
+
+* FindBugs looks for cases where a code path will or could cause a null pointer exception.
```
junkiattan: same for the bullet points here

contents/staticAnalysis/FindBugs.md
tshradheya: Thanks for the review 🎉 

Regarding 2), The content was added by the previous author. Anyways I find it correct cause FindBugs works on `.class` which can only check for the RCN rule.

tshradheya:
Thanks @junkiattan for the review :)
While reviewing, you could use the following way to help improve the review thread

- Press "start a review" instead of "add single comment"
image

contents/staticAnalysis/FindBugs.md
```diff
@@ -1,16 +1,83 @@
...
+1. Security
+1. Dodgy code
+
+[A comprehensive list](http://findbugs.sourceforge.net/bugDescriptions.html) of bugs is provided to explain the meaning of each bug.
+
+FindBugs analyses bytecode in compiled Java `.class` file and checks multiple files at the same time. This is unlike [CheckStyle](checkStyle.md) or [PMD](PMD.md) which can only check files one by one and analyse Java source code, allowing FindBugs to spot errors that would have been missed by CheckStyle and PMD. For example, one of the bug patterns in FindBugs is `RCN: Redundant nullcheck of value known to be non-null`. FindBugs will analyse all the assignments to a particular variable in the code base and then check whether the `nullcheck` for the variable is redundant or not.
+
+## Examples of Bugs that can be found using FindBugs
+
+### Find hash-equals mismatch
```
xpdavid: Find equals/hash-equals mismatch?

Your first point mainly talk about overriding equals method in a wrong way. This is not relevant to hash right? Therefore, I suggest either u extract it out and form a new section or change the title of the current section.

xpdavid (commented):
A comment :P, can change or ignore.

contents/staticAnalysis/FindBugs.md
```diff
@@ -1,16 +1,85 @@
 # FindBugs
 
-Author: [Xiao Pu](https://nus-oss.github.io/cs3281-website/students/AY1617S2/xiaoPu/xiaoPu-Resume.html)
+Author: [Xiao Pu](https://nus-oss.github.io/cs3281-website/students/AY1617S2/xiaoPu/xiaoPu-Resume.html), [Shradheya Thakre](https://github.com/tshradheya)
 
 ## Overview
 
-FindBugs is a static analysis tool to find "bugs" in **Java** programme. It looks for "bug patterns" in the code and signals possible violations. For example, the bug "null pointer dereference" has the pattern — A programme declares a non-nullable variable but assigns `null` to the variable somewhere and uses it later.
+FindBugs is a static analysis tool to find "bugs" in **Java** programme. It looks for "bug patterns" in the code and signals possible violations. Potential errors are classified in four ranks:
```
damithc: `"bug patterns"` (and other places) use single quotes instead? https://www.scribendi.com/advice/when_to_use_double_or_single_quotation_marks.en.html

damithc: `programme` -> `programs`?

contents/staticAnalysis/FindBugs.md
```diff
@@ -1,16 +1,85 @@
...
+1. Correctness
+1. Experimental
+1. Internationalization
+1. Malicious code vulnerability
+1. Multithreaded correctness
+1. Performance
+1. Security
+1. Dodgy code
+
+[A comprehensive list](http://findbugs.sourceforge.net/bugDescriptions.html) of bugs is provided to explain the meaning of each bug.
```
damithc: Suggested:

Refer to [FindBugs official documentation]() for a comprehensive list ...

contents/staticAnalysis/FindBugs.md
```diff
@@ -1,16 +1,85 @@
...
+
+The hashCode() and equals() method are called by many `Collection` based classes like - List, Maps, Sets, etc. FindBugs helps in finding problems when a class **overrides the `equals()` but not the `hashCode()` method or vice-versa**. Overriding only one of the `equals()` or `hashCode()` method can cause methods of Collection based classes to fail and hence FindBugs helps in reporting these errors at an early stage
 
-FindBugs analyses bytecode in compiled Java `.class` file and checks multiple files at the same time. These features overcome some limitations in [CheckStyle](CheckStyle.md) and [PMD](PMD.md) (The two tools can only check files one by one and analyse Java source code). Therefore, FindBugs can spot some errors that CheckStyle and PMD cannot find. For example, one of the bug patterns in FindBugs is `RCN: Redundant nullcheck of value known to be non-null`. FindBugs will analyse all the assignments to a particular variable in the code base and then check whether the `nullcheck` for the variable is redundant or not.
+### Return value of method ignored
+
+FindBugs helps in finding places where your code has ignored the return value of method when it shouldn't have been
```
damithc: Missing full stop?

damithc (changes requested):
Some cosmetic changes suggest. If FindBugs is dead, I guess we need to rewrite this chapter for SpotBugs insead and mention FindBugs briefly as the predecessor but that can be a separate PR.

damithc:
`Advance Topic` -> `Advance Topics`

tshradheya:
Made the necessary changes. 
I could work later during the semester on making this chapter more focused on `SpotBugs`

damithc (approved):

PR #30, opened 2018-02-23
Update JavaReflections

Zhiyuan-Amos:


Zhiyuan-Amos:
@yamidark can you help me to review this? :) Since we were working on something like this on AB-Level4 recently.

contents/java/JavaReflections.md
```diff
@@ -1,173 +1,160 @@
...
+
+  private int getAge() {
+    // Since class is a reserved keyword, we use clazz in variable names
+    Class<Animal> animalClazz = Animal.class;
+
+    try {
+      Field f = animalClazz.getDeclaredField("age");
+      f.setAccessible(true);
+      return (int) f.get(this);
+    } catch (Exception e) {
```
yamidark: Maybe we can be more specific that we catch a `NoSuchFieldException` and `IllegalAccessException`?
Then, we can mention the `IllegalAccessException` occurs if the field was private but we did not set accessibly to `true`.

We can also mention another disadvantage of Reflections from the `NoSuchFieldException` that we create dependencies(?) on the method/field names, such that if the class changes the method/field names our program stops working, and we need to update the name on our side too?

contents/java/JavaReflections.md
```diff
@@ -1,173 +1,160 @@
...
 
 ### The Basics of Reflections
 
-Before getting started with reflections in Java, it is important to realize that classes is also an object. Every
-unique `Object` is assigned an immutable `Class` object by the JVM. This immutable `Class` object is fundamentally 
-different from *instances* of a class. The class object itself holds information about its name, the package it resides 
-in and other important information while an instance of a class holds the instanced values and methods as defined
-in the class.
+Before getting started with reflections in Java, it is important to realize that a class is also an object. From the [Java Class API](https://docs.oracle.com/javase/9/docs/api/java/lang/Class.html), we see that `Class` is a subclass of `Object`. Every unique `Object` is assigned an immutable `Class` object by the JVM. This immutable `Class` object is fundamentally different from *instances* of a class. The class object itself holds information about its name, the package it resides in and other important information while an instance of a class holds the instanced values and methods as defined in the class.
```
yamidark: Last sentence:

`The class object itself holds important information such as its name and the package it resides in while...`

Would this sound better?

contents/java/JavaReflections.md
```diff
@@ -1,173 +1,160 @@
...
 
-    // Other methods here...
+```java
+@Test
+public void foo() {
+  Sheep sheep = new Sheep();
+  Field field = Sheep.class.getDeclaredField("age");
+  field.setAccessible(true);
+  field.set(sheep, 20);
+  // perform test
```
yamidark: I believe this part needs to be wrapped in the `try` and `catch` block or `throws Exception`.

contents/java/JavaReflections.md
```diff
@@ -1,173 +1,160 @@
...
+public void foo() {
+  Sheep sheep = new Sheep();
+  TimeUnit.MINUTES.sleep(20);
+  // perform test
+}
+```
 
-        return 0;
-    }
+Alternatively, we can set a value to the private field using Reflection:
```
yamidark: Alternatively, a much simpler and efficient way to perform this test is to....

contents/java/JavaReflections.md
```diff
@@ -1,173 +1,160 @@
...
-about it in this [StackOverflow question](http://stackoverflow.com/questions/10638826/java-reflection-impact-of-setaccessibletrue).
+Suppose you want to test `getAge()`. However, you are only able to indirectly do so by testing `isProducingWool()`. This is not good as we are not able to directly verify the age of a sheep. However, with the help of Reflection, we can now test private methods.
+
+```java
+@Test
+public void foo() {
+  Sheep sheep = new Sheep();
+  Method method = Sheep.class.getDeclaredMethod("getAge");
+  method.setAccessible(true);
+  int age = (int) method.invoke(sheep);
```
yamidark: Maybe we can explain more about `method.invoke(sheep)`, if the method require parameters `int` then we also need to supply the arguments here or change the example here to take and supply some parameters.

Also, should need the `try` and `catch` block or `throws Exception`.

contents/java/JavaReflections.md
```diff
@@ -1,173 +1,160 @@
...
-while allowing the objects to be notified of any changes in another object.
+You might have learnt from your Software Engineering module that the Observer pattern can be used for objects that are interested to get notified if the state of another object is changed. The Observer pattern is useful because you can avoid creating bidirectional dependencies between two unrelated objects that have no business talking to each other while allowing the objects to be notified of any changes in another object.
 
-One prime example of the implementation of the Observer pattern is the Google Events bus used in [AddressBook Level 4](https://github.com/se-edu/addressbook-level4/)
-.The event bus uses reflections to observe all registered objects via `register` method for methods annotated with the
-`Subscribe` annotation.
+One prime example of the implementation of the Observer pattern is the Google Events bus used in [AddressBook Level 4](https://github.com/se-edu/addressbook-level4/).The event bus uses reflections to observe all registered objects via `register` method for methods annotated with the `Subscribe` annotation.
```
yamidark: >(https://github.com/se-edu/addressbook-level4/).The event bus uses reflections to observe

Missing whitespace here.

contents/java/JavaReflections.md
```diff
@@ -1,173 +1,160 @@
...
-And there you have it! What `Sheep` is really doing is to examine itself at runtime in order to obtain the `age` field
-inherited from `Animal`. If it cannot, it assumes its `age` is zero. This technique can be used in test cases to
-access private fields and methods in the class under test without modifying the behaviour of the class itself.
+#### Testing private methods
 
-You may notice that the `Sheep#getPrivateAge()` method sets the age `Field` object to be accessible and might wonder
-the implications. Fret not! The `Field#getDeclaredField()` actually returns a new `Field` instance - so you're just
-setting that particular local `Field` instance to be accessible, not the actual `age` field itself. You can read more
-about it in this [StackOverflow question](http://stackoverflow.com/questions/10638826/java-reflection-impact-of-setaccessibletrue).
+Suppose you want to test `getAge()`. However, you are only able to indirectly do so by testing `isProducingWool()`. This is not good as we are not able to directly verify the age of a sheep. However, with the help of Reflection, we can now test private methods.
```
yamidark: Suppose you want to perform a unit test for the method `getAge()`.

contents/java/JavaReflections.md
```diff
@@ -177,105 +164,81 @@ An example implementation (not the actual) of the `Subscribe` annotation might b
...
-class and converts it to a set. That is if you registered `Sheep extends Animal` as an event handler to the method,
-both `Sheep` and `Animal` will be captured by the first line. The following lines will then examine all their methods
-(during runtime!) for the `Subscribe` annotation and register the method so that it will receive the specified event
-when it is fired. Of course this implementation leaves out a lot of details but you get the idea of how java reflections
-works.
+The first line of the `findAllEventHandlersInClass` method finds all classes and its parent classes of the registered class and converts it to a set. That is if you registered `Sheep extends Animal` as an event handler to the method, both `Sheep` and `Animal` will be captured by the first line. The following lines will then examine all their methods (during runtime!) for the `Subscribe` annotation and register the method so that it will receive the specified event when it is fired. Of course this implementation leaves out a lot of details but you get the idea of how java reflections works.
```
yamidark: but you get the idea of how **Java** reflections works.

contents/java/JavaReflections.md
```diff
@@ -177,105 +164,81 @@ An example implementation (not the actual) of the `Subscribe` annotation might b
...
 
 * Reflections convert a compile-time error to a potentially destructive run-time error.
 
-  Compile time errors are easy to catch. Whenever you compile your code, the compiler cleverly spots any error you
-  missed and points it out (along with line number and other useful information) to you before quitting. But by using
-  reflections, you are bypassing these checks because there is no way to check such errors during compile time. These 
-  uncaught errors may cause your program to fail during runtime instead, turning into runtime errors.
+  Compile time errors are easy to catch. Whenever you compile your code, the compiler cleverly spots any error you missed and points it out (along with line number and other useful information) to you before quitting. But by using reflections, you are bypassing these checks because there is no way to check such errors during compile time. These  uncaught errors may cause your program to fail during runtime instead, turning into runtime errors.
```
yamidark: >These  uncaught errors may cause your program to fail during runtime instead

Extra whitespace between `these` and `uncaught`.

contents/java/JavaReflections.md
```diff
@@ -177,105 +164,81 @@ An example implementation (not the actual) of the `Subscribe` annotation might b
...
-    @Subscribe
-    public void handleWeatherChangeEvent(WeatherChangeEvent event) {
-        if (event.weather == Weather.RAIN) {
-            hide();
-        }
+  private static final EventsCenter EVENTS_BUS = EventsCenter.getInstance();
+  
+  public Sheep() {
+    super();
+    EVENTS_BUS.register(this);
```
yamidark: I believe this should be `EVENTS_BUS.register(this.getClass())`, since the method asks for a `Class<?>` parameter and not `Object`.

contents/java/JavaReflections.md
```diff
@@ -1,173 +1,160 @@
...
+There are many webpages dedicated to explaining the details of reflections in Java; so this will not repeat what is being made readily available on the web. One good place to start is this [article by JavaWorld](http://www.javaworld.com/article/2077015/java-se/take-an-in-depth-look-at-the-java-reflection-api.html).
 
-One important point is that while Java reflections are powerful, its implementations are not very straightforward.
-There are however some libraries out there to make life easier such as the [Google's Guava
-library](https://github.com/google/guava/wiki/ReflectionExplained) which contains many utility methods and classes to
-make your life easier.
+One important point is that while Java reflections are powerful, its implementations are not very straightforward. There are however some libraries out there to make life easier such as the [Google's Guava
+library](https://github.com/google/guava/wiki/ReflectionExplained) which contains many utility methods and classes.
```
yamidark: Extra line break?

There are however some libraries out ~~there to make life easier~~ such as the [Google's Guava
library](https://github.com/google/guava/wiki/ReflectionExplained) which contains many utility methods and classes **that can make our life easier**.

yamidark (commented):
A few comments.

contents/java/JavaReflections.md
Zhiyuan-Amos: Sounds good! I will remove the word `important` though; seems unnecessary.

contents/java/JavaReflections.md
Zhiyuan-Amos: Will make the changes!

> We can also mention another disadvantage of Reflections from the NoSuchFieldException that we create dependencies(?) on the method/field names, such that if the class changes the method/field names our program stops working, and we need to update the name on our side too?

For this, I think I won't update it; it's somewhat covered [here](https://github.com/se-edu/learningresources/pull/30/files#diff-0b9124fd618f982f90c7723ef5b09389R223)

contents/java/JavaReflections.md
Zhiyuan-Amos: > Maybe we can explain more about method.invoke(sheep), if the method require parameters int then we also need to supply the arguments here or change the example here to take and supply some parameters.

I think there's no need to cos this is meant to be a simple guide? :)

Zhiyuan-Amos:
@yamidark free to review for me? :)

contents/java/JavaReflections.md
```diff
@@ -1,173 +1,163 @@
...
+And there you have it! What `Sheep` is really doing is to examine itself at runtime in order to obtain the `age` field inherited from `Animal`. This technique can be used in test cases to access private fields and methods in the class under test without modifying the visibility modifiers of the fields and methods in the class itself.
 
-        int age = getPrivateAge();
+You may notice that the `Sheep#getAge()` method sets the age `Field` object to be accessible and might wonder the implications. Fret not! The `Field#getDeclaredField()` actually returns a new `Field` instance - so you're just setting that particular local `Field` instance to be accessible, not the actual `age` field itself. You can read more about it in this [StackOverflow question](http://stackoverflow.com/questions/10638826/java-reflection-impact-of-setaccessibletrue).
 
-        if (age > 20) {
-            setHasWool();
-        }
-    }
+Do take note that two exceptions need to be handled when accessing private fields:
```
yamidark: when accessing ~~private~~ fields:

contents/java/JavaReflections.md
```diff
@@ -1,173 +1,163 @@
...
+```
 
-    // Other methods here...
+Alternatively, a much simpler and efficient way to perform this test is to set a value to the private field using Reflection:
+
+```java
+@Test
+public void foo() throws Exception {
+  Sheep sheep = new Sheep();
+  Field field = Sheep.class.getDeclaredField("age");
```
yamidark: This should be `getField()` instead since `Sheep` does not declare the `age` field (it is declared in `Animal`).
Maybe can add this note the difference between the 2 methods above.

contents/java/JavaReflections.md
```diff
@@ -1,173 +1,163 @@
...
+    age = 0;
+  }
 }
 ```
 
-And let us assume that you, for some reason, cannot modify this code. But you are interested in making a new class `Sheep`
-that extends `Animal` and do something when his age reaches certain threshold. But the annoying thing is that somebody
-decided that it is a good idea to make the age value `private` instead of `protected` in a top-level class such as this!
-So you cannot access the age of your `Sheep` even though it is an `Animal`.
+And let us assume that you, for some reason, cannot modify this code. But you are interested in making a new class `Sheep` that extends `Animal` and do something when his age reaches certain threshold. But the annoying thing is that somebody decided that it is a good idea to make the age value `private` instead of `protected` in a top-level class such as this! So you cannot access the age of your `Sheep` even though it is an `Animal`.
```
yamidark: Can add that the person did not provide a `getter` method for the `private` field too which is why we can't access the `age` (without Reflection).

contents/java/JavaReflections.md
```diff
@@ -1,173 +1,163 @@
...
-the implications. Fret not! The `Field#getDeclaredField()` actually returns a new `Field` instance - so you're just
-setting that particular local `Field` instance to be accessible, not the actual `age` field itself. You can read more
-about it in this [StackOverflow question](http://stackoverflow.com/questions/10638826/java-reflection-impact-of-setaccessibletrue).
+Suppose you want to perform a unit test for the method `getAge()`. However, you are only able to indirectly do so by testing `isProducingWool()`. This is not good as we are not able to directly verify the age of a sheep. However, with the help of Reflection, we can now test private methods.
+
+```java
+@Test
+public void foo() throws Exception {
+  Sheep sheep = new Sheep();
+  Method method = Sheep.class.getDeclaredMethod("getAge");
```
yamidark, line 149: Same as above.

contents/java/JavaReflections.md
```diff
@@ -177,105 +167,81 @@ An example implementation (not the actual) of the `Subscribe` annotation might b
...
 * Bad Security
 
-  The second example demostrated a way to access the private fields of a class using reflections.
-  This should be very concerning if your software deals with sensitive information because 
-  other classes can access fields that they are not supposed to.
+  The second example demostrated a way to access the private fields of a class using reflections. This should be very concerning if your software deals with sensitive information because other classes can access fields that they are not supposed to.
+
+* Indication of bad class design
+
+  Having to use reflection in order to bypass a class' encapsulation is usually indicative of an API design problem. We can remove the the usage of Reflection in the examples given [above](#accessing-private-fields) by adding a getter and setter method for `age`. See this [post](https://stackoverflow.com/questions/34571/how-do-i-test-a-private-function-or-a-class-that-has-private-methods-fields-or/34658#34658) for further discussion.
```
yamidark: Since we mentioned earlier that we can't touch the `Animal` class code for some reason, the alternate solution here doesn't really fit well here since the getter and setter method would most probably need to use Reflection too.

Maybe we can mention that we create our own implementation of `Animal` class with the getter and setter method.

Extra `the` in `We can remove the the usage of Reflection...`.

contents/java/JavaReflections.md
```diff
@@ -1,173 +1,163 @@
...
+#### Updating private fields
 
-        try {
-            Field f = animalClazz.getDeclaredField("age");
-            f.setAccessible(true);
-            return (int) f.get(this);
-        } catch (Exception e) {
-            e.printStackTrace();
-        }
+Suppose that you need to write tests for `Sheep`. As part of setting up the test, you need to create a sheep with age = 20. Suppose that the age of the sheep is updated automatically as time passes, whereby the age increases by 1 after every minute. A naive way of creating a sheep with age = 20 is to simply wait for 20minutes before performing the test:
```
yamidark: Missing whitespace in `20minutes`.

yamidark (commented):
Some minor comments.

contents/java/JavaReflections.md
Zhiyuan-Amos: Ah, this looks like a problem. `getField()` only works for public methods.

Edit: Oh we can just do `Animal.class.getDeclaredField("age")`. Also, since I'm not using `getField()` in this example, I think it's kinda random to suddenly bring up another similar API.

contents/java/JavaReflections.md
Zhiyuan-Amos: Sheep has the method `getAge()` :P

Zhiyuan-Amos:
@yamidark thanks for your review! :) I've made a patch in the latest commit.

yamidark (approved):
Looks good, have not other comments.

damithc (approved):

PR #31, opened 2018-02-24
Swift

Zhiyuan-Amos:


Zhiyuan-Amos:
@vivekscl can help review? ^^

contents/swift/welcome-to-swift.md
```diff
@@ -76,6 +74,34 @@ To learn more about Optionals, such as Optional Chaining and "dangerously" Force
 
 If you would like to seek a second (or more) opinion about Optionals, check out this [StackOverflow answer](http://stackoverflow.com/questions/24003642/what-is-an-optional-value-in-swift).
 
+## Guard Statement
+
+Notice that the golden path in the code above is indented:
```
rachx: It will be good to explain how `guard` statement works.
The code within a `guard` block is only executed if the condition is not true.

Maybe explicitly mention/show that `path` can be used outside the indented block but it doesn't work for `if`?

Also, you can mention how guard clauses are often use to enforce the pre-conditions of a method (related to fail fast and defensive programming)

contents/swift/welcome-to-swift.md
```diff
@@ -58,16 +58,14 @@ func yearAlbumReleased(name: String) -> Int? {
...
 func timeTravel(album: String) {
-    let year = yearAlbumReleased(album) // type of year: Int?
-    if let past = year {
-        // past is a normal (non-optional) Int value
-        // equal to the value stored in year
-        travelTo(past)
-    } else {
-        // year did not hold a value
-        liveInPresent()
+    let year: Int? = yearAlbumReleased(album)
```
rachx: This makes the type of year clearer but not sure if you want to use/make a note about type inference

contents/swift/welcome-to-swift.md
```diff
@@ -126,24 +174,67 @@ Next, we introduce the structs that conform to the protocols above.
 ```swift
 // Penguins can't fly ):
 struct Penguin: Bird {
-    let name: String
-    let canFly = false
+    var canFly: Bool {
```
rachx: Why do you have to make this a computed property?

contents/swift/welcome-to-swift.md
```diff
@@ -126,24 +174,67 @@ Next, we introduce the structs that conform to the protocols above.
...
 }
 ```
 
 And if you haven't noticed, protocols are extremely similar to interfaces in Java. 
 
 To understand more about POP, watching this [WWDC 2015 talk](https://www.youtube.com/watch?v=g2LwFZatfTI) is highly recommended.
 
+## Extensions
+
+Extensions add new functionality to an existing class, structure, enumeration, or protocol type. Using the example of `Eagle` above, we can segment the code neatly by using extensions to implement each protocol separately:
```
rachx: Perhaps, you can add some method examples (`func`)?

Can consider adding:
It is not possible to add stored property in extensions.
Another separate section on computed properties?

contents/swift/welcome-to-swift.md
```diff
@@ -102,6 +128,29 @@ You can think of Structs as a way to create instances that have their own unique
 
 If you wish to find out more, here is an [article](https://medium.com/capital-one-developers/reference-and-value-types-in-swift-de792db330b2) that explains the difference between the 2 types, as well as the benefits of value types and when to use them.
 
+## Enum
+
+Swift's enums can have associated values. For example:
```
rachx: Explain associated values and why will you use such enums?

rachx:
Gave an unasked-for review :P

contents/swift/welcome-to-swift.md
mauris: also to avoid pyramid of doom (nested if-if-if-if...)

contents/swift/welcome-to-swift.md
```diff
@@ -102,6 +128,29 @@ You can think of Structs as a way to create instances that have their own unique
...
+
+```swift
+enum Barcode {
+    case upc(Int, Int, Int, Int)
+    case qrCode(String)
+}
+
+let productBarcode = Barcode.qrCode("foo")
+
+switch productBarcode {
```
mauris: switch for enums in Swift must be exhaustive. might be good to mention this magic fact (Xcode would complain this and not compile)

contents/swift/welcome-to-swift.md
```diff
@@ -102,6 +128,29 @@ You can think of Structs as a way to create instances that have their own unique
...
+case let .upc(numberSystem, manufacturer, product, check):
+    print("UPC : \(numberSystem), \(manufacturer), \(product), \(check).") // no break statement required
+case let .qrCode(productCode):
+    print("QR code: \(productCode).")
+}
+// Prints "QR code: foo."
+```
+
+This feature is not supported in languages such as [Java](https://stackoverflow.com/questions/30044334/how-can-i-create-a-java-enum-with-associated-values-like-swift-enum) and using a workaround to implement enums with associated values results in code verbosity. Take a look at [Swift's documentation](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Enumerations.html) for more information about enums.
```
mauris: might want to be specific what "This feature" exactly is.

mauris (changes requested):

mauris:
Had some spare time so those are some of my thoughts. apologies if i'm disrupting your workflow

contents/swift/welcome-to-swift.md
vivekscl: Maybe you could explain this part a bit more? As for programmers new to Swift like me, it's hard to understand how it adds new functionality to the protocol type.
`add new functionality to an existing class, structure, enumeration, or protocol type`

contents/swift/welcome-to-swift.md
```diff
@@ -126,24 +174,67 @@ Next, we introduce the structs that conform to the protocols above.
...
+}
+    
+extension Eagle: Flyable {
+    var airspeedVelocity: Double {
+        return 160.0
+    }
+}
+```
+
+Extensions also allow us to extend define instance methods and type methods for types which you do not have access to the original source code. For example:
```
vivekscl: I'm not sure what `extend define instance methods` means. Did you mean extend **and** define?

vivekscl (commented):
I have minimal knowledge of Swift, so hopefully, my review from that perspective helps since this guide is an intro to Swift and its cool features :smile: .

contents/swift/welcome-to-swift.md
Zhiyuan-Amos: Ok, I've added this as a separate segment above so that I don't have to introduce a new concept of `type inference` here.

contents/swift/welcome-to-swift.md
Zhiyuan-Amos: > Another separate section on computed properties?

I'll just put a short note on this and refer the user to the Swift docs? :P

Zhiyuan-Amos:
@rachx @mauris @vivekscl THANKS FOR ALL OF YOUR REVIEWS! :) They are really helpful. I've addressed your comments and have added the commits as patches.

@rachx I was secretly hoping that you'll review for me actually. Thanks my CS3217 classmate 😃 
@mauris thanks too! So good to have CS3217 classmates to review for me 😆

mauris (approved):

mauris:
@Zhiyuan-Amos (: I'd take CS3216/17 again, but I got kicked out of school.

Zhiyuan-Amos:
> I'd take CS3216/17 again

I didn't want to take CS3216, that's why I'm in this mod ☺️ :P Thanks again! :))

rachx (approved):

contents/swift/welcome-to-swift.md
```diff
@@ -76,6 +83,48 @@ To learn more about Optionals, such as Optional Chaining and "dangerously" Force
 
 If you would like to seek a second (or more) opinion about Optionals, check out this [StackOverflow answer](http://stackoverflow.com/questions/24003642/what-is-an-optional-value-in-swift).
 
+## Guard Statement
```
damithc: Should we use plurals for all titles of this nature i.e., `Guard Statements` ?

contents/swift/welcome-to-swift.md
```diff
@@ -21,6 +21,15 @@ And if you are really pressed for time, here's a couple of cheatsheets with code
...
 
+## Type Inference
+
+Swift supports type inference whereby the compiler automatically deduces the type of a variable during compilation by examining the values assigned to it.
+
+```swift
+let str1: String = "foo"
+let str2 = "foo" // compiler infers that str2 is of type String
+```
+
```
damithc: Should we mention how this is different from _dynamic typing_ (e.g., as done in Python)?

contents/swift/welcome-to-swift.md
```diff
@@ -58,16 +67,14 @@ func yearAlbumReleased(name: String) -> Int? {
```
damithc, line 74: Although not part of this PR, this is the only level-3 title in this page and looks like can do without it.

contents/swift/welcome-to-swift.md
```diff
@@ -76,6 +83,48 @@ To learn more about Optionals, such as Optional Chaining and "dangerously" Force
 
 If you would like to seek a second (or more) opinion about Optionals, check out this [StackOverflow answer](http://stackoverflow.com/questions/24003642/what-is-an-optional-value-in-swift).
 
+## Guard Statement
+
+Notice that the golden path in the code above is indented:
```
damithc: Golden path is not something you can expect every reader to know. At least point to a resource that explains it. If we were using Markbind, we could have used to tool tip :-D

contents/swift/welcome-to-swift.md
```diff
@@ -76,6 +83,48 @@ To learn more about Optionals, such as Optional Chaining and "dangerously" Force
...
+    // past contains the non-nil value of year; proceed to do something with past
+    // past remains defined till the function exits
+}
+```
+
+Let's understand how the code above works:
+1. The code within a `guard` block is only executed if `year` contains a nil value.
+1. As the `guard` statement is used to transfer program control out of a scope, you must call one of the following functions within the `guard` block: `return`, `break`, `continue`, `throw`. As such, the `guard` statement is meant to enforce the pre-conditions of a method and to perform early return.
+
+Here are some of the benefits of using `guard` statement over `if-let` statement:
```
damithc, line 131: Reader is supposed to know what `if-let` means?

contents/swift/welcome-to-swift.md
```diff
@@ -102,6 +151,35 @@ You can think of Structs as a way to create instances that have their own unique
 
 If you wish to find out more, here is an [article](https://medium.com/capital-one-developers/reference-and-value-types-in-swift-de792db330b2) that explains the difference between the 2 types, as well as the benefits of value types and when to use them.
 
+## Enum
```
damithc: `Enums`?

contents/swift/welcome-to-swift.md
```diff
@@ -102,6 +151,35 @@ You can think of Structs as a way to create instances that have their own unique
...
+
+Swift's enums can have associated values. This enables you to store additional custom information along with the case value, and permits this information to vary each time you use that case in your code. For example, we can have an enum `Barcode` with case values `upc` and `qrCode`. We want to be able to distinguish within each value as each `upc` and `qrCode` can take on different values:
```
damithc: Unexplained term: `case value`

contents/swift/welcome-to-swift.md
```diff
@@ -102,6 +151,35 @@ You can think of Structs as a way to create instances that have their own unique
...
+    }
+}
+
+let barcode1 = Barcode.qrCode("foo")
+let barcode2 = Barcode.qrCode("bar")
+barcode1.printCode() // prints "QR code: foo."
+barcode2.printCode() // prints "QR code: bar."
+```
+
+Do note that every `switch` statement must be exhaustive; all possible values must be matched by one of the `switch` cases. Otherwise, you have to define a `default` case to handle any values that are unmatched. Also, notice that no `break` statement is required in between each `switch` case as Swift does not support implicit fallthrough. Take a look at [Swift's documentation on Control Flow](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/ControlFlow.html) to find out more.
```
damithc: >Do note that every `switch` statement must be exhaustive; all possible values must be matched by one of the `switch` cases. Otherwise, you have to define a `default` case to handle any values that are unmatched. Also, notice that no `break` statement is required in between each `switch` case as Swift does not support implicit fallthrough.

This kind of information is only useful if the document is supposed to _teach_ Swift.

contents/swift/welcome-to-swift.md
damithc: Don't assume the reader knows what an `enum` is. Explain what a typical enum is and then explain how Swift's enums are better.

damithc (changes requested):
Some suggestions added. Feel free to tidy up the remainder of the document too.

It would be useful to provide a series of resources someone can use to learn Swift from scratch or convert to Swift from another language but that can be a different PR altogether.

contents/swift/welcome-to-swift.md
Zhiyuan-Amos: > enum `Barcode` with case values `upc` and `qrCode`

It's unexplained, but I think it's kinda intuitive with the examples, and after you've seen the code sample below? ><

contents/swift/welcome-to-swift.md
Zhiyuan-Amos: Is this slight [change](https://github.com/se-edu/learningresources/pull/31/files#diff-c8f1811d71847d4fb3cc1ebbfdfc6955R75) sufficient?

Zhiyuan-Amos:
Ready for your review Prof @damithc.

contents/swift/welcome-to-swift.md
```diff
@@ -54,20 +70,18 @@ func yearAlbumReleased(name: String) -> Int? {
 }
 ```
 
-### Optional Binding
+## Optional Binding
```
damithc, line 73: Shouldn't this be just a continuation of the previous section?

damithc (approved):
LGTM 👍

contents/swift/welcome-to-swift.md
Zhiyuan-Amos: Hmm as there's a change in topic (the previous section was on returning an optional value, while this section is on unwrapping an optional value), I thought that we should have some sort of segmentation to make it clearer.

The section below (`Guard Statements`) can also be seen of as part of `Optional` because guard statements are only used in the context of unwrapping optionals. 

As such, I'm thinking if we should actually use `###` for both `Optional Binding` and `Guard Statements`.

PR #32, opened 2018-02-25
Update GPGPU with Jetson

bqnguyen94:


contents/gpgpu/gpgpu.md
```diff
@@ -48,6 +48,18 @@ The SIMT (Single Instruction, Multiple Threads) architecture of GPUs means that
 - CUDA: [Official website](http://www.nvidia.com/object/cuda_home_new.html)
 - OpenCL: [Official website](https://www.khronos.org/opencl/)
 
+### § 1.6 GPGPU in action
+The benefits of GPGPU are even more pronounced when it comes to embedded systems and Internet of Things (IoT) applications, where computing power is often compromised for physical space, weight, and power consumption. NVIDIA's Jetson is a computation processor board that delivers the processing capability of the Tegra GPU unit in a compact and light package; the latest version - Jetson TX2, measures at only 50 mm x 87 mm - the size of a business card<sup>[[6]](#footnote6)</sup>.
```
PierceAndy: **1. Point of sentence is not immediately obvious**
> NVIDIA's Jetson is a computation processor board that delivers the processing capability of the Tegra GPU unit in a compact and light package; the latest version - Jetson TX2, measures at only 50 mm x 87 mm - the size of a business card

As an introductory piece, brands and models don't carry any weight or meaning to the reader. This sentence introduces too many product names that aren't meaningful to the point of the sentence. I assume the point is about providing an example of one such benefit of GPGPU in embedded systems, where computing power is not compromised for space, based on the first sentence of the paragraph. 

In the first half of the sentence, the reader is thrown a bunch of product names out of the blue, and doesn't understand where this sentence is going (How are these Jetson and Tegra things tied to the benefits of GPGPU?), only to get your point at the end, requiring the reader to reread the entire sentence with that point in mind to make sense of the whole sentence.

It could be better rephrased to emphasize it being an example of your point. What do you think about rephrasing it this way instead?

> For instance, NVIDIA's Jetson TX2, a computation processor board that delivers the processing capability of the Tegra GPU unit in a compact and light package, measures at only 50 mm x 87 mm - the size of a business card.

I don't think TX2 being the newest Jetson product adds any value to the point, so it was omitted. 

**2. What's the significance of a Tegra GPU unit?**
What's significant about its processing capability? Is there someway to draw comparisons with another processor's performance in respect to its size?

**3. GPU unit?**
Graphics Processing Unit unit?

contents/gpgpu/gpgpu.md
```diff
@@ -48,6 +48,18 @@ The SIMT (Single Instruction, Multiple Threads) architecture of GPUs means that
...
+
+![NVIDIA's Jetson TX2](jetson_tx2.jpg "NVIDIA's Jetson TX2")
+
+Packed with a NVIDIA Pascal<sup>TM</sup> GPU with 256 CUDA cores at maximally 1300 MHz<sup>[[6]](#footnote6)</sup>, the TX2 is capable of parallel computing-intensive tasks such as real-time vision processing or deploying deep neural networks, allowing mobile platforms to overcome physical constraints and solve complex, real-world problems.
```
PierceAndy: **1. Why is Pascal here affixed with the trademark logo?**
Other trademarked or copyrighted terms on this page don't have the same treatment. 
We should try to be consistent in style.

**2. What's the significance of naming and specifying the GPU's specs?**
Once again, this ties back to point number 2 above. 
What's a Pascal GPU? Wasn't it supposed to be a Tegra GPU? 
What are these terms supposed to mean to a beginner?

**3. What physical constraints are overcome?**
Yes, it's small. But in comparison to what? What problem does it solve by being small? 
Smaller things can be more convenient to carry around, but convenience doesn't necessarily mean a constraint is overcome.

**4. I like how you included examples of parallel compute-intensive tasks and images**

**5. You should also update the author's information at the top of this article**

PierceAndy (commented):
Thanks for contributing! I like how you brought up IoT's relevance to GPGPU.

Here are a few comments and questions I have:

contents/gpgpu/gpgpu.md
bqnguyen94: <b>1.</b> Yeah I think your recommendation is better. My English is cluttered with jargons haha.
<b>2.</b> Tegra is the SoC, containing the CPU and GPU. I omitted it entirely and just talked about the Pascal GPU.
<b>3.</b> My mistake :P

contents/gpgpu/gpgpu.md
bqnguyen94: <b>1.</b> On NVIDIA's website they always put the term on Pascal, as it's the name of their GPU architecture. I just removed it.
<b>2.</b> I wanted to say that the Pascal GPU is capable of parallel computing, but couldn't find a better evidence for it.

PierceAndy: 2. I think reducing the number of jargon in the earlier section, and clarify others made it easier to understand this part. If you added a small part clarifying the significance of CUDA cores (`CUDA cores are parallel processors` or something along this line), it may help convey your point of the Pascal GPU being capable of parallel computing. Thoughts?

PierceAndy:
This PR is mostly fine in its current iteration. Let me know when you're done by requesting a review from me, and I'll request a final review from Damith if you don't have anyone else you wanna request reviews from.

contents/gpgpu/gpgpu.md
```diff
@@ -2,6 +2,8 @@
 
 ###### Author: [Pierce Anderson Fu](https://github.com/PierceAndy)
 
+###### Author: [Nguyen Quoc Bao](https://github.com/bqnguyen94)
```
PierceAndy: You could combine the two lines of authors. See: https://github.com/se-edu/learningresources/pull/37/files#diff-e237ae1cc0dfe433d7b40fab5d5775c8R3

contents/gpgpu/gpgpu.md
```diff
@@ -48,6 +50,18 @@ The SIMT (Single Instruction, Multiple Threads) architecture of GPUs means that
 - CUDA: [Official website](http://www.nvidia.com/object/cuda_home_new.html)
 - OpenCL: [Official website](https://www.khronos.org/opencl/)
 
+### § 1.6 GPGPU in action
```
PierceAndy, line 52: I almost missed this, but to be consistent, it would be great to update the table of contents at the top of the document too.

PierceAndy (approved):

PierceAndy:
@damithc PR looks good to me. For your final review, sir.

damithc (changes requested):
Can cite the source of images?

bqnguyen94:
@damithc done 👍

damithc:
Need a line break after the image here?
image

bqnguyen94:
@damithc done 👍

damithc:
> @damithc PR looks good to me. For your final review, sir.

Thanks for the reviews @PierceAndy 💯

PR #33, opened 2018-02-25
Update Scalability by adding more ways to scale

vivekscl:


contents/scalability/scalable-development.md
```diff
@@ -33,9 +33,29 @@ There are a few common bottlenecks that can be avoided such as database querying
...
 
-Performance bottlenecks ([resource](https://www.apicasystem.com/blog/5-common-performance-bottlenecks/))
+Performance bottlenecks ([resource](https://www.apicasystem.com/blog/5-common-performance-bottlenecks))
 
-Non-Relational Database ([resource](http://www.jamesserra.com/archive/2015/08/relational-databases-vs-non-relational-databases/)) ([resource](https://www.pluralsight.com/blog/software-development/relational-non-relational-databases))
+Non-Relational Database ([resource](http://www.jamesserra.com/archive/2015/08/relational-databases-vs-non-relational-databases)) ([resource](https://www.pluralsight.com/blog/software-development/relational-non-relational-databases))
+
+### Caching
+When more and more data is fetched from the database, the response becomes slower. This is where caching comes in. A cache is a key-value store that resides between the application and the database. By retrieving data from the cache instead of the database in the back-end server, the response time reduces greatly.
```
Zhiyuan-Amos: Isn't clear why the response becomes slower as more and more data is fetched from database. Also, perhaps can be more explicit on where the cache is being stored.

contents/scalability/scalable-development.md
```diff
@@ -33,9 +33,29 @@ There are a few common bottlenecks that can be avoided such as database querying
...
+
+The next thing to consider is what to cache. The rule of thumb is to cache data that is frequently accessed and read-mostly, meaning that an entry is read more often than it is updated. With this, cache hits would be more often than cache misses and as a result, the speed increase would outweigh the cost of maintaining the cache.
```
Zhiyuan-Amos: > The rule of thumb is to cache data that is frequently accessed and ~~read-mostly, meaning that an entry is~~ read more often than it is updated.

contents/scalability/scalable-development.md
```diff
@@ -33,9 +33,29 @@ There are a few common bottlenecks that can be avoided such as database querying
...
+When more and more data is fetched from the database, the response becomes slower. This is where caching comes in. A cache is a key-value store that resides between the application and the database. By retrieving data from the cache instead of the database in the back-end server, the response time reduces greatly.
+
+The next thing to consider is what to cache. The rule of thumb is to cache data that is frequently accessed and read-mostly, meaning that an entry is read more often than it is updated. With this, cache hits would be more often than cache misses and as a result, the speed increase would outweigh the cost of maintaining the cache.  
+
+Caching ([resource](http://www.lecloud.net/post/9246290032/scalability-for-dummies-part-3-cache))
+
+Scalability best practices ([resource](https://www.infoq.com/articles/ebay-scalability-best-practices))
+
+### Go Asynchronous 
+Unlike synchronous operations, which runs sequentially and waits for the previous operation to complete before moving on to the next, asynchronous operations don't wait for other operations to finish before proceeding. Instead, they continue and allow other operations to finish.
```
Zhiyuan-Amos: > asynchronous operations don't wait for other operations to finish before proceeding. Instead, they continue and allow other operations to finish.

Rephrase: Asynchronous operations run in parallel, so they do not wait for prior operations to finish before running.

contents/scalability/scalable-development.md
```diff
@@ -33,9 +33,29 @@ There are a few common bottlenecks that can be avoided such as database querying
...
+
+By using asynchronous operations, components are less tightly coupled and can be scaled independently of one another. As a result, if one component is down, the other component can still move forward.
```
Zhiyuan-Amos: > By using asynchronous operations, components are less tightly coupled and can be scaled independently of one another. As a result, if one component is down, the other component can still move forward.

Quite unclear for a newb like me.

contents/scalability/scalable-development.md
```diff
@@ -33,9 +33,29 @@ There are a few common bottlenecks that can be avoided such as database querying
...
+
+In situations in which responding to request is crucial, asynchronous operations can reduce the latency experienced by the requester. This is done by prioritising the quickness of the response to the user over the speed in which other processes complete.
```
Zhiyuan-Amos: I find this segment kinda unclear too ><

contents/scalability/scalable-development.md
```diff
@@ -1,6 +1,6 @@
 # Scalable Development - An Introduction
 
-Authors: Loh Jia Shun Kenneth
+Authors: Loh Jia Shun Kenneth, Vivek Lakshmanan
 
 ## Scalable Development
 When Pokémon Go first launched in July 6, the heavy traffic from users caused its servers to crash. However, as of 2017, Google handles at least 3.5 billion searches daily, Whatsapp handles at least 10 billion messages a day, and Facebook hosts 1.2 billion daily active users. What makes these companies different?
```
junkiattan: I know this part was not touched in the commit, but the example of Pokémon Go didn't link very well to the point. Is it because the heavy traffic far exceeded that of Facebook/Google, or because their infrastructure was not scalable? Furthermore no numbers were given so it is are to link it to the point

contents/scalability/scalable-development.md
```diff
@@ -33,9 +33,29 @@ There are a few common bottlenecks that can be avoided such as database querying
 
 There are many program profilers available that you can use to find the bottlenecks in your server. By finding out where the server spends most computational time in, you can potentially save huge amounts of time.
 
-Performance bottlenecks ([resource](https://www.apicasystem.com/blog/5-common-performance-bottlenecks/))
+Performance bottlenecks ([resource](https://www.apicasystem.com/blog/5-common-performance-bottlenecks))
 
-Non-Relational Database ([resource](http://www.jamesserra.com/archive/2015/08/relational-databases-vs-non-relational-databases/)) ([resource](https://www.pluralsight.com/blog/software-development/relational-non-relational-databases))
```
junkiattan: I think it would be more readable if you number the points.

1. Performance bottlenecks
1. Non-relational database

Furthermore, consider describing the resources you link in the page so that readers can have an idea of what the link is about without needing to break the flow of reading the rest of the webpage

contents/scalability/scalable-development.md
junkiattan: Consider replacing

> Whenever more data is fetched from a database, it's response becomes slower.

Further, it would be good to explain the reason why response is slower, and therefore how caching can come in and help solve the issue

contents/scalability/scalable-development.md
```diff
@@ -33,9 +33,29 @@ There are a few common bottlenecks that can be avoided such as database querying
...
+Non-Relational Database ([resource](http://www.jamesserra.com/archive/2015/08/relational-databases-vs-non-relational-databases)) ([resource](https://www.pluralsight.com/blog/software-development/relational-non-relational-databases))
+
+### Caching
+When more and more data is fetched from the database, the response becomes slower. This is where caching comes in. A cache is a key-value store that resides between the application and the database. By retrieving data from the cache instead of the database in the back-end server, the response time reduces greatly.
+
+The next thing to consider is what to cache. The rule of thumb is to cache data that is frequently accessed and read-mostly, meaning that an entry is read more often than it is updated. With this, cache hits would be more often than cache misses and as a result, the speed increase would outweigh the cost of maintaining the cache.  
+
+Caching ([resource](http://www.lecloud.net/post/9246290032/scalability-for-dummies-part-3-cache))
```
junkiattan: Consider providing descriptions to the links

contents/scalability/scalable-development.md
```diff
@@ -33,9 +33,29 @@ There are a few common bottlenecks that can be avoided such as database querying
...
+Scalability best practices ([resource](https://www.infoq.com/articles/ebay-scalability-best-practices))
+
+### Go Asynchronous 
+Unlike synchronous operations, which runs sequentially and waits for the previous operation to complete before moving on to the next, asynchronous operations don't wait for other operations to finish before proceeding. Instead, they continue and allow other operations to finish.
+
+By using asynchronous operations, components are less tightly coupled and can be scaled independently of one another. As a result, if one component is down, the other component can still move forward.
+
+In situations in which responding to request is crucial, asynchronous operations can reduce the latency experienced by the requester. This is done by prioritising the quickness of the response to the user over the speed in which other processes complete.
+
+High level overview of asynchronous processes ([resource](https://www.youtube.com/watch?time_continue=2&v=8aGhZQkoFbQ))
```
junkiattan: Same here for the links

contents/scalability/scalable-development.md
junkiattan: Does this mean that the other process you are referring to is in no way related to generating the response? Ultimately if you reply on the processes to generate a whole response, will the quickness of some processes matter?

Just some questions I have for someone who doesn't really know about these kind of operations. Addressing these would help make your paragraphs clearer

vivekscl: >Does this mean that the other process you are referring to is in no way related to generating the response?

They are related but, some processes don't affect the user. For instance, tracking the user's activity for recommending items in the future doesn't affect the user at the current moment. So the UI can be updated before processing this.

vivekscl:
@Zhiyuan-Amos @junkiattan Thanks for your reviews! I've made the requested changes and hopefully its more clear now! :smile:

contents/scalability/scalable-development.md
```diff
@@ -33,29 +33,29 @@ There are a few common bottlenecks that can be avoided such as database querying
...
+1. Non-Relational Database ([resource](http://www.jamesserra.com/archive/2015/08/relational-databases-vs-non-relational-databases)) ([resource](https://www.pluralsight.com/blog/software-development/relational-non-relational-databases))
 
 ### Caching
-When more and more data is fetched from the database, the response becomes slower. This is where caching comes in. A cache is a key-value store that resides between the application and the database. By retrieving data from the cache instead of the database in the back-end server, the response time reduces greatly.
+With an increasing user base, your server has to deal with a larger number of requests along with the bottlenecks mentioned previously such as network congestion and database querying. As a result, the response becomes slower. This is where caching comes inorder to meet this demand. A cache is a key-value store that resides between the application and the database which can either be in the browser or part of your server infrastructure itself. By retrieving data from the cache instead of the database, the response time reduces greatly.
```
Zhiyuan-Amos: missing whitespace `inorder`

contents/scalability/scalable-development.md
```diff
@@ -33,29 +33,29 @@ There are a few common bottlenecks that can be avoided such as database querying
...
 
-The next thing to consider is what to cache. The rule of thumb is to cache data that is frequently accessed and read-mostly, meaning that an entry is read more often than it is updated. With this, cache hits would be more often than cache misses and as a result, the speed increase would outweigh the cost of maintaining the cache.  
+The next thing to consider is what to cache. The rule of thumb is to cache data that is frequently accessed and read more often than it is updated. With this, cache hits would be more often than cache misses and as a result, the speed increase would outweigh the cost of maintaining the cache.
```
Zhiyuan-Amos: > the speed increase would outweigh the cost of maintaining the cache.  

Hmm I'm thinking we should rephrase this. Above we use the term `response time` but we use the term `speed` here instead. So, perhaps we should say `the time saved from faster accesses and reads outweighs the extra time taken to populate the cache.`

contents/scalability/scalable-development.md
```diff
@@ -33,29 +33,29 @@ There are a few common bottlenecks that can be avoided such as database querying
...
 
-The next thing to consider is what to cache. The rule of thumb is to cache data that is frequently accessed and read-mostly, meaning that an entry is read more often than it is updated. With this, cache hits would be more often than cache misses and as a result, the speed increase would outweigh the cost of maintaining the cache.  
+The next thing to consider is what to cache. The rule of thumb is to cache data that is frequently accessed and read more often than it is updated. With this, cache hits would be more often than cache misses and as a result, the speed increase would outweigh the cost of maintaining the cache.  
 
-Caching ([resource](http://www.lecloud.net/post/9246290032/scalability-for-dummies-part-3-cache))
+1. A beginner's guide to scalability using caches which also describes in more detail on what to cache and how to do so ([resource](http://www.lecloud.net/post/9246290032/scalability-for-dummies-part-3-cache))
```
Zhiyuan-Amos: This seems too long. `Scalability using caches`?

contents/scalability/scalable-development.md
```diff
@@ -33,29 +33,29 @@ There are a few common bottlenecks that can be avoided such as database querying
...
 
 ### Go Asynchronous 
-Unlike synchronous operations, which runs sequentially and waits for the previous operation to complete before moving on to the next, asynchronous operations don't wait for other operations to finish before proceeding. Instead, they continue and allow other operations to finish.
+Unlike synchronous operations, which runs sequentially and waits for the previous operation to complete before moving on to the next, asynchronous operations run in parallel, so they do not wait for prior operations to finish before running. Instead, they continue and allow other operations to finish.
 
-By using asynchronous operations, components are less tightly coupled and can be scaled independently of one another. As a result, if one component is down, the other component can still move forward.
+If one component A calls another component B synchronously, they are tightly coupled as both components depend on each other. As a result, to scale A, you must also scale B. By using asynchronous operations, components are decoupled and can be scaled independently of one another.
```
Zhiyuan-Amos: > they are tightly coupled as both components depend on each other.

This is a tautology :P tightly coupled == depend on each other.

contents/scalability/scalable-development.md
```diff
@@ -33,29 +33,29 @@ There are a few common bottlenecks that can be avoided such as database querying
...
 
-In situations in which responding to request is crucial, asynchronous operations can reduce the latency experienced by the requester. This is done by prioritising the quickness of the response to the user over the speed in which other processes complete.
+In situations in which responding to request is crucial, asynchronous operations can reduce the latency experienced by the requester. This is done by prioritising the quickness of the response to the user over the speed of other processes such as execution latency (how quickly we get everything done). For instance, rather than waiting for some processes such as tracking the user's activity for recommending products in the near future to finish, it's better to asynchronously update the user interface and then finish these processes. In this way, the user experiences lesser latency.
```
Zhiyuan-Amos: > such as execution latency (how quickly we get everything done)

Is `(how quickly the request is done executing)` better?

> For instance, rather than waiting for some processes such as tracking the user's activity for recommending products in the near future to finish

Maybe we can use a simpler example, such as `downloading file`? So while the file downloads, your ui doesn't hang.

contents/scalability/scalable-development.md
junkiattan: What you said makes sense and I can understand. Use this example and tone to make your section clearer

contents/scalability/scalable-development.md
```diff
@@ -1,9 +1,9 @@
 # Scalable Development - An Introduction
 
-Authors: Loh Jia Shun Kenneth
+Authors: Loh Jia Shun Kenneth, Vivek Lakshmanan
 
 ## Scalable Development
-When Pokémon Go first launched in July 6, the heavy traffic from users caused its servers to crash. However, as of 2017, Google handles at least 3.5 billion searches daily, Whatsapp handles at least 10 billion messages a day, and Facebook hosts 1.2 billion daily active users. What makes these companies different?
+When Pokémon Go first launched in 2016, the heavy traffic from users caused its servers to crash as the server was built to handle a expected maximum of five times the storage size instead of the reality which was fifty times the initial estimate. However, as of 2017, Google handles at least 3.5 billion searches daily, Whatsapp handles at least 10 billion messages a day, and Facebook hosts 1.2 billion daily active users. What makes these companies different?
```
crphang: `to handle a expected maximum` -> `to handle an expected maximum`

contents/scalability/scalable-development.md
```diff
@@ -33,9 +33,29 @@ There are a few common bottlenecks that can be avoided such as database querying
 
 There are many program profilers available that you can use to find the bottlenecks in your server. By finding out where the server spends most computational time in, you can potentially save huge amounts of time.
 
-Performance bottlenecks ([resource](https://www.apicasystem.com/blog/5-common-performance-bottlenecks/))
+1. More details on the common performance bottlenecks ([resource](https://www.apicasystem.com/blog/5-common-performance-bottlenecks))
 
-Non-Relational Database ([resource](http://www.jamesserra.com/archive/2015/08/relational-databases-vs-non-relational-databases/)) ([resource](https://www.pluralsight.com/blog/software-development/relational-non-relational-databases))
+1. Non-Relational Database ([resource](http://www.jamesserra.com/archive/2015/08/relational-databases-vs-non-relational-databases)) ([resource](https://www.pluralsight.com/blog/software-development/relational-non-relational-databases))
```
crphang: Just having the word `Non-Relational Database` seems weird, maybe have it more descriptive like above when adding the link?

contents/scalability/scalable-development.md
```diff
@@ -33,9 +33,29 @@ There are a few common bottlenecks that can be avoided such as database querying
...
+
+The next thing to consider is what to cache. The rule of thumb is to cache data that is frequently accessed and read more often than it is updated. With this, cache hits would be more often than cache misses and as a result, the speed increase would outweigh the cost of maintaining the cache.  
+
+1. A beginner's guide to scalability using caches which also describes in more detail on what to cache and how to do so ([resource](http://www.lecloud.net/post/9246290032/scalability-for-dummies-part-3-cache))
+
+1. Scalability best practices ([resource](https://www.infoq.com/articles/ebay-scalability-best-practices))
+
+### Go Asynchronous 
+Unlike synchronous operations, which runs sequentially and waits for the previous operation to complete before moving on to the next, asynchronous operations run in parallel, so they do not wait for prior operations to finish before running. Instead, they continue and allow other operations to finish.
```
crphang: `So they do not wait for prior operations to finish before running`. I think it's more like, other operations do not have to wait for asynchronous operations to finish before they run? Your next statement makes it clearer.

Also strictly speaking, asynchronous operations do not need to run in parallel e.g. Javascript implementation of async operations (even though it feels like it is running in parallel). To be more precise you can say it is `non-blocking`.

crphang (changes requested):
Some minor comments. Overall great job!

contents/scalability/scalable-development.md
darrenwee: I think that you should elaborate more on what an asynchronous operation is (this is related to the bigger issue of the subtle difference between concurrency and parallelism, which are, while related, not quite the same thing). I do not think that it is a good idea to say

> asynchronous operations run in parallel

because it adds to the confusion over the nebulous concept of asynchronicity and parallelism. Perhaps you could rewrite it to something like

> asynchronous operations are non-blocking and operate independently of each other

contents/scalability/scalable-development.md
```diff
@@ -33,9 +33,29 @@ There are a few common bottlenecks that can be avoided such as database querying
 
 There are many program profilers available that you can use to find the bottlenecks in your server. By finding out where the server spends most computational time in, you can potentially save huge amounts of time.
 
-Performance bottlenecks ([resource](https://www.apicasystem.com/blog/5-common-performance-bottlenecks/))
+1. More details on the common performance bottlenecks ([resource](https://www.apicasystem.com/blog/5-common-performance-bottlenecks))
 
-Non-Relational Database ([resource](http://www.jamesserra.com/archive/2015/08/relational-databases-vs-non-relational-databases/)) ([resource](https://www.pluralsight.com/blog/software-development/relational-non-relational-databases))
+1. Non-Relational Database ([resource](http://www.jamesserra.com/archive/2015/08/relational-databases-vs-non-relational-databases)) ([resource](https://www.pluralsight.com/blog/software-development/relational-non-relational-databases))
+
+### Caching
```
darrenwee: You may also wish to discuss [sharding](https://stackoverflow.com/questions/992988/what-is-sharding-and-why-is-it-important) since it's more common nowadays and an alternative to caching.

darrenwee (commented):
I guess I was too late to the party so I only have content suggestions. I don't know if it's within your scope, but you could also dedicate a section on challenges that only arise as you scale up. For example, how to synchronize large, distributed systems, e.g. like how changes are propagated through a large CDN, or how to perform load balancing to prevent individual nodes from being overloaded.

contents/scalability/scalable-development.md
vivekscl: >Is (how quickly the request is done executing) better?

`request is done executing` sounds a bit off though.  Maybe `how quickly the request is processed`?

contents/scalability/scalable-development.md
vivekscl: Adding a section on sharding is a great idea! However, the section on challenges doesn't fall within my scope just yet. Maybe I'll work on it in another PR! Thanks for the content suggestions :+1: .

vivekscl:
@damithc Ready for Review :smile: .

vivekscl:
@Crphang maybe you wanna take a look since you requested changes? :smile:

contents/scalability/scalable-development.md
```diff
@@ -33,9 +33,40 @@ There are a few common bottlenecks that can be avoided such as database querying
...
+1. More details on the common performance bottlenecks ([resource](https://www.apicasystem.com/blog/5-common-performance-bottlenecks))
 
-Non-Relational Database ([resource](http://www.jamesserra.com/archive/2015/08/relational-databases-vs-non-relational-databases/)) ([resource](https://www.pluralsight.com/blog/software-development/relational-non-relational-databases))
+1. Relational vs Non-Relational Database ([resource](http://www.jamesserra.com/archive/2015/08/relational-databases-vs-non-relational-databases)) ([resource](https://www.pluralsight.com/blog/software-development/relational-non-relational-databases))
+
+### Caching
+With an increasing user base, your server has to deal with a larger number of requests along with the bottlenecks mentioned previously such as network congestion and database querying. As a result, the response becomes slower. This is where caching comes in order to meet this demand. A cache is a key-value store that resides between the application and the database which can either be in the browser or part of your server infrastructure itself. By retrieving data from the cache instead of the database, the response time reduces greatly.
+
```
damithc: It's not fully clear from this text why retrieving from cache is faster than retrieving from the database.

contents/scalability/scalable-development.md
```diff
@@ -33,9 +33,40 @@ There are a few common bottlenecks that can be avoided such as database querying
...
+Sharding is essential when your dataset becomes too large to store in a single database. It reduces the number of rows in each table and as such improves search performance since the search is done on a smaller table. 
+
+Querying the databases containing only the relevant partitions becomes possible as well. For instance, if a database contains a column for age, then you can partition the rows according to an age group and store them in different databases. Whenever there is a need to access the data of a particular age group, instead of querying the whole database, you just need to query the partition that contains that age group. This allows your database to scale along with your data and traffic growth.
```
damithc: If there are two benefits, it's clearer if you gave them as a numbered list or a bullet list rather than two paragraphs.

damithc: The term `partition` is being used without explanation?

damithc: I think the two points here are a bit tangled with each other; you can try to tease them apart.
1. sharding can store more data than a single database can hold
2. sharding can make queries faster

contents/scalability/scalable-development.md
```diff
@@ -33,9 +33,40 @@ There are a few common bottlenecks that can be avoided such as database querying
...
+
+1. How Sharding Works ([resource](https://medium.com/@jeeyoungk/how-sharding-works-b4dec46b3f6))
+
+1. A StackOverflow post on Sharding ([resource](https://stackoverflow.com/questions/992988/what-is-sharding-and-why-is-it-important))
+
+### Go Asynchronous 
+Unlike synchronous operations, which runs sequentially and waits for the previous operation to complete before moving on to the next, asynchronous operations are non-blocking and operate independently of each other instead. As a result other operations do not have to wait for asynchronous operations to finish.
```
damithc: the meaning of `non-blocking` is not explained

contents/scalability/scalable-development.md
```diff
@@ -33,9 +33,40 @@ There are a few common bottlenecks that can be avoided such as database querying
...
+
+If one component A calls another component B synchronously, they are tightly coupled. As a result, to scale A, you must also scale B. By using asynchronous operations, components are decoupled and can be scaled independently of one another.
```
damithc: Hard to follow this argument. Do you need to mention `coupling` at all? What do you mean by 'scaling a component'?

contents/scalability/scalable-development.md
```diff
@@ -33,9 +33,40 @@ There are a few common bottlenecks that can be avoided such as database querying
...
+
+In situations in which responding to request is crucial, asynchronous operations can reduce the latency experienced by the requester. This is done by prioritising the quickness of the response to the user over the speed of other processes such as execution latency (how quickly the request is processed). For instance, rather than waiting for some processes such as downloading the requested file to finish, it's better to asynchronously update the user interface and then finish these processes. In this way, the user experiences lesser latency since the user interface is updated instantly instead of appearing to have crashed due to waiting for the downloaded file.
+
```
damithc: 🤔 I wonder if this is actually related to scaling. Sounds more like a lag/response-time issue.

contents/scalability/scalable-development.md
```diff
@@ -33,9 +33,40 @@ There are a few common bottlenecks that can be avoided such as database querying
...
+### Caching
+With an increasing user base, your server has to deal with a larger number of requests along with the bottlenecks mentioned previously such as network congestion and database querying. As a result, the response becomes slower. This is where caching comes in order to meet this demand. A cache is a key-value store that resides between the application and the database which can either be in the browser or part of your server infrastructure itself. By retrieving data from the cache instead of the database, the response time reduces greatly.
+
+The next thing to consider is what to cache. The rule of thumb is to cache data that is frequently accessed and read more often than it is updated. With this, cache hits would be more often than cache misses and as a result, the time saved from faster accesses and reads outweighs the extra time taken to populate the cache.  
+
+1. Scalability using caches ([resource](http://www.lecloud.net/post/9246290032/scalability-for-dummies-part-3-cache))
+
+1. Scalability best practices ([resource](https://www.infoq.com/articles/ebay-scalability-best-practices))
+
+### Sharding
```
damithc: I wonder if this topic is at the same level as the other topics. Sharing is a specific way of distributing data over multiple databases. In fact, optimizing the data storage is already mentioned under `Avoiding Bottlenecks`

contents/scalability/scalable-development.md
```diff
@@ -33,9 +33,40 @@ There are a few common bottlenecks that can be avoided such as database querying
...
+
+Sharding is essential when your dataset becomes too large to store in a single database. It reduces the number of rows in each table and as such improves search performance since the search is done on a smaller table. 
+
+Querying the databases containing only the relevant partitions becomes possible as well. For instance, if a database contains a column for age, then you can partition the rows according to an age group and store them in different databases. Whenever there is a need to access the data of a particular age group, instead of querying the whole database, you just need to query the partition that contains that age group. This allows your database to scale along with your data and traffic growth.
+
+1. How Sharding Works ([resource](https://medium.com/@jeeyoungk/how-sharding-works-b4dec46b3f6))
+
+1. A StackOverflow post on Sharding ([resource](https://stackoverflow.com/questions/992988/what-is-sharding-and-why-is-it-important))
+
+### Go Asynchronous
```
damithc: Similarly, this topic too seems to be a specific way of avoiding bottlenecks. Overall, you need to rethink if the new topics introduced are in fact subsumed under the existing topics.

damithc (changes requested):
The content you added are useful; just now sure if they are added at the right headings. In addition to my comments, consider addressing 'Horizontal vs Vertical Scaling' as a new heading. It's important enough to be a top level heading in this chapter.

contents/scalability/scalable-development.md
crphang: You might want to remove (operate independenly of each other instead). It suggests that synchronous tasks are dependent, while asynchronous tasks are independent. But it could not be the case. Synchronous task can be sequentially independent of each other, while asynchronous task could be logically dependent (just not sequentially). You might just want to elaborate further on non-blocking.

crphang (commented):
Minor Comments

contents/scalability/scalable-development.md
vivekscl: >The term partition is being used without explanation?

In this case, partition just means to divide into parts. Since it's not really a technical term, there's no need for an explanation right?

contents/scalability/scalable-development.md
vivekscl: > What do you mean by 'scaling a component'?

By that, I meant if component A has to be modified to deal with higher traffic to component A only, then component B has to be modified as well since A is tightly coupled to B. If asynchronous programming is done, then they are decoupled and only A has to be modified to deal with the higher traffic instead of modifying a component that is not affected by the traffic growth.

But then again if this concept is more confusing than anything, I should remove it instead?

contents/scalability/scalable-development.md
vivekscl: So it is more of a performance issue right? In that case, if I were to place this under the `Avoid Bottlenecks` section, it would make more sense?

vivekscl:
@damithc @Crphang Ready for Review.

[view the rendered version here](https://github.com/vivekscl/learningresources/blob/19aa14c1e90f166ce7d1c2e26f5364cf8c8a3e07/contents/scalability/scalable-development.md)

damithc (commented):
i'm ok with the structure. I'll wait for a peer to approve before I do a detailed read.

vivekscl:
@Zhiyuan-Amos wanna take a look? :smile:

contents/scalability/scalable-development.md
```diff
@@ -1,9 +1,9 @@
 # Scalable Development - An Introduction
 
-Authors: Loh Jia Shun Kenneth
+Authors: Loh Jia Shun Kenneth, Vivek Lakshmanan
 
 ## Scalable Development
-When Pokémon Go first launched in July 6, the heavy traffic from users caused its servers to crash. However, as of 2017, Google handles at least 3.5 billion searches daily, Whatsapp handles at least 10 billion messages a day, and Facebook hosts 1.2 billion daily active users. What makes these companies different?
+When Pokémon Go first launched in 2016, the heavy traffic from users caused its servers to crash as the server was built to handle an expected maximum of five times the storage size instead of the reality which was fifty times the initial estimate. However, as of 2017, Google handles at least 3.5 billion searches daily, Whatsapp handles at least 10 billion messages a day, and Facebook hosts 1.2 billion daily active users. What makes these companies different?
```
Zhiyuan-Amos: > server was built to handle an expected maximum of five times the storage size instead of the reality which was fifty times the initial estimate.

What does "storage size" refer to? I can't quite get what this sentence is saying because it talks about "five times the storage size" vs "fifty times the initial estimate", which seems to be comparing about different things.

contents/scalability/scalable-development.md
```diff
@@ -15,12 +15,33 @@ So what is scalability? Scalability is the capacity to which a system can cope w
 
 Let's say you have a server running on a single host, but due to the growing number of users, it has been slowing down. What are the different ways you can speed it up?
 
-### Get Better Hardware
+### Horizontal vs Vertical Scaling
+
+#### Horizontal Scaling
+
+Buy more machines to host your server. This is called horizontal scaling. This method can help improve scalability by a huge factor, as each additional unit of processing power and RAM costs the same, no matter how many machines you buy. As this is a promising way to increase scalability, it is no surprise why large services such as Google and Facebook use thousands of computer servers just to provide services to all their users.
```
Zhiyuan-Amos: > Buy more machines to host your server. This is called horizontal scaling.

Unnecessary verbosity :p. "Horizontal scaling is simply buying more machines to host your server."

contents/scalability/scalable-development.md
```diff
@@ -15,12 +15,33 @@ So what is scalability? Scalability is the capacity to which a system can cope w
...
+
+However, to fully reap the benefits of multiple machines, you will need to learn various techniques to modify your server infrastructure. In most cases you cannot simply run multiple instances of the server and expect it to work.
+
+Here are some tips to get started with horizontal scaling:
+- Understand parallel computing and how to use it to speed up computation.
+- Learn how to utilize MapReduce on [Hadoop](https://www.tutorialspoint.com/hadoop/index.htm). The Hadoop ecosystem is so versatile and widely-used that there are companies offering Hadoop as a service, so you do not have to set up the server yourself
```
Zhiyuan-Amos: Missing period.

contents/scalability/scalable-development.md
```diff
@@ -15,12 +15,33 @@ So what is scalability? Scalability is the capacity to which a system can cope w
...
+### Horizontal vs Vertical Scaling
+
+#### Horizontal Scaling
+
+Buy more machines to host your server. This is called horizontal scaling. This method can help improve scalability by a huge factor, as each additional unit of processing power and RAM costs the same, no matter how many machines you buy. As this is a promising way to increase scalability, it is no surprise why large services such as Google and Facebook use thousands of computer servers just to provide services to all their users.
+
+However, to fully reap the benefits of multiple machines, you will need to learn various techniques to modify your server infrastructure. In most cases you cannot simply run multiple instances of the server and expect it to work.
+
+Here are some tips to get started with horizontal scaling:
+- Understand parallel computing and how to use it to speed up computation.
```
Zhiyuan-Amos: This is the only point without a hyperlink to an external resource :P

contents/scalability/scalable-development.md
```diff
@@ -15,12 +15,33 @@ So what is scalability? Scalability is the capacity to which a system can cope w
...
+
+#### Vertical Scaling
 
 This is arguably the easiest way to speed up your server, as it is a general solution to most problems. Simply get faster CPUs, larger RAM, etc. This is called vertical scaling. If you get multi-core processors, use threading in your server to allow parallelization of certain processes to take advantage of the cores.
 
 However, due to Amdahl's law, each additional processor will give a less-than-proportionate speed up. Coupled with the exponential increase in price as the computing power of the CPU increases, this is typically not scalable in the long run.
 
+#### Should you choose one or the other?
+
+With vertical scaling, you get to take advantage of processing power and concurrency. But as mentioned above, this isn't sustainable over the long run. Furthermore, you can't run your server on just one machine as it results in a single point of failure and prevents changes to be made to the code without bringing the entire system down. Moreover, you can't distribute your application geographically.
```
Zhiyuan-Amos: This can be done in list form. 

This isn't sustainable over the long run because:
1. Single point of failure
1. Prevents changes made to code...

contents/scalability/scalable-development.md
```diff
@@ -29,26 +50,42 @@ In certain scenarios, the runtime complexity can be reduced as a trade-off betwe
...
+The next thing to consider is what to cache. The rule of thumb is to cache data that is frequently accessed and read more often than it is updated. With this, cache hits would be more often than cache misses and as a result, the time saved from faster accesses and reads outweighs the extra time taken to populate the cache.  
 
-Non-Relational Database ([resource](http://www.jamesserra.com/archive/2015/08/relational-databases-vs-non-relational-databases/)) ([resource](https://www.pluralsight.com/blog/software-development/relational-non-relational-databases))
+#### Sharding 
+As your traffic increases your data increases as well and as a result, your database gets overloaded. One way to mitigate this is to scale your database by sharding. Sharding is a method of splitting and storing a single logical dataset in multiple databases. More specifically, it is the storing of data horizontally - storing rows of a same table in multiple database nodes instead of storing them in the usual vertical way - storing different tables & columns in a separate database.
 
-### Get More Machines
+What are the benefits of Sharding?
+* Sharding can store more data than a single database can hold -
```
Zhiyuan-Amos: Sharding allows you to store more data...

contents/scalability/scalable-development.md
```diff
@@ -29,26 +50,42 @@ In certain scenarios, the runtime complexity can be reduced as a trade-off betwe
...
+Sharding is essential when your dataset becomes too large to store in a single database. It reduces the number of rows in each table and as such improves search performance since the search is done on a smaller table. 
 
-Buy more machines to host your server. This is called horizontal scaling. This method can help improve scalability by a huge factor, as each additional unit of processing power and RAM costs the same, no matter how many machines you buy. As this is a promising way to increase scalability, it is no surprise why large services such as Google and Facebook use thousands of computer servers just to provide services to all their users.
+* Sharding can make queries faster - Querying the databases containing only the relevant partitions becomes possible as well. For instance, if a database contains a column for age, you can partition the rows according to an age group and store them in different databases. Whenever there is a need to access the data of a particular age group, instead of querying the whole database, you just need to query the partition that contains that age group. This allows your database to scale along with your data and traffic growth.
```
Zhiyuan-Amos: Sharding allows for quicker query response time...

contents/scalability/scalable-development.md
```diff
@@ -29,26 +50,42 @@ In certain scenarios, the runtime complexity can be reduced as a trade-off betwe
...
 
-However, to fully reap the benefits of multiple machines, you will need to learn various techniques to modify your server infrastructure. In most cases you cannot simply run multiple instances of the server and expect it to work.
+#### Go Asynchronous 
+Unlike synchronous operations, which runs sequentially and waits for the previous operation to complete before moving on to the next, asynchronous operations are non-blocking i.e. does not block further execution until that operation finishes. As a result other operations do not have to wait for asynchronous operations to finish.
```
Zhiyuan-Amos: > i.e. does not block further execution until that operation finishes.

Quite hard to understand. I think it's unnecessary for this to be here, since you already have a "As a result..." to explain what asynchronous operations mean.

contents/scalability/scalable-development.md
```diff
@@ -29,26 +50,42 @@ In certain scenarios, the runtime complexity can be reduced as a trade-off betwe
...
+
+1. Scalability best practices ([resource](https://www.infoq.com/articles/ebay-scalability-best-practices))
+
+1. How Sharding Works ([resource](https://medium.com/@jeeyoungk/how-sharding-works-b4dec46b3f6))
+
+1. A StackOverflow post on Sharding ([resource](https://stackoverflow.com/questions/992988/what-is-sharding-and-why-is-it-important))
+
+1. An overview of how asynchronous programming works with concepts such as event loops and callbacks, explained using Javascript  ([resource](https://www.youtube.com/watch?time_continue=2&v=8aGhZQkoFbQ))
+
+To find these bottlenecks, you can use the many program profilers available to find the bottlenecks in your server. By finding out where the server spends most computational time in, you can potentially save huge amounts of time.
```
Zhiyuan-Amos: > ~~To find these bottlenecks,~~ you can use the many program profilers available to find the bottlenecks in your server ~~By finding out where the server spends most computational time in,~~ thereby potentially saving huge amounts of time.

contents/scalability/scalable-development.md
```diff
@@ -29,26 +50,42 @@ In certain scenarios, the runtime complexity can be reduced as a trade-off betwe
...
+##### Resources
+1. More details on the common performance bottlenecks ([resource](https://www.apicasystem.com/blog/5-common-performance-bottlenecks))
+
+1. Scalability using caches ([resource](http://www.lecloud.net/post/9246290032/scalability-for-dummies-part-3-cache))
+
+1. Scalability best practices ([resource](https://www.infoq.com/articles/ebay-scalability-best-practices))
+
+1. How Sharding Works ([resource](https://medium.com/@jeeyoungk/how-sharding-works-b4dec46b3f6))
+
+1. A StackOverflow post on Sharding ([resource](https://stackoverflow.com/questions/992988/what-is-sharding-and-why-is-it-important))
```
Zhiyuan-Amos: > A StackOverflow post on Sharding

This is really undescriptive :P

Zhiyuan-Amos (commented):
I don't think I have time for another review, so I won't give an explicit approval / changes requested.

contents/scalability/scalable-development.md
vivekscl: `i.e. does not block further execution until that operation finishes`
Oh, this was to explain the term non-blocking as mentioned by Prof. So it's better to replace `non-blocking` with  `does not block further execution until that operation finishes` right?

contents/scalability/scalable-development.md
```diff
@@ -29,26 +53,45 @@ In certain scenarios, the runtime complexity can be reduced as a trade-off betwe
...
+With this, cache hits would be more often than cache misses and as a result, the time saved from faster accesses and reads outweighs the extra time taken to populate the cache.  
 
-### Get More Machines
+#### Sharding 
+As your traffic increases your data increases as well and as a result, your database gets overloaded. One way to mitigate this is to scale your database by sharding. Sharding is a method of splitting and storing a single logical dataset in multiple databases. More specifically, it is the storing of data horizontally - storing rows of a same table in multiple database nodes instead of storing them in the usual vertical way - storing different tables & columns in a separate database.
 
-Buy more machines to host your server. This is called horizontal scaling. This method can help improve scalability by a huge factor, as each additional unit of processing power and RAM costs the same, no matter how many machines you buy. As this is a promising way to increase scalability, it is no surprise why large services such as Google and Facebook use thousands of computer servers just to provide services to all their users.
+What are the benefits of Sharding?
```
darrenwee: Use a subheading and rephrase to `##### Benefits of Sharding`

darrenwee (approved):
LGTM. Just 1 formatting nitpick, but it's minor.

vivekscl:
@damithc Ready for review.

[View the rendered version here](https://github.com/vivekscl/learningresources/blob/add-ways-to-scale/contents/scalability/scalable-development.md).

contents/scalability/scalable-development.md
```diff
@@ -29,26 +53,45 @@ In certain scenarios, the runtime complexity can be reduced as a trade-off betwe
...
+With this, cache hits would be more often than cache misses and as a result, the time saved from faster accesses and reads outweighs the extra time taken to populate the cache.  
 
-### Get More Machines
+#### Sharding 
+As your traffic increases your data increases as well and as a result, your database gets overloaded. One way to mitigate this is to scale your database by sharding. Sharding is a method of splitting and storing a single logical dataset in multiple databases. More specifically, it is the storing of data horizontally - storing rows of a same table in multiple database nodes instead of storing them in the usual vertical way - storing different tables & columns in a separate database.
 
-Buy more machines to host your server. This is called horizontal scaling. This method can help improve scalability by a huge factor, as each additional unit of processing power and RAM costs the same, no matter how many machines you buy. As this is a promising way to increase scalability, it is no surprise why large services such as Google and Facebook use thousands of computer servers just to provide services to all their users.
+##### Benefits of Sharding
```
damithc: Shall we remove this and other level 5 headings?

contents/scalability/scalable-development.md
```diff
@@ -29,26 +53,45 @@ In certain scenarios, the runtime complexity can be reduced as a trade-off betwe
...
+
+#### Go Asynchronous 
+Unlike synchronous operations that run sequentially and wait for previous operations to complete, asynchronous operations do not block further execution and as a result, other operations do not have to wait.
+
+In situations in which responding to request is crucial, asynchronous operations can reduce the latency experienced by the requester and thereby avoid that bottleneck by prioritising the quickness of the response to the user over the speed of other processes such as execution latency (how quickly the request is processed). For instance, rather than waiting for some processes such as downloading the requested file to finish, it's better to asynchronously update the user interface and then finish these processes. In this way, the user experiences lesser latency since the user interface is updated instantly instead of appearing to have crashed due to waiting for the downloaded file.
+
+##### Resources
```
damithc: This should be a level 4 heading instead of level 5?

damithc: In addition, you can also split this up and move each resource to the section that describe the topic.

damithc (changes requested):
Looks good. A few minor suggestions.

vivekscl:
@damithc Ready for Review.

damithc (approved):
LGTM 👍

PR #35, opened 2018-02-27
Add section for Promises (JavaScript)

danielbrzn:
This PR aims to add a section regarding Promises in Javascript.


contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,104 @@
+# Javascript: Promises
+
+Author: Daniel Berzin Chua
+
+In Javascript, there are occasions where a method call may not resolve instantly and as such, the code may not run in the sequence that you would expect due this asynchronous nature. It can be especially confusing for beginners learning Javascript.
```
mauris: I wouldn't call it a "method call". perhaps you can say that an operation is executed asynchronously and its result may not be obtained immediately as a function return value would.

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,104 @@
...
+
+A pertinent example would be HTTP requests. When making a HTTP request, the information from the server is not instantly returned as there is a round trip time that involves the time taken for the request to be sent, and for the response from the server to be sent back to the client. In this time, Javascript would not necessarily wait for the response to be received, but instead execute other functions first, even if those functions depend on the result of that HTTP response.
+
+In the code snippet below, `console.log(price)` would return undefined as the asynchronous `retrievePrice()` function has not finished executing.
```
mauris: a diagram that shows the execution timeline would be nice to have.

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,104 @@
+# Javascript: Promises
+
+Author: Daniel Berzin Chua
+
+In Javascript, there are occasions where a method call may not resolve instantly and as such, the code may not run in the sequence that you would expect due this asynchronous nature. It can be especially confusing for beginners learning Javascript.
+
+A pertinent example would be HTTP requests. When making a HTTP request, the information from the server is not instantly returned as there is a round trip time that involves the time taken for the request to be sent, and for the response from the server to be sent back to the client. In this time, Javascript would not necessarily wait for the response to be received, but instead execute other functions first, even if those functions depend on the result of that HTTP response.
```
mauris: good example. a simpler example would actually be timer, i.e. `setTimeout()`. You can explain what is callback straight up here in a setTimeout example and then talk about async execution and event loop

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,104 @@
...
+In the code snippet below, `console.log(price)` would return undefined as the asynchronous `retrievePrice()` function has not finished executing.
+
+```
+var price;
+price = retrievePrice("https://www.amazon.com/gp/product/B009S331VU/");
+console.log(price);
+
+```
+
+As such, it can be difficult to predict how your Javascript code executes as asynchronous methods such as the above would not necessarily execute in the sequence that is desired.
```
mauris: "it can be difficult to predict how" - I think maybe it's better to not describe as "predict" (as though it's a random variable). Perhaps we can say that retrievePrice is non-blocking (this is keyword) and the I/O operation would be done in a different thread (depends on platform implementation) etc.

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,104 @@
...
+
+This is where Promises would come in handy.
+
+
+## Introduction
+Promises in Javscript behave the same way as Promises do in real life. It is a proxy for a value that is not known at the time of Promise creation. It has 3 states: pending, fulfilled and rejected. Promises provide the power to control the execution flow of your program by allowing the user to wait for a promise to be resolved before carrying on with the rest of the program execution.
+
+## Benefits
+
+- Avoid callback hell
```
mauris: as mentioned before, an explanation of callback-based programming first before introducing Promises can help with this.

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,104 @@
...
+
+![Callback Hell](http://blog.mclain.ca/assets/images/callbackhell.png)
```
mauris: avoid linking to external resources that may become unavailable sometime later. if need be and allowed, put the image in the repository too!

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,104 @@
...
+- Chain multiple asynchronous methods
+
+Calling asynchronous code can be tedious, especially if a later function relies on data that is produced by an earlier function. By using Promises, these asynchronous methods can be chained and variables will be correctly set instead of being undefined when the variable needs to be used.
+
+- Control execution flow
+
+As mentioned earlier, calling asynchronous methods in Javascript may not necessarily result in a sequential execution flow. However, it is possible to ensure that methods get called in the order that they are written in, as long as they are wrapped in a Promise.
+
+
+## How
```
mauris: titles can be more elaborated and to the point, perhaps here we can say "How to use Promises"

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,104 @@
...
+    console.log(result);    // prints 'Success'
+}).catch(function(result) {
+    console.log(result);    // prints 'Failure'
+});
+
+```
+
+## Other functions
+
+Sometimes multiple promises may have to be used at a time, and Javascript provides excellent support with the `Promise.all` and `Promise.race` functions.
```
mauris, line 187: there are other libraries that expand on the native Promise library provided by default. I personally use Bluebird where they have the .spread() functionality. check them out and see if you're able to add in to the doc

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,104 @@
...
+Calling asynchronous code can be tedious, especially if a later function relies on data that is produced by an earlier function. By using Promises, these asynchronous methods can be chained and variables will be correctly set instead of being undefined when the variable needs to be used.
+
+- Control execution flow
+
+As mentioned earlier, calling asynchronous methods in Javascript may not necessarily result in a sequential execution flow. However, it is possible to ensure that methods get called in the order that they are written in, as long as they are wrapped in a Promise.
+
+
+## How
+
+Promises are created in Javascript by using the `new` constructor as follows:
```
mauris: It'd be good to mention that Promises are available from which versions of JavaScript and on which platforms.

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,104 @@
...
+Whenever you want to act on a result from a Promise, it is good practice to use the above 2 functions to ensure certainty and code readability.
+
+An example would be as follows:
+
+```
+
+var isSuccessful = false;
+
+var myFirstPromise = new Promise(function(resolve, reject) {
+    if (isSuccessful)
```
mauris: use `{` and `}` even if it's one line inside the if-branch.

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,104 @@
+# Javascript: Promises
+
+Author: Daniel Berzin Chua
+
```
junkiattan, line 4: You can add a Context title to let readers know the context of Promises in Javascript

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,104 @@
...
+As such, it can be difficult to predict how your Javascript code executes as asynchronous methods such as the above would not necessarily execute in the sequence that is desired. 
+
+This is where Promises would come in handy.
+
+
+## Introduction
+Promises in Javscript behave the same way as Promises do in real life. It is a proxy for a value that is not known at the time of Promise creation. It has 3 states: pending, fulfilled and rejected. Promises provide the power to control the execution flow of your program by allowing the user to wait for a promise to be resolved before carrying on with the rest of the program execution.
+
+## Benefits
```
junkiattan: I feel the flow would be better if you had put the Benefits section before the Other Functions section. As a reader, I would want to know the implementation details first to judge the usability before understanding the benefits to further solidifying the point that Promises are good

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,104 @@
...
+resolve(someValue);
+reject(someValue);
+resolve();
+reject();
+
+```
+
+## Handling
+
+As promises have 2 states, there are 2 functions available to deal with these 2 scenarios.
```
junkiattan: Earlier in your text, you said that Promises have 3 states. Do be careful not to confuse readers. Perhaps you were referring to "2 completed states" instead.

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,104 @@
...
+Author: Daniel Berzin Chua
+
+In Javascript, there are occasions where a method call may not resolve instantly and as such, the code may not run in the sequence that you would expect due this asynchronous nature. It can be especially confusing for beginners learning Javascript.
+
+A pertinent example would be HTTP requests. When making a HTTP request, the information from the server is not instantly returned as there is a round trip time that involves the time taken for the request to be sent, and for the response from the server to be sent back to the client. In this time, Javascript would not necessarily wait for the response to be received, but instead execute other functions first, even if those functions depend on the result of that HTTP response.
+
+In the code snippet below, `console.log(price)` would return undefined as the asynchronous `retrievePrice()` function has not finished executing.
+
+```
+var price;
```
junkiattan: In general for all your code examples, is it possible to color code them like in actual code? I have seen some of the resources with code examples that do this. 

This makes the code stand out more, and helps in the flow.

contents/javascript/Javascript-promises.md
junkiattan: The example is good and easy to understand for me

contents/javascript/Javascript-promises.md
danielbrzn: Managed to do it with Markdown code syntax highlighting. Thanks for the feedback!

contents/javascript/Javascript-promises.md
danielbrzn: Good catch.

contents/javascript/Javascript-promises.md
danielbrzn: Do you have an example for this? I'm not quite sure what you mean.

contents/javascript/Javascript-promises.md
danielbrzn: Good catch, thanks!

contents/javascript/Javascript-promises.md
junkiattan: What I meant was to have a Context title, like your Introduction title or How to use Promises title.

That way people would instantly be mentally prepped of what to anticipate for your first section. This is a small change and completely optional.

danielbrzn:
Hi Prof @damithc,

May I get a review on this PR?

Thanks!

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,129 @@
...
+console.log(price);
+
+```
+
+Actual Execution Flow:
+![Image](asyncRequestSequence.png)
+
+Since `retrivePrice()` is a non-blocking function and it's only handled after the `console.log()` has executed, there must be a way to ensure that these functions are executed in the sequence that's desired.
+
+This is where Promises come in handy.
```
damithc: This text up to this point probably need another go. Start with a simple example and gently explain what synchronous/asynchronous means and build up until the reader understands the problem context. As @junkiattan mentioned, you can give a heading to the section too. e.g., **Why Promises?**

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,129 @@
...
+
+
+## Introduction
```
damithc: If you name the previous section `Why Promises?`, you can name this `What is a Promise?`

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,129 @@
...
+Promises in Javscript behave the same way as Promises do in real life. It is a proxy for a value that is not known at the time of Promise creation. It has 3 states: pending, fulfilled and rejected. Promises provide the power to control the execution flow of your program by allowing the user to wait for a promise to be resolved before carrying on with the rest of the program execution.
```
damithc: The first sentence hints that promises can be understood by relating to a similar concept in real life; yet the second sentence says something that is hardly related to real life. On a similar note, what is meant by '_resolve_ a promise'. That's not something a casual reader can be expected to know.

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,129 @@
...
+
+## How to use Promises
+
+The Promise API is only supported natively from ES6 onwards, which is the version that modern browsers support. The code below will thus work in any modern browser and also in Node.js (from version 6).
```
damithc: Given that this resource is not meant to _teach_ someone how to use promises, I wonder if we should omit this type of nitty-gritty details.

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,129 @@
...
+
+Since `retrivePrice()` is a non-blocking function and it's only handled after the `console.log()` has executed, there must be a way to ensure that these functions are executed in the sequence that's desired.
+
+This is where Promises come in handy.
+
+
+## Introduction
+Promises in Javscript behave the same way as Promises do in real life. It is a proxy for a value that is not known at the time of Promise creation. It has 3 states: pending, fulfilled and rejected. Promises provide the power to control the execution flow of your program by allowing the user to wait for a promise to be resolved before carrying on with the rest of the program execution.
+
+## How to use Promises
```
damithc: May be reframe this as 'How Promises are used' and give some nice examples of how Promises are used? The current title suggests that the section teaches how to use Promises. Normally our learning resources don't try to teach topics. Rather, we want to motivate the reader to learn the topic and point to other resources for the actual learning of the topic.

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,129 @@
...
+```
+
+## Benefits
+
+- Avoid callback hell
+
+![Callback Hell](callbackhell.png)
+(Credit: [Brett McLain](http://blog.mclain.ca/))
+
+In Javascript, callback functions are usually used to perform some action on a value returned from another function. However, this may result in deeply nested code as shown above, especially if there are many asynchronous method calls. When using promises, the nesting is kept at a minimum and results in cleaner and readable code as shown below.
```
damithc: This is where you explain what a callback is, but I remember seeing the term callback near the beginning of this chapter. :-p

damithc (changes requested):
Some changes suggested. Read the current from a 'knowledgeable outsider' POV and you'll realize it's not easy to follow. Probably you don't see it because you already know the topic. Read a few other similar articles on Promises like [this one](https://developers.google.com/web/fundamentals/primers/promises) and try to come up with a shorter version that combines the best of those articles (and point to the original articles for a longer version).

danielbrzn:
@damithc may I get another review? I made some changes for the content to be more user-friendly, and also removed the section about how to use them, replacing it with an "how they are used" section instead.

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,159 @@
+# Javascript: Promises
+
+Author: Daniel Berzin Chua
+
+## Why Promises?
+
+In Javascript, there are times where an operation may not return a value immediately.
```
damithc: This is not specific to JavaScript.

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,159 @@
...
+It can be especially confusing to trace the code since it may not run in the sequence that you would normally expect.
+
+Think about making a HTTP request to a website: There's a round trip time that involves the time for request to be sent to the server you are querying, and the time for the server to send the response back to you. It may take varying amounts of time depending on your internet speed and location.
+This is known as an asynchronous operation, as the response isn't returned immediately. Compare this to a synchronous operation, where a value is returned near instantly after the operation is called.
+
+Asynchronous operations such as the above can cause programming and debugging to be difficult, because you would need some sort of way to know when the operation has finished, or in the case of debugging, when exactly the operation is called.
+
```
damithc: The description up to this point can be tweaked slightly along this flow:
* Usually, ops are synchronous i.e., wait for completion before proceeding
* But some ops take a while to finish. e.g., http request 
* Doing such ops in synchronous manner is not good for user experience
* In such cases, we can use async ops. they don't wait for completion.
* but code of async ops are harder to understand/debug than sync ops

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,159 @@
...
+setTimeout(function (){
+    x = "I have been updated";
+}, 1000);
+
+console.log(x);
+
+```
+
+To fix this, we can simply shift the `console.log(x)` into the callback to get the expected result. It will now print the correct value.
```
damithc: have you explained what a callback is?

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,159 @@
...
+
+setTimeout(function (){
+    x = "I have been updated";
+    console.log(x);
+}, 1000);
+
+```
+
+However, this fix will only go so far. If we had another `setTimeout()` that depended on the result of the earlier `setTimeout()`, we would have to nest the functions within each other which would make for unreadable code if the nesting is too deep. An extreme example of this is known as callback hell, which is elaborated on in the Benefits section of this chapter.
+
```
damithc: Best explain it now rather than .. ahem ... "promise" to explain it later. 
In any case, this example is supposed to show why async code is hard to read/debug (following from the previous paragraph) but doesn't seem to do so.

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,159 @@
...
+These situations correspond to the 3 states of Promises in Javascript.
+
+| State | Description |
+| ------ | ----------- |
+|Pending | You don't know if he would return your money. |
+|Fulfilled | He returned your money. |
+|Rejected | He refused to return your money. |
+
+
+Promises provide the power to control the execution flow of your program by allowing  to wait for a promise to be resolved before carrying on with the rest of the program execution.
```
damithc: What is meant by `resolved`?

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,159 @@
...
+            }
+        );
+    });
+}
+
+getPriceFromUrl()
+    .then(price => console.log(price))
+    .catch(errorMsg => console.log(errorMsg));
+
+```
```
damithc: Well, this code looks harder to read than the setTimeout() code, which contradicts your claim about Promises :-p

damithc: explain what's so good about this code.

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,159 @@
...
+        });
+    })
+
+};
+
+readFileWithPromise('path/to/file')
+    .then(data => console.log(data))
+    .catch(err => console.log(err));
+
+```
```
damithc, line 183: This code looks simpler than the httprequest example. May be use this as the running example for the entire chapter? First show how it is without Promises and then show how Promises makes it better. It's easier to follow when you use the same example throughout. e.g.,

Add two numbers locally : this can be a snyc op
Add two numbers remotely: this takes time. Need to be async.
Let's use callbacks. Here is the code.
What if we need to chain a few remote add ops? Here is the code. Oops! callback hell.
Promises to the rescue. 
Here's how they work:
First we need to create a promise. Here is the code. 
Next we need to consume the promise. Here is the code. 
Here is how we can chain. See how much better it is from callbacks.
There are some more advance stuff you can do with promises: x, y, z
There are some alternatives to promises: a, b, c.

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,159 @@
...
+As mentioned earlier, calling asynchronous methods in Javascript may not necessarily result in a sequential execution flow. However, it is possible to ensure that methods get called in the order that they are written in, as long as they are wrapped in a Promise.
+
+## Other functions
+
+Sometimes multiple promises may have to be used at a time, and Javascript provides excellent support with the `Promise.all` and `Promise.race` functions.
+
+There is an excellent write up on these methods [here](https://davidwalsh.name/promises), which will allow you to use Promises even more efficiently.
+
+In addition, there are other libraries such as [Bluebird](http://bluebirdjs.com/docs/getting-started.html) and [Q](https://github.com/kriskowal/q) which offer even more functionality such as Promise monitoring and synchronous inspection of Promises.
+
```
damithc: Both images in this PR used?

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,159 @@
+# Javascript: Promises
```
damithc, line 1: Better than before 👍 Some way to go before you can make it understandable to a 'knowledgeable outsider' (not only to existing JS programmers). I suspect you can some inspiration from this article https://scotch.io/tutorials/javascript-promises-for-dummies (remember to acknowledge it).

damithc (changes requested):
Good progress. Need a few more iterations though, as this is not an easy topic to make accessible to a 'knowledgeable outsider'.

contents/javascript/Javascript-promises.md
danielbrzn: Yes it's been explained in the paragraph before the code snippet was shown.

contents/javascript/Javascript-promises.md
danielbrzn: I think it touches on why async code is hard to debug as I explain that the program execution isn't what the user expected, e.g. wait for 1s then log with the updated message.

contents/javascript/Javascript-promises.md
danielbrzn: Added a snippet to acknowledge their example.

contents/javascript/Javascript-promises.md
danielbrzn: Removed.

contents/javascript/Javascript-promises.md
danielbrzn: I feel like using the SetTimeout example would be easier for new users of JS as it's something they can try out on their own in Chrome/IDE to see how it actually works, and how we can resolve the problems of async operations using Promises. 

This file reading example is just a way of showing how you can still read files fast using async ops that allow it to be read in the background.

danielbrzn:
@damithc, may I get another review?

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,164 @@
...
+These situations correspond to the 3 states of Promises in Javascript.
+
+| State | Description |
+| ------ | ----------- |
+|Pending | You don't know if he would return your money. |
+|Fulfilled | He returned your money. |
+|Rejected | He refused to return your money. |
+
+
+Promises provide the power to control the execution flow of your program by allowing you to wait for a asynchronous operation to complete before carrying on with the rest of the program execution.
```
damithc: This statement sounds like "promises provide a way to do things synchronously" (because it says that the rest of the program will not execute until the async code is executed). I don't know promises myself, but the statement didn't sound right to me. Did you mean "promises allows you to specify how the execution of _some part_ of the code to depend on the status of an asynchronous code" or something of the sort.

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,164 @@
...
+
+## How Promises are used
+
+There are many interesting ways to use Promises. The following examples are written with Node.js.
```
damithc: hmm... it's unfortunate that you need to bring node into this, given that we consider node to be a separate topic altogether. Is there any way to avoid it?

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,164 @@
...
+
+### HTTP Requests
+
+Earlier in this chapter, I mentioned how HTTP requests are asynchronous operations. By using Promises, you would be able to act on the result from the request without having to use callbacks or wait an arbitrarily set amount of time for the response to be returned.
```
damithc: Better not use first person (`I`) as this chapter might have more than one author in future.

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,164 @@
...
+
+The following code sends a request to an Amazon product URL and retrieves the price of the product using Promises. By using Promises instead of callbacks, we have a nice little web scraping snippet that's easy for anyone to use and to debug if any issues arise.
```
damithc: Web scraping is another topic. Do we need to mention it here?

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,164 @@
...
+
+| State | Description |
+| ------ | ----------- |
+|Pending | You don't know if he would return your money. |
+|Fulfilled | He returned your money. |
+|Rejected | He refused to return your money. |
+
+
+Promises provide the power to control the execution flow of your program by allowing you to wait for a asynchronous operation to complete before carrying on with the rest of the program execution.
+
```
damithc: While this chapter is not meant to to teach promises syntax, I think you still need to walk the reader through a small hello world type example for the reader to get some sense of what promises are. May be explain the promises basics using how to write the timer example using promises?

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+
+## Benefits
+
+- Avoid callback hell
+
+![Callback Hell](callbackhell.png)
+(Credit: [Brett McLain](http://blog.mclain.ca/))
```
damithc: Image link is broken. In any case, you don't need it as you already have a similar example at the beginning of the chapter.

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,164 @@
...
+setTimeout(function (){
+    x = "I have been updated";
+    setTimeout(function () {
+        x = "I have been updated 2 times.";
+        setTimeout(function () {
+            x = "I have been updated 3 times.";
+            // and so on...
+        }, 1000);
+    }, 1000);
+}, 1000);
```
damithc: You can mention callback hell here?

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,164 @@
...
+
+To illustrate this problem in code, we'll use `readFile()` from the `fs` package in Node.js as an example. This is an asynchronous function takes in 3 parameters, a callback and a time (in milliseconds) to wait before the callback is executed. A callback is a function that is passed as a parameter to another function (call this A), and it will be executed after A finishes.
+
+In the following code snippet, you would probably expect `console.log(x)` to print `I have been updated`. Unfortunately, it prints `I have not been updated.` Give it a try in Google Chrome's developer console.
```
damithc: setTimeout is an async function? It wasn't obvious to me at first. It is something a relatively new JS programmer supposed to know? If yes, can leave as it is. If no, explain it a bit.

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,164 @@
...
+
+To illustrate this problem in code, we'll use `readFile()` from the `fs` package in Node.js as an example. This is an asynchronous function takes in 3 parameters, a callback and a time (in milliseconds) to wait before the callback is executed. A callback is a function that is passed as a parameter to another function (call this A), and it will be executed after A finishes.
```
damithc: Where is this `readFile()` used in the text below?

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,164 @@
...
+};
+
+readFileWithPromise('path/to/file')
+    .then(data => console.log(data))
+    .catch(err => console.log(err));
+
+```
+
+
+## Benefits
```
damithc: This section and the next section seem thin. One option is to get rid of them and absorb the content into the previous section. i.e., explain them as different 'ways of using promises' and mention the benefits at the same time. For example, the benefit of avoiding callback hell can be mentioned in the timeout example.

damithc (changes requested):
Getting better; a few more kinks to work out.

contents/javascript/Javascript-promises.md
acjh: @damithc It's used in the "Disk I/O" section.

---
@danielbrzn 

> This is an asynchronous function [that] takes in 3 parameters, a callback and a time (in milliseconds) to wait before the callback is executed.

That sounds like 5 parameters in total. The 3 parameters are [path[, options], callback](https://nodejs.org/api/fs.html#fs_fs_readfile_path_options_callback).

damithc: Just like we should define variables as closer to where they are used as possible, introduce terms closer to where they are used :-)

contents/javascript/Javascript-promises.md
danielbrzn: Would 1 Node.js example (file I/O) and 1 Javascript example be a good compromise? Or should we try to make both examples pure JS.

damithc: I'll leave it to you to decide. Try to minimize dependency on node.js so that even someone who has no knowledge of node.js can understand this content, unless there is a legit reason for promises to depend on node.js

danielbrzn:
@damithc , could I get another review :)

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,198 @@
...
+
+function myAsyncFunction(url) {
+  return new Promise((resolve, reject) => {
+    const xhr = new XMLHttpRequest();
+    xhr.open("GET", url);
+    xhr.onload = () => resolve(xhr.responseText);
+    xhr.onerror = () => reject(xhr.statusText);
+    xhr.send();
+  });
+}
```
damithc, line 150: May be show how this is used too, to be consistent with other examples you have given?

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,198 @@
...
+    .then(data => console.log(data))
+    .catch(err => console.log(err));
+
+```
+
+## Other functions
+
+Sometimes multiple promises may have to be used at a time, and Javascript provides excellent support with the `Promise.all` and `Promise.race` functions.
+
+Here we will go through what `Promise.all()` can help with.
```
damithc: Can remove this statement without any loss.

contents/javascript/Javascript-promises.md
```diff
@@ -0,0 +1,198 @@
+# Javascript: Promises
+
+Author: Daniel Berzin Chua
+
+## Why Promises?
+
+We're used to operations completing their work before proceeding on with other operations in a sequential manner. These sorts of operations are synchronous, and they're usually easy to understand and debug. However, take a HTTP request for example. It is an operation that takes a while to process because there's a round trip time that involves the time for request to be sent to the server you are querying, and the time for the server to send the response back to you. It may take varying amounts of time depending on your internet speed and location.
```
damithc: `an HTTP`? https://www.quora.com/A-http-or-an-http-which-is-the-correct-grammatical-usage

damithc (commented):
Looks good. Just a few nits.

contents/javascript/Javascript-promises.md
danielbrzn: Added a snippet to show how the function is called and handled.

danielbrzn:
I've made the necessary changes.

danielbrzn:
@damithc Are there any more changes that I should make for this chapter?

damithc (approved):

damithc:
Merged. It's in good shape now. Good job doing all those revisions along the way. 👍

danielbrzn:
Thank you for the guidance along the way 👍

PR #36, opened 2018-02-27
Kotlin: Basic information about Kotlin

tshradheya:
Started a new details page for Kotlin

### DIFF
--- a/contents/kotlin/kotlin.md
+++ b/contents/kotlin/kotlin.md
@@ -17,7 +17,7 @@ Author: [Shradheya Thakre](https://github.com/tshradheya)
 
 # Overview
 
-Kotlin is a general purpose, open source, statically typed “pragmatic” programming language for the JVM and Android that combines object-oriented and functional programming features.
+Kotlin is a general purpose, open source, statically typed programming language for the JVM and Android that combines object-oriented and functional programming features.
 
 # Why Kotlin?
 
@@ -72,7 +72,7 @@ fun onLoad() {
 
 ### Tool Friendly
 
-Hyperbolically, a programming language is only as good as what its tools can provide. This is why the advantage of using Kotlin is the built-in language support from IntelliJ. Any Java IDE for e.g. IntelliJ and Eclipse, can be used to write and compile Kotlin code. It also contains the aforementioned Java-to-Kotlin converter and code generators for Java and JavaScript from Kotlin code.
+Hyperbolically, a programming language is only as good as what its tools can provide. This is why the advantage of using Kotlin is the built-in language support from IntelliJ. Any Java IDE for e.g. IntelliJ, Eclipse and Android Studio, can be used to write and compile Kotlin code. It also contains the aforementioned Java-to-Kotlin converter and code generators for Java and JavaScript from Kotlin code.
 
 # Kotlin for Android Apps
 
--- a/contents/kotlin/kotlin.md
+++ b/contents/kotlin/kotlin.md
@@ -11,9 +11,8 @@ Author: [Shradheya Thakre](https://github.com/tshradheya)
     * [Interoperable](#interoperable)
     * [Tool Friendly](#tool-friendly)
 * [Kotlin for Android Apps](#kotlin-for-android-apps)
-    * [Advantages of Shifting to Kotlin](#advantage-kotlin)
-    * [Drawbacks of Shifting to Kotlin](#drawbacks-kotlin)
-
+    * [Advantages of Shifting to Kotlin](#advantages-of-shifting-to-kotlin)
+    * [Drawbacks of Shifting to Kotlin](#drawbacks-of-shifting-to-kotlin)
 
 # Overview
 
--- a/contents/kotlin/kotlin.md
+++ b/contents/kotlin/kotlin.md
@@ -2,7 +2,7 @@
 
 Author: [Shradheya Thakre](https://github.com/tshradheya)
 
-**Table of Contents**
+**Chapter contents**
 
 * [Overview](#overview)
 * [Why Kotlin?](#why-kotlin)
--- a/contents/kotlin/kotlin.md
+++ b/contents/kotlin/kotlin.md
@@ -15,10 +15,6 @@ Author: [Shradheya Thakre](https://github.com/tshradheya)
     * [Drawbacks of Shifting to Kotlin](#drawbacks-kotlin)
 
 
-
-
-
-
 # Overview
 
 Kotlin is a general purpose, open source, statically typed “pragmatic” programming language for the JVM and Android that combines object-oriented and functional programming features.
@@ -28,11 +24,13 @@ Kotlin is a general purpose, open source, statically typed “pragmatic” progr
 ### Concise
 
 Highly reduces the amount of boilerplate code.
-- A POJO with getters, setters, equals(), hashCode(), toString() and copy() can be created in a single line
+- A Java object with getters, setters, equals(), hashCode(), toString() and copy() can be created in a single line in Kotlin
 ``` kotlin
 data class Customer(val name: String, val email: String, val company: String)
 ```
-- Assigning values based on ranges is much simpler
+- Assigning values based on ranges is much simpler than Java. Below are extracts of code with similar logic in Java and Kotlin:
+
+**Kotlin**
 ``` kotlin
 val quartile: Int
 quartile = when (playPercentage) {
@@ -42,11 +40,21 @@ quartile = when (playPercentage) {
 }
 ```
 
+**Java**
+``` java
+int quartile;
+if(playPercentage >= 0 && playPercentage <= 24) {
+    quartile = 0;
+} else if(playPercentage >= 25 && playPercentage <= 49) {
+    quartile = 1;
+}
+...
+```
 
 ### Safe
 Kotlin protects you from mistakenly operating on nullable types
 - Get compilation error when you mistakenly try operating on nullable types
-```
+``` kotlin
 val name: String? = null    // Nullable type
 println(name.length())      // Compilation error
 ```
@@ -54,7 +62,7 @@ println(name.length())      // Compilation error
 ### Interoperable
 
 - Target either the JVM or JavaScript. Write code in Kotlin and decide where you want to deploy
-```
+``` kotlin
 import kotlin.browser.window
 fun onLoad() {
     window.document.body!!.innerHTML += "<br/>Hello, Kotlin!"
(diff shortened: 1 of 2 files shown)


tshradheya:
@haozhe321 Requesting you for a review since you gave a talk on Kotlin :)

contents/kotlin/kotlin.md
```diff
@@ -0,0 +1,87 @@
...
+* [Overview](#overview)
+* [Why Kotlin?](#why-kotlin)
+    * [Concise](#concise)
+    * [Safe](#safe)
+    * [Interoperable](#interoperable)
+    * [Tool Friendly](#tool-friendly)
+* [Kotlin for Android Apps](#kotlin-for-android-apps)
+    * [Advantages of Shifting to Kotlin](#advantage-kotlin)
+    * [Drawbacks of Shifting to Kotlin](#drawbacks-kotlin)
+
```
crphang: Table of Contents is generally not used in other book chapters, so it might not be good to introduce a new style in this chapter on Kotlin.

I think we can have less blank spaces here?

contents/kotlin/kotlin.md
```diff
@@ -0,0 +1,87 @@
...
+- Assigning values based on ranges is much simpler
+``` kotlin
+val quartile: Int
+quartile = when (playPercentage) {
+    in 0..24 -> 0
+    in 25..49 -> 1
+    ...
+}
+```
+
```
crphang, line 42: Too many blank lines

contents/kotlin/kotlin.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Safe
+Kotlin protects you from mistakenly operating on nullable types
+- Get compilation error when you mistakenly try operating on nullable types
+```
```
crphang: Put kotlin beside ```

contents/kotlin/kotlin.md
```diff
@@ -0,0 +1,87 @@
...
+import kotlin.browser.window
+fun onLoad() {
+    window.document.body!!.innerHTML += "<br/>Hello, Kotlin!"
+}
+```
+- You can literally continue work on your old Java projects using Kotlin. All your favorite Java frameworks are still available, and whatever framework you’ll write in Kotlin can be easily adopted by other Java developers
+
+### Tool Friendly
+
+Hyperbolically, a programming language is only as good as its tool support. This is why advantage with using Kotlin is that IntelliJ provides built-in language support. It also contains the aforementioned Java-to-Kotlin converter and code generators for Java and JavaScript from Kotlin code
```
crphang: I think paraphrasing some of these can make it clearer. What do you think?

tool support -> is only as good as what its tools can provide.
why advantage with -> why the advantage of using Kotlin is the built-in language support from IntelliJ

Kotlin code -> Kotlin code. (add a fullstop)

contents/kotlin/kotlin.md
```diff
@@ -0,0 +1,87 @@
...
+- Get compilation error when you mistakenly try operating on nullable types
+```
+val name: String? = null    // Nullable type
+println(name.length())      // Compilation error
+```
+
+### Interoperable
+
+- Target either the JVM or JavaScript. Write code in Kotlin and decide where you want to deploy
+```
```
crphang: Similarly, I think you'll want to have Kotlin beside ```

contents/kotlin/kotlin.md
damithc: Good point. I don't mind having some sort of roadmap at the top though. May be call it `Chapter contents` instead of `ToC`?

PierceAndy: If it improves navigability, I don't think having a table of contents is an issue. I used one myself for GPGPU and CUDA.
I don't believe there's a standard format being enforced (I could be wrong about this semester!).

contents/kotlin/kotlin.md
junkiattan: Some comments

1. Is "Hyperbolically" suppose to be "Hypothetically" instead?
2. Better to also state Kotlin support for other commonly used IDEs like Eclipse, so that the scope of "tool-friendliness" can be more general. That way Kotlin would seem even more accessible
3. Your code for the Safe and Interoperable sections are not highlighted like the code in the Concise section. It would be better to highlight them all similarly to guide the reader and make the code stand out better

contents/kotlin/kotlin.md
```diff
@@ -0,0 +1,87 @@
...
+Hyperbolically, a programming language is only as good as its tool support. This is why advantage with using Kotlin is that IntelliJ provides built-in language support. It also contains the aforementioned Java-to-Kotlin converter and code generators for Java and JavaScript from Kotlin code
+
+# Kotlin for Android Apps
+
+While Java is one of the world's most widely used programming languages and is pretty much the official language of Android development, there are many reasons why Java might not always be the best option for your Android projects.
+
+The biggest issue is that Java isn’t a modern language, and although Java 8 was a huge step forward for the platform, introducing lots of features that developers had been waiting for (including lambda functions), at the time of writing Android only supports a subset of Java 8 features.
+
+### Advantages of Shifting to Kotlin
+
```
junkiattan, line 85: 1. Try to only capitalize the first letter of the first word of each bullet. E.g.
> Easy learning curve

Unless the word, e.g. "Java", is a recognized term

2. Try changing your last point to "Smaller community support". In general your bullets should be concise and easy to understand at a glance.

contents/kotlin/kotlin.md
junkiattan: If the page is long, then having a table of contents is good. But for this case, the content is short and can be easily viewed from start to end with a single scroll (if you remove the space taken up by the table of contents and its spaces). Plus it is good to also observe the rough outline used for other chapters as mentioned.

contents/kotlin/kotlin.md
```diff
@@ -0,0 +1,87 @@
...
+    * [Drawbacks of Shifting to Kotlin](#drawbacks-kotlin)
+
+
+
+
+
+
+# Overview
+
+Kotlin is a general purpose, open source, statically typed “pragmatic” programming language for the JVM and Android that combines object-oriented and functional programming features.
```
Haozhe321: I've seen some websites using the word "pragmatic" as well, but I think it adds more confusion than explanation, especially with the " ". The sentence would work just as well without the word.

contents/kotlin/kotlin.md
```diff
@@ -0,0 +1,87 @@
...
+
+# Why Kotlin?
+
+### Concise
+
+Highly reduces the amount of boilerplate code.
+- A POJO with getters, setters, equals(), hashCode(), toString() and copy() can be created in a single line
```
Haozhe321: I think giving the non-acronym form of POJO is better

contents/kotlin/kotlin.md
LiHaoTan: Hmm, I thought the pragmatic makes a huge difference. Kotlin is so attractive because of the practical approach they took instead of trying to become a perfect language, theoretically sound or whatsoever.

contents/kotlin/kotlin.md
```diff
@@ -0,0 +1,87 @@
...
+``` kotlin
+data class Customer(val name: String, val email: String, val company: String)
+```
+- Assigning values based on ranges is much simpler
+``` kotlin
+val quartile: Int
```
Haozhe321: An equivalent of example in Java would make this comparison clearer

Haozhe321 (commented):
I've left some comments :)

contents/kotlin/kotlin.md
Haozhe321: I get what you mean, but the overview should be a clear and concise introduction to the topic. Having the word "programatic" makes the user wonder what it means to be a "pragmatic programming language". So it takes attention away from what Kotlin is at a glance. If anything, this could be explained a little down in the main points, but in the overview it's a little too much to be added in. 

And also, the " " around the word doesn't help with clarifying the meaning of the word.

contents/kotlin/kotlin.md
tshradheya: I think it's better to have the ToC cause most likely there will be more content added to Kotlin and then it will be necessary

contents/kotlin/kotlin.md
tshradheya: As suggested by @Haozhe321 , its better to get rid of the word as it adds more confusion to people completely new to Kotlin

LiHaoTan: I see the point but pragmatic is a common jargon (we can always not use the jargon) and is one of the biggest selling points of Kotlin.

By removing pragmatic, Kotlin now becomes general purpose and open source. So the overview is that it has not many differences with other languages? That's just not attractive.

contents/kotlin/kotlin.md
```diff
@@ -0,0 +1,94 @@
...
+if(playPercentage >= 0 && playPercentage <= 24) {
+    quartile = 0;
+} else if(playPercentage >= 25 && playPercentage <= 49) {
+    quartile = 1;
+}
+...
+```
+
+### Safe
+Kotlin protects you from mistakenly operating on nullable types
```
LuLechuan: Need a new line here to be consistent

contents/kotlin/kotlin.md
```diff
@@ -0,0 +1,94 @@
...
+int quartile;
+if(playPercentage >= 0 && playPercentage <= 24) {
+    quartile = 0;
+} else if(playPercentage >= 25 && playPercentage <= 49) {
+    quartile = 1;
+}
+...
+```
+
+### Safe
```
LuLechuan, line 54: Need to add a new line here to be consistent

tshradheya: I have added a new line everywhere before `###` so it's consistent. 
Am I missing your point in the comment? :P

contents/kotlin/kotlin.md
```diff
@@ -0,0 +1,95 @@
...
+
+Kotlin protects you from mistakenly operating on nullable types
```
damithc: Explain the meaning of `nullable`?

contents/kotlin/kotlin.md
```diff
@@ -0,0 +1,95 @@
...
+- Interchangeability With Java
+- Easy learning curve
+- Combine the best of functional and procedural programming
+- First-class Android Studio support
+- More concise code
+
+### Drawbacks of Shifting to Kotlin
+
+- Extra runtime size due to increase in size of `.apk`
+- Initial readability of code for core Java developers
```
damithc: This is not phrased as a disadvantage.

damithc (changes requested):
Good start. If not in this PR then in a future PR, there should be some pointers on how readers can learn Kotlin. i.e., where to go from here

tshradheya:
>If not in this PR then in a future PR, there should be some pointers on how readers can learn Kotlin.

I will add that in this PR itself maybe by next week.

tshradheya:
Prof @damithc, I have made the necessary changes. 
Can review again to see if everything is fine?

damithc (approved):
LGTM 👍

PR #45, opened 2018-03-08
Null Safety in Kotlin

Haozhe321:
This pull request adds a page on null safety in Kotlin to the resources.


contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,121 @@
+# Null Safety in Kotlin
+
+Author: [Pan Haozhe](https://github.com/Haozhe321)
+
+# Overview
+Another part of this book chapter gives an introduction to Kotlin. This piece aims to discuss the null safety feature in Kotlin, and what was previously missing in Java.
```
joanneong: Comment on first sentence: Hmm actually I was under the impression that this is a book chapter in itself, i.e. you wrote a book chapter on null safety in Kotlin so I thought you meant that you'll be introducing Kotling below (but you did not). Then I thought that maybe you are referencing to another 'book chapter', but I realised that this is the first Kotlin chapter...so I am not sure if you want to even include the sentence "Another part of this book chapter gives an introduction to Kotlin" here. Perhaps it can be added in the future after such a book chapter has really been created...

joanneong: Comment on second sentence: "...what was previously missing in Java" -> I think this is a very vague statement that can leave readers wondering what you are getting at. Is there any way to state clearly what you are talking about here?

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,121 @@
...
+
+# The Billion Dollar Mistake
+“I call it my billion-dollar mistake. It was the invention of the null reference…My goal was to ensure that all use of references should be absolutely safe, with **checking performed automatically by the compiler**. But I couldn't resist the temptation to put in a null reference, simply because it was so easy to implement.”
```
joanneong: Can be put in a block quote (using `>`) instead. Also, although this quote is interesting, it is not immediately apparent how it adds value to your book chapter since you did not comment on this/ make any reference to this anywhere else in the book chapter (which is quite a waste) :/

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,121 @@
...
+Author: [Pan Haozhe](https://github.com/Haozhe321)
+
+# Overview
+Another part of this book chapter gives an introduction to Kotlin. This piece aims to discuss the null safety feature in Kotlin, and what was previously missing in Java.
+
+# The Billion Dollar Mistake
+“I call it my billion-dollar mistake. It was the invention of the null reference…My goal was to ensure that all use of references should be absolutely safe, with **checking performed automatically by the compiler**. But I couldn't resist the temptation to put in a null reference, simply because it was so easy to implement.”  
+-Tony Hoare  
+
+
```
joanneong: An extra spacing here?

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,121 @@
...
+# NullPointerException
+As we know, Kotlin is now supported as the official language of Android applications. When developing Android applications in Java, NullPointerException(NPE) was a big problem. In fact, About one third of app crashes can be attributed to NPE. To see how it happens, let's take a look at the piece of Java code below:
```
joanneong: Can consider putting a hyperlink at `NullPointerException (NPE)`

joanneong: Also, I feel like the first line "As we know, Kotlin is now supported as the official language of Android applications." should not be included in this subsection as this subsection is entirely about NPE in Java and how Java settles NPE. Instead, I see it fitting in better as the first statement of the next subsection (which is on Kotlin).

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,121 @@
...
+                //do something
+            }
+        }
+    }
+}
+```
+The deep-nested `if` statement adds to the verbosity of our code. Let's see how Kotlin deals with this issue.
+
+# Nullable and Non-nullable type
+In Kotlin, a type can be nullable or non-nullable, determined by the presence of a `?`. For example, an object of type `String` is non-nullable, while an object of type `String?` is nullable.
```
joanneong: Does the `?` have a proper name?

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+}
+```
+The deep-nested `if` statement adds to the verbosity of our code. Let's see how Kotlin deals with this issue.
+
+# Nullable and Non-nullable type
+In Kotlin, a type can be nullable or non-nullable, determined by the presence of a `?`. For example, an object of type `String` is non-nullable, while an object of type `String?` is nullable.  
+
+As the compiler catches `null` assignments to non-nullable object, the following would result in compilation error.
```
joanneong: "...to non-nullable object**s**"

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,121 @@
...
+var firstString: String = "hello world"
+firstString = null //compilation error
+```
+In comparison, the following assignment to a nullable type is allowed
+```Kotlin
+var secondString: String? = "hello world"
+secondString = null //okay
+```
+
+In the first case, we can safely call `firstString.length` without having to worry about a NPE because `firstString` can never be `null`. In the second case, `secondString` can potentially be `null`, so `secondString.length` will result in a compilation error as the compiler see the danger of such statement and blocks it early.
```
joanneong: Just to verify, does this mean that doing `secondString.length` will **always** result in a compilation error? Because this is the impression that I am getting from this paragraph here. Even if this is the case, I think it would be nice to make this clear.

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,121 @@
...
+
+# Operators in Kotlin
+Although non-nullable type is a strong feature in Kotlin, the interoperability with Java means that we have to use variables as nullable type sometimes. In the previous section, we seem to have hit an obstacle as the compiler blocks the call to `secondString.length`. In this section we look at some ways of overcoming this problem.
```
joanneong: Hmm you might want to explain briefly what "interoperability with Java" means because I don't know...

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,121 @@
...
+This chain will return `null` if any of the variables inside the chain is `null`.
+
+## Elvis Operators
+Represented by `?:`, the Elvis operator is used in this way
+```kotlin
+val length = secondString?.length ?: -1
+```
+If the expression to the left of `?:` is not null, the Elvis operator will return it as it is; else it will return a default value supplied (-1 in this case).
+
+We also notice the use of safe call operator together with Elvis operator in the same statement.
```
joanneong: Might want to add a "(`?.`)" next to "safe call operator" since readers may be new to the concept and forget that it was just introduced in the previous subsection.

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,121 @@
...
+
+But the Elvis operator is more powerful than this. `return` and `throw` statements are legitimate default values on the right side of the Elvis operator. For example:
```
joanneong: It is unclear why the fact that `return` and `throw` are legitimate default values on the right side of the Elvis operator makes the Elvis operator "powerful".

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,121 @@
...
+```kotlin
+fun myFunc(node: Node): String? {
+    val parent = node.getParent() ?: return null
+    val name = node.getName() ?: throw IllegalArgumentException("Name expected")
+    // ...
+}
+```
+
+At this point you may ask, What if I still want my NPE?
```
joanneong: Should wrap "What if I still want my NPE?" in quotation marks.

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Not-null assertion operator
+Represented by `!!`, the not-null assertion operator is used in this way
+```kotlin
+val stringLength = secondString!!.length
+```
+The operator converts any value to a non-null type and throws an exception if the value is null. In the above example, stringLength will be assigned the length of secondString if secondString is not `null`; if secondString is `null`, a NPE is thrown. In Kotlin, NPE are explicitly asked for.
```
joanneong: Several code references are not formatted here. For example, `stringLength` and `secondString`.

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,121 @@
...
+
+
```
joanneong: Redundant line here...

contents/kotlin/NullSafety.md
joanneong: Although you do explain why not throwing an NPE by default is great, it would be nice to mention here as well since your current explanation is in the **summary** --> should be a consolidation of content which you covered in all these subsections.

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,121 @@
...
+# Summary
+1. Kotlin makes your applications safer by allocating more work to the complier, instead of failing in the hands of the users. 
+2. If you expect your object to **not** take on a `null` value, make it a non-null type!
+3. Even if you make your object a nullable type, Kotlin handles it better than Java because it can help to prevent NPE. An generic NPE is hard to debug; in Kotlin a descriptive message could be given to make debugging easier(with the help of Elvis operator).
+
+
```
joanneong: Actually if you want two spacings then make the front one consistent.

joanneong (changes requested):
Was updating my book chapter so I read this and decided to review this. My review offers the perspective of someone who does not know Kotlin - so maybe you can invite another person who knows Kotlin to read this as well, and see what is lacking.

In general, I think that the examples you chose to give in this book chapter are very simple and reader-friendly, so keep these examples coming :)

However, I find that the book chapter can be better if the subsections are linked more smoothly with each other. Right now, I feel as though the subsections are still quite distinct even though I do sense some attempts to join them into a cogent whole. It would be good if you can spare some time to look into strengthening the connections between the subsections, so that concepts flow to a convincing conclusion.

contents/kotlin/NullSafety.md
LiHaoTan: Like what @joanneong mentioned this quote is interesting.

However, my opinion is that this quote is not concise because the main points are:
-  billion dollar mistake (There are lots of data on null errors, so the number did not came out of nowhere, but we need to add the data behind this?)
- Original intent is that the compiler catches all reference errors but null is only added because of convenience

So to pique interest I think the quote is great (but maybe overused?), but the quote lacks the rationale of why he made such a statement (the facts).

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,121 @@
...
+```java
+String a = null;
+if(a.length > 5) {
+    //do something
+}
+```
+When the above code is ran, a NPE will be thrown on line 2 because a `null` object has no methods. To prevent an object from taking on a `null` value, programmers resort to doing additional checks like this:
+```java
+String a = null;
+if(a != null) {
```
LiHaoTan: This example isn't so good because it has been long agreed that preventing nulls by default is a good thing. See how Java 7 introduced `Objects#requireNonNull(Object)`.

On the other hand, if this value needs to be optional, it is not a good idea to conflate nulls and optional values.

For instance, for pre-Java 8 there's Guava `Optional` and for Java 8 onwards there's also `Optional`.

Then we can do `a.ifPresent(this::doSomething);`

Also, a more readable (common) way of writing a check like this would not be so much nesting:
```java
String a = null;
if(a != null &&a.length > 5) {
    //do something
}
```

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,121 @@
...
+}
+```
+And of course, that's all fine, until we want to do something more complex. Say Bob belongs to a department, and we want to get the name of the department manager. That will look like this:
+```java
+String managerName = bob.department.manager.name;
+```
+
+Because each variable can be `null`, to prevent the NPE we put the code in the following code block:
+```java
+if(bob != null) {
```
LiHaoTan, line 37: Referencing @alexfjw's example in his talk, if we used `Optional` we could just do the following:
```Java
bob.map(Person::getDepartment)
    .map(Person::getManager)
    .map(Person::getName)
    .ifPresent(Person::doSomething);
```

contents/kotlin/NullSafety.md
LiHaoTan: Should have a citation for the app crashes?

LiHaoTan (changes requested):
Some of my thoughts :)

contents/kotlin/NullSafety.md
Haozhe321: @joanneong I was referring to the book chapter that @tshradheya is writing, introduction to Kotlin. After both are merged it would make sense to refer to his one for an overview of Kotlin. 

For the sentence on Java, perhaps I can say "...and the difference between Kotlin and Java in this aspect"?

contents/kotlin/NullSafety.md
Haozhe321: Actually the "billion dollar" is directly quoted from the author himself (the link to his talk is provided in the links section below), where he says "...the amount is probably in the order of an (American) billion - more than a tenth of a billon, less than ten billion." He took about 1hr 15mins to explain this fully, so I think interested reader should go to the link to find out more. I will put a link on the title of "Billion Dollar Mistake".

Haozhe321: There are a few areas where this idea of "check done by compiler" is being expanded. 
1) In the section on nullable and non-nullable types
2) In the summary

contents/kotlin/NullSafety.md
Haozhe321: I removed "As we know, Kotlin is now supported as the official language of Android applications." because it is not useful for the purpose of this book chapter. There's another one on it.

contents/kotlin/NullSafety.md
Haozhe321: Haha the Kotlin documentation doesn't give it a name

contents/kotlin/NullSafety.md
Haozhe321: That's right.

Haozhe321:
Changes made!

Haozhe321:
@joanneong @LiHaoTan can help me take a look again, thanks :)

Haozhe321:
@LiHaoTan @joanneong  a reminder :)

joanneong (approved):
LGTM 👍 Although the link for Java Optionals in your learning resources is the same as the link in the chapter itself - is this intended? :P

LiHaoTan (approved):
LGTM :P I'm just wondering maybe you should place this as a sub-section of Kotlin.

Haozhe321:
Removed the link to Java Optionals at the end of the chapter. I thought the introduction to Kotlin serves a good purpose of a short and brief introduction to get people interested. So this one is more specialised in a sense. 

@damithc this chapter is ready for review!

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,141 @@
+# Null Safety in Kotlin
+
+Author: [Pan Haozhe](https://github.com/Haozhe321)
+
+# Overview
+Another [part](https://github.com/se-edu/learningresources/blob/master/contents/kotlin/kotlin.md) of this series on Kotlin gives an introduction to Kotlin. This piece aims to discuss the null safety feature in Kotlin, and what was previously missing in Java.
```
damithc: Alternative:
>This document explains Kotlin's  _null safety_ feature . For an overview of Kotlin, see [here]().

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,141 @@
...
+
+
+# The Billion Dollar Mistake
+“I call it my billion-dollar mistake. It was the invention of the null reference…My goal was to ensure that all use of references should be absolutely safe, with **checking performed automatically by the compiler**. But I couldn't resist the temptation to put in a null reference, simply because it was so easy to implement.”  
+-Tony Hoare
```
damithc: You can format this as a quote. i.e.,
>“I call it my billion-dollar mistake. It was the invention of the null reference…My goal was to ensure that all use of references should be absolutely safe, with **checking performed automatically by the compiler**. But I couldn't resist the temptation to put in a null reference, simply because it was so easy to implement.” --Tony Hoare

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,141 @@
+# Null Safety in Kotlin
+
+Author: [Pan Haozhe](https://github.com/Haozhe321)
+
+# Overview
+Another [part](https://github.com/se-edu/learningresources/blob/master/contents/kotlin/kotlin.md) of this series on Kotlin gives an introduction to Kotlin. This piece aims to discuss the null safety feature in Kotlin, and what was previously missing in Java.
+
+
+# The Billion Dollar Mistake
```
damithc: This heading is unnecessary

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,141 @@
...
+“I call it my billion-dollar mistake. It was the invention of the null reference…My goal was to ensure that all use of references should be absolutely safe, with **checking performed automatically by the compiler**. But I couldn't resist the temptation to put in a null reference, simply because it was so easy to implement.”  
+-Tony Hoare  
+
```
damithc: Remainder of the document can be structured as follows:
# What is Null Safety?

{Explain null safety, using Java examples}

# Null Safety in Kotlin 

{Explain how Kotlin improves null safety}

damithc (changes requested):
Good start. Some comments added.
Let's tweak the overall structure to make it a better match for the reader's mind set. i.e., instead of structuring it as 'what the author knows about topic X', structure it as 'what the reader is likely to want to know about topic X'. For example, the first question in reader's mind is likely to be 'what is null safety?'

Haozhe321:
@damithc changes made. 
- Added more information on null safety
- Restructured the book chapter. There are 2 main sections, both of the section headings are in question format to match the readers' mindset
- Change section heading of "Null Safety in Kotlin" to "How does Kotlin handle Null Safety?" to avoid confusion with book chapter title

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,142 @@
...
+>“I call it my billion-dollar mistake. It was the invention of the null reference…My goal was to ensure that all use of references should be absolutely safe, with **checking performed automatically by the compiler**. But I couldn't resist the temptation to put in a null reference, simply because it was so easy to implement.”  
+-Tony Hoare
+
+This document explains Kotlin's null safety feature . For an overview of Kotlin, see [here](https://github.com/se-edu/learningresources/blob/master/contents/kotlin/kotlin.md).
+
+# What is Null Safety?
+Null Safety (or void safety) is the guarantee that no object reference will have a `null` value.
+In object-oriented languages, access to objects is achieved through references. A typical function call is of the form `object.func()`; `object` denotes a reference to a certain object, and `func` denotes a function call. At execution time, the reference to `object` can be `void`, leading to run-time exceptions (In the case of Java, a NullPointerException) and often abnormal termination of the program.
+
```
damithc: When introducing a new term, you can use italics. e.g., _Null Safety_

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,142 @@
...
+## NullPointerException
+ When developing Android applications in Java, [NullPointerException(NPE)](https://docs.oracle.com/javase/9/docs/api/java/lang/NullPointerException.html) was a big problem. In fact, About [one third of app crashes can be attributed to NPE](https://image.slidesharecdn.com/droidcon-bugsense-130408170720-phpapp01/95/droid-con-bugsense-16-638.jpg?cb=1365440918). To see how it happens, let's take a look at the piece of Java code below:
+
+```java
+String a = null;
+if(a.length > 5) {
+    //do something
+}
+```
+When the above code is ran, a NPE will be thrown on line 2 because a `null` object has no methods. To prevent an object from taking on a `null` value, programmers typically resort to doing additional checks like this:
```
damithc: ran -> run
a NPE -> an NPE (because it is read as "an en pee ee")

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,142 @@
...
+        if(manager != null) {
+            String name = manager.name;
+            if(name != null) {
+                //do something
+            }
+        }
+    }
+}
+```
+The deep-nested `if` statement adds to the verbosity of our code.
```
damithc: `adds to the verbosity of our code` -> `reduces readability of our code`

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,142 @@
...
+
+The other way is to use [Java Optionals](http://www.oracle.com/technetwork/articles/java/java8-optional-2175753.html).
+For the first example above, we can do
+``` java
+a.ifPresent(this::doSomething);
+```
+
+And for the deeply-nested `if` statements, the verbosity can be reduced with
```
damithc, line 58: A quick mention of why `map` is used here?

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,142 @@
...
+```Kotlin
+var secondString: String? = "hello world"
+secondString = null //okay
+```
+
+In the first case, we can safely call `firstString.length` without having to worry about a NPE because `firstString` can never be `null`.   
+In the second case, `secondString` can potentially be `null`, so `secondString.length` will result in a compilation error as the compiler see the danger of such statement and blocks it early.
+
+
+## Operators in Kotlin
```
damithc: This heading is too general. `Operators used for Null Safety` or something like that?

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,142 @@
...
+
+At this point you may ask, "What if I still want my NPE?"
+
+
+### Not-null assertion operator
+Represented by `!!`, the not-null assertion operator is used in this way
+```kotlin
+val stringLength = secondString!!.length
+```
+The operator converts any value to a non-null type and throws an exception if the value is null. In the above example, `stringLength` will be assigned the length of `secondString` if `secondString` is not `null`; if secondString is `null`, a NPE is thrown. Kotlin tries to reduce the number of NPE as an NPE is difficult to debug a generic NPE, and also it creates so many app crashes. Hence NPEs in Kotlin are explicitly asked for.
```
damithc: This phrase is hard to understand `an NPE is difficult to debug a generic NPE`

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,142 @@
...
+
+
+# Summary
+1. Kotlin makes your applications safer by allocating more work to the complier, instead of failing in the hands of the users.
```
damithc: -> `Kotlin increases null safety of programs because some of the work required to ensure null safety it offloaded from the programmer to the compiler` ?

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,142 @@
...
+2. If you expect your object to **not** take on a `null` value, make it a non-null type!
```
damithc: Did you mean `non-nullable type`?

contents/kotlin/NullSafety.md
```diff
@@ -0,0 +1,142 @@
...
+3. Even if you make your object a nullable type, Kotlin handles it better than Java because it can help to prevent NPE. An generic NPE is hard to debug; in Kotlin a descriptive message could be given to make debugging easier(with the help of Elvis operator).
```
damithc: This point about descriptive messages was not mentioned in the main text?

contents/kotlin/NullSafety.md
damithc: non-null -> non-nullable?

damithc (changes requested):
Overall structure is good now. 👍 Some cosmetic tweaks suggested.

contents/kotlin/NullSafety.md
Haozhe321: It was implied in the section and example on Elvis Operator. I've added a sentence in the Elvis Operator section to make it more obvious.

Haozhe321:
@damithc changes made. Added another summary point which I think will help to differentiate Kotlin's system vs Java's Optionals.

damithc (approved):

damithc:
Good work! 👍

damithc:
@Haozhe321 Looks like this is not linked from the main table of contents. Can you do a quick PR to add it there?

Haozhe321:
Sure!

PR #46, opened 2018-03-10
SQL Injection

Rinder5:
This PR hopes to add a section on SQL Injection to the book. It is pretty similar to the section on XXS though, as I used that as a base. xD

### DIFF
--- a/contents/security/sqlInjection.md
+++ b/contents/security/sqlInjection.md
@@ -1,0 +1,171 @@
+# SQL Injection
+
+Author: [Lewis Koh](https://github.com/nus-cs3281/2018/blob/master/students/lewisKoh/lewisKoh-Resume.md)
+
+## Overview
+
+SQL (Structured Query Language) Injection is a vulnerability found in websites using databases.
+
+SQL is a common language which is used by websites to communicate with databases. Databases can be used to store persistent data, such as usernames and passwords, sensitive account data, or other important information used by the website. Databases are usually made of many "tables" 
+which are organised in rows and columns. Each row is a separate entry in the table, and each column is a specific parameter which can be used by the entry.
+
+As an example, a table in a database may look like this:
+
+|   UserId   |    Username   |Password|
+|:----------:|:-------------:|:------:|
+|      1     |     Admin     | 123456 |
+|      2     |     Alice     | pw1234 |
+|      3     |      ...      |   ...  |
+
+
+SQL is used to interact with the database by sending "queries" which the database responds to. Some common SQL commands used include: 
+* `SELECT` - retrieves information from a table
+* `UPDATE` - changes information from a table
+* `INSERT INTO` - adds a new entry into a table
+* `DELETE FROM` - deletes information from a table
+
+Websites combine SQL syntax with user input to query the underlying database of the website. For example, blog websites may ask for a username and password from users for them to log in to the site. This input is used to create the SQL statement that will be sent to the database. A website which is vulnerable to SQL injection attacks may construct the SQL query like this:
+
+```sql
+String query = "SELECT * FROM Users WHERE Username = ‘" + input_username +"’ 
+                AND Password = ‘" + input_password + "’";
+```
+
+As an example, if a user entered their username, `Alice`, and password, `pw1234` into the website to try to gain access:
+
+```
+Username: Alice
+Password: pw1234
+```
+
+The constructured query would look like this:
+
+```sql
+SELECT * FROM Users WHERE Username = ‘Alice’ 
+AND Password = ‘pw1234’
+```
+
+This query would find all entries in the `Users` table in the database, and return any entries where the `Username` is `Alice` and the `Password` is `pw1234`. If the previous table was searched, it would return the details of the second row as the result. Since there was a result returned, the website would be able to tell that a legitimate username and password combination was entered since the query requires that both are matched to retrieve the data. Thus, the website would know that the user is legitimate, and the user would be allowed to log into the site. On the other hand, if no result was returned by the database, the website would know that the username and password combination does not match any of the users in the database, and would deny access to the person trying to log in.
+
+
+However, are users limited to just filling in legitimate username and password strings? Some websites directly use the information given by the user without any validation, thus the users are able to write code as input for their own malicious purposes. For example, users are able to add parameters to the query.
+
+```sql
+Username: foo
+Password: bar’ OR ‘’=’
+```
+
+The SQL command string built from this input would be as follows:
+
+```sql
+SELECT * FROM Users WHERE Username = ‘foo’ 
+AND Password = ‘bar’ OR ‘’=’’
+```
+
+In SQL, `AND` operations are checked before `OR` operations.
+This query will check the database for entries where:
+
+```
+Username = foo AND Password = bar
+OR
+‘’=’’
+```
+
+This query will always return true, as `‘’=’’` is always true.
+As such, the query can be simplified to this:
+
+```sql
+SELECT * FROM Users
+```
+
+This will return all the rows from the `Users` table in the database, regardless of username or password entered. If used in user authentication, the user will likely be able to gain access to the system.
+
+
+## Preventing SQL Injection
+
+There are a couple of ways to protect your website against SQL injection attacks. The two most common ways are:
+1. Prepared Statements using Parameterized Queries
+1. Whitelist Input Validation
+
+The sections below explain the two options in more detail.
+
+### Solution 1: Prepared Statements using Parameterized Queries
+
+By defining all the SQL code first and passing in the parameters afterwards, you can make the database distinguish the difference between code and data. It would treat the values entered by the user as a parameter, and would not allow it to alter the query being executed. The way to achieve this varies by language, but it is easy to implement and effective.
+
+For example, instead of writing this in Java:
+
+```java
+String query = "SELECT * FROM Users WHERE Username = " + input_username +"’ 
+                AND Password = ‘" + input_password + "’";
+Statement statement = connection.createStatement();
+ResultSet results = statement.executeQuery( query );
+```
+
+You can prepare the statement like this:
+
+```java
+ String query = "SELECT * FROM Users WHERE Username = ? 
+                AND Password = ? ";
+ PreparedStatement pstmt = connection.prepareStatement( query );
+ pstmt.setString( 1, input_username ); 
+ pstmt.setString( 2, input_password );
+ ResultSet results = pstmt.executeQuery( );
+```
+
+The `?` in the query string is a placeholder for a string value. In the example, the first `?` is substituted with the value of `input_username` using the call `pstmt.setString(1, input_username)`. If the earlier attack was attempted, The query being submitted to the database will look like this:
+
+```sql
+SELECT * FROM Users WHERE Username = 'foo' AND Password = 'bar' OR ‘’=’’
+```
+
+However, since it is treated as a value to be used, the database will not allow it to modify the query, and it would not be able to affect the structure of the SQL statement. As such, the query will check the database for entries where:
+
+```
+Username = foo
+AND 
+Password = bar OR ‘’=’’
+```
+
+As such, the query is safe from SQL code being injected by users.
+
+
+### Solution 2: Whitelist Input Validation
+
+By applying a whitelist to the values a user is allowed to use, you can remove undesired symbols in the query being passed to the database 
+(e.g. whitelisting only alphanumeric characters for a username). This ensures that attackers won't be allowed to enter special characters 
+which may have unwanted effects when executed. (e.g. special characters in SQL such as `'`, `@`, `^` and `_`)
+
+
+## Resources
+
+References:
+
+1. https://www.owasp.org/index.php/SQL_Injection
+(Basic description of SQL Injection attack taken from here)
+1. http://projects.webappsec.org/f/WASC-TC-v2_0.pdf (page 105-110)
+(Description of the two types of SQL injection attack taken from here)
+
+Additional Reading Resources:
+
+1. http://www.sqlinjection.net/
+(Good starting point for diving deeper into SQL injection)
+1. https://www.owasp.org/index.php/Testing_for_SQL_Injection_(OTG-INPVAL-005)
+(How to test your website for SQL injection attacks)
+1. https://www.owasp.org/index.php/Query_Parameterization_Cheat_Sheet
+(Parameterized query examples for some common languages)
+1. https://www.owasp.org/index.php/SQL_Injection_Prevention_Cheat_Sheet
+(A list of possible preventions, which contains even more ways to protect your site from SQL injection attacks).
+1. http://guides.rubyonrails.org/security.html#sql-injection
+(Discussion of how SQL injection attacks work, some possible scenarios of such attacks, and possible preventive measures using Ruby on Rails)
+1. https://www.owasp.org/index.php/Blind_SQL_Injection
+(Discussion about a type of SQL injection attack, when only generic messages are displayed by the website)
+1. https://cwe.mitre.org/data/definitions/89.html
+(Comprehensive coverage of the aspects of SQL injection)
+
+Additional Resources:
+
+1. https://github.com/sqlmapproject/sqlmap
+(Penetration testing tool that detects and exploits SQL injection flaws)
+1. https://find-sec-bugs.github.io/
+(Plugin to detect security vulnerabilities in Java web applications, including SQL injection)
--- a/contents/ToC.md
+++ b/contents/ToC.md
@@ -45,6 +45,7 @@
   * [Cross Site Scripting](security/crossSiteScripting/crossSiteScripting.md)
   * [Cryptography](security/cryptography.md)
   * [HTTPS](security/Https.md)
+  * [SQL Injection](security/sqlInjection.md)
 * Multi-Threading
   * [GPGPU](gpgpu/gpgpu.md)
   * [CUDA](gpgpu/cuda.md)


Rinder5:
@tran-tien-dat hope you can help to review this, since you've been doing all the security talks ^.^

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,145 @@
+# SQL Injection
+
+Author: [Lewis Koh](https://github.com/nus-cs3281/2018/blob/master/students/lewisKoh/lewisKoh-Resume.md)
+
+## Overview
+
+SQL (Structured Query Language) Injection is a vulnerability found in websites using databases.
+
```
wenmogu, line 8: I think probably you can introduce the sql query language a bit, since the injection makes use of the query language structure and how the query is parsed

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,145 @@
...
+or password entered. If used in user authentication, the user will likely be able to gain
+access to the system.
+
+
+## Methods of SQL Injection
+
+There are two common methods for attackers to identify and exploit an SQL injection weakness
+in websites:
+
+* SQL Injection - The attacker makes use of errors generated from executed queries. The
```
wenmogu: to me the formatting would look clearer if you use a ordered list instead of the unordered list you are using here, since you mentioned that there are two common methods.

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,145 @@
...
+## Methods of SQL Injection
+
+There are two common methods for attackers to identify and exploit an SQL injection weakness
+in websites:
+
+* SQL Injection - The attacker makes use of errors generated from executed queries. The 
+errors from the database allows the attacker to fine-tune their attack so their next 
+attempt would be more successful.
+
+For example, if when attempting this attack to retrieve data from a separate table:
```
wenmogu: emmm I was a bit confused when i saw the term "a separate table". Probably you can say, to retrieve data from table named "sysobjects"? do correct me if i m wrong orz

wenmogu (commented):
hope u dont mind me reviewing before Dat :D Nice choice of topic and I had an enjoyable read! In general it is quite good, and out of personal preference, i think it would be even better if you can embed the references to the related parts inside the book chapter

contents/security/sqlInjection.md
tran-tien-dat: Yep, I think you should explain some basic concepts of the database like table, select, etc. Also, it will be easier to understand if you show how when a legitimate user logs in to the website, the website just construct a SQL query from the user input, i.e. `"SELECT Username FROM Users WHERE Username = ‘" + input_username +"’
AND Password = ‘" + input_password + "’"`. This will make the explanation of the later part clearer.

contents/security/sqlInjection.md
tran-tien-dat: Yeah, I think you should give more context to the example given. What is known/not known before the attacker sees the error? What information does the attacker gain from seeing the error? How does the attacker use this information to "fine-tune their attack" (as mentioned above)? These questions are not answered clearly here.

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,145 @@
...
+operator must have an equal number of expressions
+in their target lists. 
+```
+
+Then the attacker would know that the separate table exists, and would only need 
+to change the number of columns retrieved to make the query successful.
+
+* Blind SQL Injection - Even when there is no detailed error from the database 
+displayed, attackers can gather information though other means, such as by testing 
+different ture or false statements.
```
tran-tien-dat: spelling of `true` and `through` (1 line above)

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,145 @@
...
+```
+
+However, if this statement was used instead:
+
+```html
+http://example/article.asp?ID=2+and+1=0
+```
+
+Since 1=0 is always false, the page should always return an error. If both queries 
+lead to the same result, then the website is probably safe from SQL injection.
```
tran-tien-dat: I think the point of this 2nd method is that the error message is generic so that the attacker cannot gain the wanted information like in the 1st method. However, there are tricks that allow the attacker to gain the information, i.e. the method you are presenting here: by asking true/false questions. This point is not very clear in the writing. I think to make it easier to understand, you will need to give a more contextualized example.

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,145 @@
...
+## Preventing SQL Injection
+
+There are a couple of ways to protect your website against SQL injection attacks. The two most
+common ways are:
+
+* Using Prepared Statements using Parameterized Queries: By defining all the SQL code first 
+and passing in the parameters afterwards, you can make the database distinguish the difference 
+between code and data. It would treat the values entered by the user as a parameter, and would
+not allow it to alter the query being executed. The way to achieve this varies by language, 
+but it is easy to implement and effective.
```
tran-tien-dat: I think this thing is best explained by example. Just pick a server-side language that support SQL prepared statement and show the code difference between normal SQL queries and SQL prepared statement.

tran-tien-dat (commented):
Thanks for trusting me with this. Overall, I think this is a good basic overview of SQL injection, but some points need more detailed examples in order for those unfamiliar with SQL to understand.

Rinder5:
@wenmogu @tran-tien-dat Thanks so much for all the help! I really appreciate it a lot. I tried to explain the ideas more carefully, with (hopefully) clearer examples. @wenmogu I'm not too sure where is a good place or good way to embed references inside the sections. Do you have a good example I can look at? >.>

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,216 @@
+# SQL Injection #
+
+Author: [Lewis Koh](https://github.com/nus-cs3281/2018/blob/master/students/lewisKoh/lewisKoh-Resume.md)
+
+## Overview ##
+
+SQL (Structured Query Language) Injection is a vulnerability found in websites using databases.
+
+SQL is a common language which is used by websites to communicate with databases. This allows 
+websites to store persistent data, such as usernames and passwords, sensitive account data,
```
tran-tien-dat: I would say the "store persistent data..." part is an explanation of a database. So i would prefer if u make database the subject of the sentence e.g. Databases are programs that store...

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,216 @@
...
+parameter which can be used by the entry.
+
+SQL is used to interact with the database by sending "queries" which the database responds to. 
+Some common SQL commands used include: 
+* SELECT - retrieves information from a table
+* UPDATE - changes information from a table
+* INSERT INTO - adds a new entry into a table
+* DELETE FROM - deletes information from a table
+
+Text boxes are commonly used in forms in websites to ask the user for input. For example,
```
tran-tien-dat: Abrupt change of topic. Say something like let's now look at an actual example of how a website interact with a database to authenticate a user. This is usually done by user entering their username and password in a HTML form on the website.

Actually I don't think the name text box is necessary. The idea is the string comes from user input

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,216 @@
...
+blog websites may provide a text box for owners of the blog to log in to the site. These 
+text boxes are used to query the underlying database of the website. A website which is 
+vulnerable to SQL injection attacks may construct the SQL query like this:
+
+```html
+String query = "SELECT * FROM Users WHERE Username = ‘" + input_username +"’ 
+                AND Password = ‘" + input_password + "’";
+```
+
+This query would look at the "Users" table in the database, and return any entries where the
```
tran-tien-dat: For clearer illustration, you should discuss a specific example like username is "alice" password is "password" and how the sql query will look like

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,216 @@
...
+String query = "SELECT * FROM Users WHERE Username = ‘" + input_username +"’ 
+                AND Password = ‘" + input_password + "’";
+```
+
+This query would look at the "Users" table in the database, and return any entries where the 
+column "Username" matches "input_username" and the column "Password" matches "input_password". 
+The intention may be to allow access to the website if the query returns any entry, as a result 
+would imply a matching username and password.
+
+However, are users limited to just typing in plain text? Some websites do not check the
```
tran-tien-dat: Plain text is usually undertood as unencrypted text in a security context. You should find another way to express the sentence.

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,216 @@
...
+                AND Password = ‘" + input_password + "’";
+```
+
+This query would look at the "Users" table in the database, and return any entries where the 
+column "Username" matches "input_username" and the column "Password" matches "input_password". 
+The intention may be to allow access to the website if the query returns any entry, as a result 
+would imply a matching username and password.
+
+However, are users limited to just typing in plain text? Some websites do not check the 
+input, thus the users are able to use escape strings to alter the query, or even to add
```
tran-tien-dat: I think you shouldn't introduce the idea of escaping special characters here and save it for analysis and mitigation section later. Here is just a presentation of how sql works. You can say the website uses the user input without any modifications

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,216 @@
...
+
+```html
+ORA-01789: query block has incorrect number of result columns.
+```
+
+Then the attacker would have found a table which exists in the database, as the error message 
+has changed. Now that an existing table has been found, the attacker can refine the query further. 
+In this example, the attack failed because the number of columns was inconsistent with the expected 
+result. The attacker may try "SELECT Username, Password" next, which will limit the retrieved data 
+to be two columns, one for each parameter requested.
```
tran-tien-dat: The ending is a bit abrupt. It's not clear what the point of saying all these is?

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,216 @@
...
+Username: foo
+Password: bar’; DROP TABLE Users;--
+```
+
+## Preventing SQL Injection ##
+
+There are a couple of ways to protect your website against SQL injection attacks. The two most
+common ways are:
+
+### Using Prepared Statements using Parameterized Queries ###
```
tran-tien-dat: Double "using"?

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,216 @@
...
+
+Then the attacker knows that the form is vulnerable to following attacks. Once a SQL 
+injection vulnerability is discovered, the attacker can continue to exploit the 
+database as per normal SQL Injection. For example, if the attacker knew of a table 
+in the database named 'Users', the attacker can enter this malicious string, which 
+would delete the entire table and its associated data:
+
+```html
+Username: foo
+Password: bar’; DROP TABLE Users;--
```
tran-tien-dat: I don't think this is an example of a blind sql injection. The purpose of blind sql injection attack is to steal data, not to delete data. The point of blind sql is that the error message is generic so the attacker cannot see the actual data he wants to see

contents/security/sqlInjection.md
tran-tien-dat: And what's the purpose of the attacker?

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,216 @@
...
+```
+
+You can prepare the statement like this:
+
+```html
+ String custname = request.getParameter("customerName");
+ String query = "SELECT account_balance FROM user_data WHERE user_name = ? ";
+ PreparedStatement pstmt = connection.prepareStatement( query );
+ pstmt.setString( 1, custname ); 
+ ResultSet results = pstmt.executeQuery( );
```
tran-tien-dat: You shoukd explain a bit more about the meaning of "?" and its connection to the next statement

tran-tien-dat (changes requested):
I think you are trying to cover too much, especially in the types of sql attack section. In the end, without giving sufficient details, this section is misleading. I suggest that you focus on explaining the other sections in a clear manner, with sufficient details for a student not exposed to database or security to understand first. After that you can look at how long the book chapter is and decide whether to write about more sophisticated sql injection (which is basically the different types of sql injection section is about)

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,216 @@
...
+Password: bar’ OR ‘’=’
+```
+
+The SQL command string built from this input would be as follows:
+
+```html
+SELECT * FROM Users WHERE Username = ‘foo’ 
+AND Password = ‘bar’ OR ‘’=’’
+```
+
```
rachx: maybe use ```sql?
I think that will give you syntax highlighting

Rinder5:
@tran-tien-dat @rachx  Thanks for all the help! I think I agree that the methods are not as important to discuss, and the book chapter was getting a bit too long already, so I removed it for now.

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,149 @@
...
+## Overview ##
+
+SQL (Structured Query Language) Injection is a vulnerability found in websites using databases.
+
+SQL is a common language which is used by websites to communicate with databases. Databases 
+can be used to store persistent data, such as usernames and passwords, sensitive account data, 
+or other important information used by the website. Databases are usually made of many "tables" 
+which are organised in rows and columns. Each row is a separate entry in the table, and 
+each column is a specific parameter which can be used by the entry.
```
tran-tien-dat: I think you can give an example of a table here, like in your talk. To connect it with the later part, maybe a Users table with simple username and password.

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,149 @@
...
+
+```sql
+SELECT * FROM Users WHERE Username = ‘Alice’ 
+AND Password = ‘password’
+```
+
+This query would look at the "Users" table in the database, and return any entries where the 
+column "Username" matches "Alice" and the column "Password" matches "password". 
+The intention may be to allow access to the website if the query returns any entry, as a result 
+would imply a matching username and password.
```
tran-tien-dat: (Continuing from my previous comment) You can referred to the proposed table here.

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,149 @@
...
+
+
+However, are users limited to just filling in the parameters? Some websites directly use the input given,
```
tran-tien-dat: `filling in the paramters` does not sound very precise. Maybe `filling in legitimate username and password strings`

tran-tien-dat: `the input given` is vague. Maybe `the input given by the user without any modifications`.

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,149 @@
...
+```
+
+This query would look at the "Users" table in the database, and return any entries where the 
+column "Username" matches "Alice" and the column "Password" matches "password". 
+The intention may be to allow access to the website if the query returns any entry, as a result 
+would imply a matching username and password.
+
+
+However, are users limited to just filling in the parameters? Some websites directly use the input given, 
+thus the users are able to modify the query, or even to add in their own. This can easily be used
```
tran-tien-dat: I would say `SQL query` to be more precise. Also, the next sentence can be joined here as `for malicious purposes` to make the paragraph more concise. Because, for a  newbie, the high level description is quite cryptic, so I think it's good to make this part concise and quickly introduce the example to illustrate the point.

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,149 @@
...
+ String custname = request.getParameter("customerName");
+ String query = "SELECT account_balance FROM user_data WHERE user_name = ? ";
+ PreparedStatement pstmt = connection.prepareStatement( query );
+ pstmt.setString( 1, custname ); 
+ ResultSet results = pstmt.executeQuery( );
+```
+
+The "?" in the query string is a placeholder for a value. Since it is treated as a value to be used, 
+the database will not allow it to modify the query, and it would not be able to affect the structure 
+of the SQL statement. As such, the query will always execute the command as was intended.
```
tran-tien-dat: I would prefer your old example with 2 `?`, and then line number 4 in the code will use `custname` as a string value for the first occurence of `?`, and if you write `pstmt.setstring(2, ...)` thenit will replace the value for the 2nd occurence of `?`.

Basically I would like to see more detailed explanation  of the substitution line  of code,  and  also a slightly more sophisticated example to illustrate the use of `?`

tran-tien-dat:
I think the overall structure is better now. i just have a few suggestions to make the chapter more digestable.

contents/security/sqlInjection.md
Rinder5: I changed the query so it uses the same one used in the examples earlier on. I'm not too sure how much space I should spend on explaining `?`, because it's not used in every language, so it might not be too relevant rather than just discussing that it exists and helps.

contents/security/sqlInjection.md
Rinder5: I added a table. Not sure if it's what you were expecting, let me know! =)

Rinder5:
@nicholaschuayunzhi as someone who has no prior experience with SQL Injection, please let me know what you think of this PR =)

contents/security/sqlInjection.md
tran-tien-dat: I think even if the other languages don't use `?`, they should use some similar constructs, so it is okay to explain the idea more clearly. My point is that the current explanation `The "?" in the query string is a placeholder for a value` is too short and people might not be able to understand. You can explain it more like "... is a placeholder for a string value. It is later substituted with the string value of `input_username` by the string value by the call `pstmt.setString(1, input_username)`"

Rinder5: Ok, I see your point. Updated it now, thanks! =)

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,181 @@
...
+in to the site. This input is used to create the SQL statement that will be sent to the database. 
+A website which is vulnerable to SQL injection attacks may construct the SQL query like this:
+
+```sql
+String query = "SELECT * FROM Users WHERE Username = ‘" + input_username +"’ 
+                AND Password = ‘" + input_password + "’";
+```
+
+As an example, if a user entered their username, "Alice", and password, "password" into the website 
+to try to gain access, the constructured query would look like this:
```
nicholaschuayunzhi: Can show 
```
Username: Alice
Password: password
```
here.

Also, maybe you could change "password" to something else like pw1234, because
1. password is not restricted to characters
1. easier to phrase

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,181 @@
...
+
+```sql
+SELECT * FROM Users WHERE Username = ‘Alice’ 
+AND Password = ‘password’
+```
+
+This query would look at the "Users" table in the database, and return any entries where the 
+column "Username" matches "Alice" and the column "Password" matches "password". If the previous 
+table was searched, it would return the second row as the result. The website would be able to tell
```
nicholaschuayunzhi: What do you mean by returning the row. And how does getting row give access to the site? If there is no entry, would it mean there's no row?

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,181 @@
...
+```
+
+This query would look at the "Users" table in the database, and return any entries where the 
+column "Username" matches "Alice" and the column "Password" matches "password". If the previous 
+table was searched, it would return the second row as the result. The website would be able to tell 
+that a legitimate username and password was entered, and would let the user log into the site.
+
+
+However, are users limited to just filling in legitimate username and password strings? Some websites 
+directly use the input given by the user without any modifications, thus the users are able to
```
nicholaschuayunzhi: Perhaps you could say.. users are able to write code as input To make it flow better.

Also i think a more accurate phrasing is validation/sanitation instead of modification which suggests that the valid input is also changed.

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,181 @@
...
+Username: foo
+Password: bar’ OR ‘’=’
+```
+
+The SQL command string built from this input would be as follows:
+
+```sql
+SELECT * FROM Users WHERE Username = ‘foo’ 
+AND Password = ‘bar’ OR ‘’=’’
+```
```
nicholaschuayunzhi, line 63: Could you write a non code version of this.

Find all users whose user name is foo and their password can be anything?

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,181 @@
+# SQL Injection #
```
nicholaschuayunzhi: Headers don't need the # at the end

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,181 @@
...
+Since it is treated as a value to be used, the database will not allow it to modify the 
+query, and it would not be able to affect the structure of the SQL statement. As such, 
+the query will always execute the command as was intended.
+
+
+### Whitelist Input Validation ###
+
+By applying a whitelist to the values a user is allowed to use, you can 
+remove undesired symbols in the query being passed to the database 
+(e.g. whitelisting only alphanumeric characters for a username).
```
nicholaschuayunzhi: Give an example of some characters that are not in the whitelist

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,181 @@
...
+ pstmt.setString( 1, input_username ); 
+ pstmt.setString( 2, input_password );
+ ResultSet results = pstmt.executeQuery( );
+```
+
+The "?" in the query string is a placeholder for a string value. In the example, the first "?" 
+is substituted with the value of `input_username` using the call `pstmt.setString(1, input_username)`. 
+Since it is treated as a value to be used, the database will not allow it to modify the 
+query, and it would not be able to affect the structure of the SQL statement. As such, 
+the query will always execute the command as was intended.
```
nicholaschuayunzhi: Using the malicious example, could you show what is the resulting sql query?

nicholaschuayunzhi (commented):
hi @Rinder5. Overall I think its easy to follow. Left some comments to make the chapter clearer to an uninformed reader.

Rinder5:
@nicholaschuayunzhi Thanks for the help!

Rinder5:
@tran-tien-dat, @nicholaschuayunzhi I made some changes since your last comments. Would appreciate if you could let me know what you think of it now. =)

Rinder5:
@damithc can you help me review this please?

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,217 @@
...
+		<td>2</td>
+		<td>Alice</td>
+		<td>pw1234</td>
+	</tr>
+	<tr>
+		<td>3</td>
+		<td>...</td>
+		<td>...</td>
+	</tr>
+</table>
```
damithc: Why not use a markdown table?

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,217 @@
...
+Thus, the website would know that the user is legitimate, and the user would be allowed to log 
+into the site. On the other hand, if no result was returned by the database, the website would know 
+that the username and password combination does not match any of the users in the database, and would 
+deny access to the person trying to log in.
+
+
+However, are users limited to just filling in legitimate username and password strings? 
+Some websites directly use the information given by the user without any validation or 
+sanitization, thus the users are able to write code as input for their own malicious 
+purposes. For example, users are able to add parameters to the query.
```
damithc: Sanitation is a technical term used here as if the reader is supposed to know what it means.

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,217 @@
...
+query, and it would not be able to affect the structure of the SQL statement. As such, 
+the query will check the database for entries where:
+
+```
+Username = foo
+AND 
+Password = bar OR ‘’=’’
+```
+
+As such, the query is safe from external modifications.
```
damithc: Find a better alternative to _external modification_? e.g., safe from SQL code being _injected_ by users.

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,217 @@
...
+
+
+### Whitelist Input Validation ###
+
+By applying a whitelist to the values a user is allowed to use, you can 
+remove undesired symbols in the query being passed to the database 
+(e.g. whitelisting only alphanumeric characters for a username).
+This ensures that attackers won't be allowed to enter special characters 
+which may have unwanted effects when executed. 
+(e.g. special characters in SQL such as `'`, `@`, `^` and `_`)
```
damithc: Remove these extra line breaks. We do not require markdown text to wrap at a certain length.

damithc (changes requested):
Looks good 👍 
Some minor nits

Rinder5:
@damithc Thank you for the feedback! I have made the changes requested, please do take another look and see if it looks better.

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,167 @@
...
+
+|   UserId   |    Username   |Password|
+|:----------:|:-------------:|:------:|
+|      1     |     Admin     | 123456 |
+|      2     |     Alice     | pw1234 |
+|      3     |      ...      |   ...  |
+
+
+SQL is used to interact with the database by sending "queries" which the database responds to. Some common SQL commands used include: 
+* SELECT - retrieves information from a table
```
damithc: Like this?
>* `SELECT` - retrieves information from a table

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,167 @@
...
+As such, the query can be simplified to this:
+
+```sql
+SELECT * FROM Users
+```
+
+This will return all the rows from the Users table in the database, regardless of username or password entered. If used in user authentication, the user will likely be able to gain access to the system.
+
+
+## Preventing SQL Injection ##
```
damithc: Like this?
>## Solution 1: Preventing SQL Injection ##

damithc: We usually don't use `##` at the end of a heading, although it may be valid.

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,167 @@
...
+
+There are a couple of ways to protect your website against SQL injection attacks. The two most common ways are:
```
damithc: Here, you need to give the two solutions. e.g.,
>The two most common ways are:
>1. abc
>1. xyz
>
>The two sections below explains the two options in some more details.

contents/security/sqlInjection.md
```diff
@@ -0,0 +1,167 @@
...
+In SQL, `AND` operations are checked before `OR` operations.
+This query will check the database for entries where:
+
+```
+Username = foo AND Password = bar
+OR
+‘’=’’
+```
+
+This query will always return true, as ‘’=’’ is always true.
```
damithc: Use code formatting when you refer to code. e.g.,
This query will always return true, as `‘’=’’` is always true.

damithc (changes requested):
A few more nits.

Rinder5:
@damithc Thanks for the input! Made the changes requested.

PR #47, opened 2018-03-11
REST architectural style

wenmogu:
This PR is to add a new book chapter about REST. This book chapter is basically a summary of some parts of the academic paper that brought about REST. This PR is pretty long (and probably very dry...) but i hope it is an easy-to-understand introduction to REST. (Hope you can) Enjoy!


darrenwee:
Not a review, but you may want to double-check your git config settings since your identity looks a little weird to git and GitHub. It shows up in the patch as

```
From: Mushroom Wen <Wen Xin@DESKTOP-IVHHP1V.(none)>
```

Need to run this inside Git Bash
```
$ git config --global user.name "John Doe"
$ git config --global user.email johndoe@example.com
```

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,226 @@
+# REST Architectural Style
+
+Author: [Wen Xin](https://github.com/wenmogu)
+
+This book chapter introduces the REST architectural style, which is the guideline of development for the Web we use today. The following content is a summary of selected parts from the famous [Roy Fielding's paper about REST](https://www.ics.uci.edu/~fielding/pubs/dissertation/fielding_dissertation.pdf) with my own understanding, which I hope can serve as a relatively brief, easy-to-understand and at the same time detailed introduction to REST.
```
joanneong: I think it should be "...guideline **for** the development **of** the Web..."

joanneong: As a beginner, I have no idea who Roy Fielding is, nor why his paper is so important to the REST architectural style. 

Additionally, I think it is unnecessary to mention the phrase "with my own understanding" (although this may be the truth) because phrases like give a less 'professional' impression.

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,226 @@
...
+
+**Requirement of Web development --- Background of REST architectural style**
```
joanneong: You might want to consider using headers instead (e.g. `##`) to better distinguish section headers from section content...Right now it's the same size as everything else in the chapter. Same goes for all the section headings below in this chapter.

joanneong: Also, is it possible to shorten this header to maybe something like just "Background of REST architectural style"?

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,226 @@
...
+
+The forefather of Web Sir Tim Berners-Lee envisioned the Web to be &quot;a shared information space through which people and machines could communicate.&quot; Fielding&#39;s understanding of &quot;this shared information space&quot; is that the Web is a place where everyone&#39;s information is structured and stored in a way such that their own information &quot;could be usable to themselves and others&quot;, and they can &quot;reference and structure&quot; others&#39; information such that they do not need to keep a local copy of it.
```
joanneong: Since there are several inline quotes present within this single paragraph, why not extract everything into a blockquote, followed by your analysis afterwards? This can allow for cleaner presentation, as well as provide a clearer structure for the content. It may also help with the issue where you seem to be chaining too many quotes together, resulting in long sentences that are somewhat difficult to digest (e.g. the second sentence which spans three lines).

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,226 @@
...
+
+The design of the Web needs to consider the properties of its users and their information. The target audience of the Web are people in different organizations all over the world connected via the Internet. Their machines to access the Web might be very different, with different operating systems and requesting different file formats. These people&#39;s information also has a large variety in terms of the content type and the format. These restraints give rise to the following requirements of the Web:
```
joanneong: How does this paragraph relate to what Sir Tim Berners-Lee envisioned? It would be nice if the transition is made clearer.

joanneong: "As such, the machines people use to access the Web might be very different..."

joanneong: I believe you mean "constraint" rather than "restraint"

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,226 @@
...
+
+- Low entry barrier: the Web should have a simple, uniform interface for everyone to use
+- Extensibility: the Web service should be able to evolve to cater to the users&#39; evolving needs
+- Distributed Hypermedia: Hypermedia is a medium for information like graphics, audio, video, plain text and hyperlinks. The Web should use hypermedia as the user interface &quot;because of its simplicity and generality&quot;. Hypermedia has &quot;the same interface&quot; which can be used &quot;regardless of the information source&quot;, and can be used to structure information flexibly. Moreover, the hyperlinks embedded can guide the reader through the information. However, since hypermedia contains information like graphics, audio etc., the amount of information to be transferred is huge. Hence, the Web should be able to transfer a large amount of information within acceptable time limits.
```
joanneong: Are all the inline quotes from Fielding's paper? :/

joanneong: The point on distributed hypermedia can definitely be made clearer - there are actually only two main points (distributed and hypermedia) but as the paragraph grew haphazardly the main points have been obscured.

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,226 @@
...
+- Internet Scale:
+  - Anarchic scalability: There is no central entity to control the whole system. Hence, firstly, the Web elements should be able to cope with unexpected load. As a result, it is not feasible for the clients to keep the knowledge about all the servers, and for the servers to keep track of the state of information at each of their clients. It is also costly to keep &quot;back-pointers&quot; inside the hypermedia data elements to refer back to the elements which reference them, as potentially there can be a huge number of references. Secondly, the Web elements should be able to ensure security against malformed or maliciously structured, which requires the Web to be able to &quot;communicate authentication data and authorization controls&quot;.
+  - Independent Deployment: to cater to the needs of different users, the Web has to allow the co-existence of both old and new implementations of the Web elements and undergo &quot;gradual and fragmented change&quot;. Hence the design of the Web should be able to ease the deployment of the Web elements &quot;in a partial, iterative fashion, since it is not possible to force deployment in an orderly manner&quot;.
+
```
joanneong: Unfortunately I find this section almost incomprehensible for a beginner like myself. 

Granted, I can get a sense of the direction in which you are heading, but there are a whole load of foreign terms and abrupt references to concepts that were not mentioned beforehand. It is probably not wise to assume that readers know what things like "load", "back-pointers", "reference", "malformed [sic]", "authentication data and authorization controls" etc...

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,226 @@
...
+**REST architectural style**
+
+Fielding defined architectural style as &quot;a coordinated set of architectural constraints that restricts the roles/features of architectural elements and the allowed relationships among those elements within any architecture that conforms to that style&quot;.
```
joanneong: Is this paragraph even necessary at all? It feels like it does not add any value - conversely, it may even confuse readers.

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,226 @@
...
+
+After surveying some common architectures for network-based applications on how the architectural constraints induce their corresponding architectural properties, Fielding came up with a new architectural style – Representational State Transfer, which satisfies the requirements of the Web development.
+
+The table below describes the REST architectural constraints and the architectural properties these constraints induce. Detailed information about the architectural properties can be found in **Appendix**.
+
+<table>
```
joanneong: A few comments:
1) Why has the order of the constraints changed? Why not keep them in accordance with the order established in the previous section for easier reference?
2) What does the column on "Explanation" explain? Why the different constraints exist in the Web?
3) Inconsistencies in capitalisation for "Properties induced" column

In fact, is it possible to further simplify the content in the table, so that it is more easy to digest?

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,226 @@
...
+			</ul>
+		</td>
+	</tr>
+</table>
+
+The 4 interface constraints defining the Uniform Interface architectural constraint for REST are elaborated in the table below:
+
+<table>
+	<tr>
+		<th> Interface Constraint </th>
```
joanneong: What's the difference between these constraints and REST constraints above? What are "Interface Constraints"?

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,226 @@
...
+					the server does not drive the change. Hence, the server does not need to store the state of resources across different requests 
+				</li> 
+			</ul>
+		</td>
+	</tr>
+</table>
+
+**Appendix: Summary of Part of Chapter 2**
+
+Architectural properties of key interest in general:
```
joanneong: What is key interest? Instead of having an appendix section, why not have a section for other related resources so that this book chapter is more in line with what the entire learning resources sets out to accomplish?

joanneong (changes requested):
It is apparent that you have spent a fair bit of time reading up this topic (even plowing through that intimidating Fielding's paper!) :P

Unfortunately, the resulting technical depth may have caused several major issues:
1) Difficult to understand: I find myself rereading sections of the book chapter (and sometimes still going ??? after doing that)
2) Unclear takeaways: By the end of the book chapter, I am still uncertain about what it is **exactly** that I should take away from this book chapter
3) Dry reading: As you have noted yourself, this makes for some pretty dry reading...

While it's definitely interesting to see Fielding's revolutionary ideas in this chapter, I think you may want to diversify your sources, and also consider what exactly you want readers to learn from this chapter. This will also be more aligned with [the learning resources's goal](https://github.com/se-edu/learningresources/blob/master/GuidelinesForContributors.md), which is to "help SE students learn a relevant topic quickly by locating good learning resources".

contents/architecture/RESTArchitecturalStyle.md
LuLechuan: Perhaps you can explain at the start that REST stands for Representation State Transfer and a brief definition for it.

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,226 @@
...
+  - Independent Deployment: to cater to the needs of different users, the Web has to allow the co-existence of both old and new implementations of the Web elements and undergo &quot;gradual and fragmented change&quot;. Hence the design of the Web should be able to ease the deployment of the Web elements &quot;in a partial, iterative fashion, since it is not possible to force deployment in an orderly manner&quot;.
+
+**REST architectural style**
+
+Fielding defined architectural style as &quot;a coordinated set of architectural constraints that restricts the roles/features of architectural elements and the allowed relationships among those elements within any architecture that conforms to that style&quot;.
+
+After surveying some common architectures for network-based applications on how the architectural constraints induce their corresponding architectural properties, Fielding came up with a new architectural style – Representational State Transfer, which satisfies the requirements of the Web development.
```
LuLechuan: Maybe can shift the definition of Representation State Transfer at the start and explain what it do. Then the background story of why develop this style can be put into a section (background) just after the intro of REST

contents/architecture/RESTArchitecturalStyle.md
LuLechuan: Same as @joanneong , a bit confuse about the term "key interest". In addition, are the following only refer to REST architecture? Or they are traits that all architectures should follow?
Since REST is an alternative to SOAP, can you provide the advantages of REST over SOAP as well?

contents/architecture/RESTArchitecturalStyle.md
Haozhe321: This sentence is quite a mouthful because it contains several long words. Try breaking up the sentence into bitesize.

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,226 @@
+# REST Architectural Style
```
Haozhe321, line 1: It'd be good to have links to other resources for students to refer to, just like the other book chapters

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,226 @@
...
+		<td>
+			Separation of concerns: separate the user interface concern from the data storage concern  
+		</td>
+		<td>
+			<ul>
+				<li>
+					Portability of user interface across platforms which might differ in their data storage structures
+				</li>
+				<li>
+					Free the server from managing the user interface =&gt; simplified server component =&gt; enhanced scalability
```
Haozhe321: `=>` introduced here is a little confusing. It could mean an arrow, or it could mean greater than (in this context, more important than), and I think as computing students we are prone to think it is used as a greater than or equals to operator. :arrow_right: find this symbol [here](https://gist.github.com/rxaviers/7360908)

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,67 @@
+# REST Architectural Style
+
+Author: [Wen Xin](https://github.com/wenmogu)
+
+## Definition:
+Representational State Transfer (REST) is the architecture style of the World Wide Web we are using today.
+
+## Why learn REST: 
+It is the set of guiding principles of the Web development (i.e. how the Web should be developed such that it can be efficient and widely applicable). It also serves as the model for Web protocols e.g. HTTP and the guiding principles for RESTful API. Hence, understanding REST will give us an idea of the rationale behind the Web design and let us have a better understanding of the key properties of RESTful API, which would come in handy when we need to use other people’s RESTful APIs and implement our own RESTful APIs.
```
joanneong: Should probably be either:

"It is the set of guiding principles of Web development..." or "It is the set of guiding principles of the Web's development..."

joanneong: Same for "...the rationale behind the Web design..."

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,67 @@
...
+
```
joanneong: Might want to introduce who Roy Fielding is, e.g. "Roy Fielding - the originator of the REST architectural style - states in his paper..."

joanneong: While the example of the` CS3281StudentList` is helpful, it can be made even clearer. For instance, you can give an example of how the student list might change over time so that it is easier for a casual reader to understand.

joanneong: You can also consider splitting the part after the StudentList example (i.e. "As the combination of...") since the section after the StudentList example is more about **why** we should have the abstraction "resource" while the previous section is more about **what** is the abstraction "resource".

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,67 @@
...
+Resource and the term “representational state” are closely related to the client-model server outlined as one of the rules of REST. Resource is the information stored at the server, whereas what the clients see when they access the resource are actually the “representations” of the resource at a specific point of time, i.e. the “representational state” of the resource. Giving the users the representation of the resource instead of the resources itself allows [dynamic binding](https://en.wikipedia.org/wiki/Late_binding) of the reference to a representation of the resource, and this enables the users to have access to and operate on this resource in the formats they want, such as JSON, HTML, XML etc. When the user wishes to make some changes to the resource, he/she can change the representation of this resource and send the representation back to server to update the resource. As such, the server is freed from managing different [application states](https://ruben.verborgh.org/phd/hypermedia/#the-statelessness-constraint-p-8) across requests. 
+
```
joanneong: "Resource" and the term...

joanneong: Isn't it "client-server model" instead of "client-model server"? Also, you can consider adding a hyperlink to the client-server model.

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,67 @@
...
+ * the Web should have a simple, uniform interface to present various types of information (e.g. video, audio, graphics, text etc.) 
+ * the Web should be efficient in transmitting information 
+ * the Web should be able to evolve
+ * the Web should not have a central entity to control the whole system, which means 1) the Web elements should be able to cope with unexpected load 2) the Web should be able to communicate authentication data and authorization controls
```
joanneong: "...1) the Web elements should be able to cope with unexpected load 2) the Web should be able to communicate authentication data and authorization controls" -> this is quite unclear. If you do not wish to expand these points, at least provide hyperlinks for the crucial terms (e.g. "load", "authentication data", "authorization controls" etc.)

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,67 @@
...
+ * the Web should allow its element to undergo incremental changes for evolution 
+REST is the abstraction of the design that satisfy the above requirements.
```
joanneong: The last line is not displaying properly in a new line (use the GitHub's view feature/ render this in your browser to see)

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Details of REST
+
+Fielding defined architectural style as “a coordinated set of architectural constraints that restricts the roles/features of architectural elements and the allowed relationships among those elements within any architecture that conforms to that style”. 
+After surveying some common architectures for network-based applications on how the architectural constraints induce their corresponding architectural properties, Fielding came up with REST. REST has 6 constraints which aim to induce the properties the Web should have. 
+
```
joanneong: >Fielding defined architectural style as “a coordinated set of architectural constraints that restricts the roles/features of architectural elements and the allowed relationships among those elements within any architecture that conforms to that style”. 
After surveying some common architectures for network-based applications on how the architectural constraints induce their corresponding architectural properties, Fielding came up with REST.

Is this part necessary here? How does it relate to the rest of the section? Personally, I find the quote somewhat obtuse...

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,67 @@
...
+### Client-Server
+A system in REST style should separate the user interface concern from the data storage concern. As such, the server is freed from managing the user interface and the user interface is detached from the server. This separation of concern allows the server to evolve without impacting the user interface and makes upgrading the server easier. It also enables the system to have a uniform interface regardless of the different data structures of the servers.
+
```
joanneong: I am not too sure whether "...data structures of the servers..." is the right phrasing...

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,67 @@
...
+### Stateless
+A system in REST style should have stateless communications. This means that all the messages must be self-sufficient, containing all the necessary information to understand the message without referring to information outside the message. By making all the messages self-sufficient, the workload of the server is reduced as the server does not have to keep track of the state of application at the clients. The coupling between the server and the client is further reduced by the statelessness of the system, and thus the server can be scaled up and down according to the amount of workload (e.g. the number of requests from the client at certain point of time). Moreover, by eliminating cross referencing to other messages when interpreting one, this reduces the chances of error and increases reliability of the system. The statelessness of the system also allows the system to make use of intermediaries.  
+
```
joanneong: >The statelessness of the system also allows the system to make use of intermediaries.

What does this mean?

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,67 @@
...
+### Layered System
+There should be multiple layers between client and server. The layers are made up by various [intermediaries](https://www.techopedia.com/definition/24378/web-intermediary-wbi) between the client and the server facilitating the processing tasks. Some examples of intermediaries are load-balancer, cache etc. The system elements have no knowledge of the things outside their own layers. For example, a client would not know if it is connected directly to the server or to an intermediary, and an intermediary would not know if it is connected to another intermediary. By limiting the scope of the system element, the complexity of the system is greatly reduced. The system becomes more modifiable as there are less dependencies between the system elements. By facilitating the processing tasks, intermediaries can enhance the server performance by reducing the redundant server processing. Moreover, as the intermediaries can carry out a wide range of tasks including encryption and request conversion, the organizations can use intermediaries to enhance their security and their control over the information. 
+
```
joanneong: The link for "intermediaries" here can be linked earlier!

joanneong: Links to "encryption" and "request conversion"?

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,67 @@
...
+### Uniform Interface
+There should be a way for the server, the client and the intermediaries in the layers in between to communicate with each other. Hence, there should be a uniform interface in the system. The existence of the uniform interface is the foundation for the other 4 architectural constraints. Each component is encapsulated by the interface and hence become more independent of each other, allowing each to evolve independent of the rest. By having a uniform interface in the system, interactions between the layers can be monitored as the set of interactions are predefined. By allowing the interactions to be inspected by mediators (e.g., network firewalls), the security of the system is enhanced. However, the existence of the uniform interface might compromise the efficiency of the system as the information is transmitted in a standard format rather than catering to each component’s needs.
+There are four sub-constraints which further specify the Uniform Interface constraint.
```
joanneong: Your lines are all not appearing on new lines when rendered.

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,67 @@
...
+ * Identification of resources: as mentioned before, “resource” is an organization of information and the identifiers of the resources need to remain constant. An example to illustrate this constraint is [URI](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier). 
+ * Manipulation of resources through representations: as mentioned before, the “representation” of the resource at one point i.e. the representation of the state of the resource, is what the users see and modify, instead of the resource itself. 
+ * Self-descriptive messages: A message in a communication between the web components should contain all the information needed for the web components to understand its content. 
+ * Hypermedia as the engine of application state (HATEOAS): There should be hyperlinks embedded inside the representations given to the client, such that all the future actions that the client might take are within these representations. Hence, the client can interact with and navigate through the application without any prior knowledge of how to do so. Hence, the client and the server are more independent of each other. 
+
+### Code On Demand (Optional)
```
joanneong: What does the "Optional" here mean?

joanneong (changes requested):
I think it is a lot clearer this time as compared to the first version :)

Because this topic is highly theoretical in nature, it can get a bit dry and dense at times. I guess this is hard to avoid, and one way to mitigate this is to provide even more relevant hyperlinks for relevant concepts (see comments below). Otherwise, it is probably up to the reader's personal motivation to understand this topic further :P

contents/architecture/RESTArchitecturalStyle.md
wenmogu: emmm to me it seems like if i add a hyperlink to title "client-server model", i m compelled to add hyperlinks to all such titles... 🤣

wenmogu:
@joanneong im ready for another round of review thx 🙏

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,76 @@
+# REST Architectural Style
+
+Author: [Wen Xin](https://github.com/wenmogu)
+
+## Definition:
+Representational State Transfer (REST) is the architecture style of the World Wide Web we are using today.
+
+## Why learn REST: 
+It is the set of guiding principles of the development of the Web (i.e. how the Web should be developed such that it can be efficient and widely applicable). It also serves as the model for Web protocols e.g. HTTP and the guiding principles for RESTful API. Hence, understanding REST will give us an idea of the rationale behind the design of the Web and let us have a better understanding of the key properties of RESTful API, which would come in handy when we need to use other people’s RESTful APIs and implement our own RESTful APIs. 
+
```
joanneong: >Hence, understanding REST will give us an idea of the rationale behind the design of the Web and let us have a better understanding of the key properties of RESTful API, which would come in handy when we need to use other people’s RESTful APIs and implement our own RESTful APIs.

This is quite a mouthful - can you break it down into two parts?

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,76 @@
...
+## <a name="meaningOfRest"></a>Meaning of REST:
+REST stands for “Representational State Transfer”. In this section, I will explain the meaning of this name.
+Before we understand the term “representational state”, we have to understand the term “resource”. In the context of REST, the term “resource” is used as the way to organize and represent data/information. Roy Fielding -- the originator of REST, states in his paper that “any information that can be named can be a resource”, and the identification of resources is done by using unique identifiers of the resources i.e. unique name of the resources. The information represented by one resource can change at any time, but the name of the resource i.e. the identifier of the resource has to remain the same for identification purposes. For example, the student list of the course CS3281 can be named as “CS3281StudentList”, and hence can be a resource. The student list might change over time, e.g. from those who enrolled in Year 2017 to those who enrolled in Year 2018, but the identifier “CS3281StudentList” always refers to the student list regardless of its state/value.
```
joanneong: I think the proper grammar is "Roy Fielding - the originator of REST - states in his paper that..."

I have never really seen the usage of "--" in prose before...

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,76 @@
...
+
+REST is the abstraction of the design that satisfies the above requirements.
+
+## Details of REST
+
+REST is an architectural style. Fielding defined architectural style as “a coordinated set of architectural constraints that restricts the roles/features of architectural elements and the allowed relationships among those elements within any architecture that conforms to that style”. 
+
+After surveying some common architectures for network-based applications on how the architectural constraints induce their corresponding architectural properties, Fielding came up with REST. REST has 6 constraints which aim to induce the properties the Web should have. 
+
+### <a name="clientServer"></a>Client-Server
```
joanneong: Hmm...actually I think using [Markdown Anchors](https://gist.github.com/asabaylus/3071099) should work as well.

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,76 @@
...
+
+### <a name="cache"></a>Cache
+Data within the messages for communications between the server and the client should be indicated if it is cacheable. If cacheable, the caches at the elements along the line of communication (e.g. the client cache, the server cache, the proxy cache etc.) will store the data and reuse if for identical requests later. The cacheability of information reduces the amount of interactions needed between the client and the server to access information, and thus improves the efficiency of the system. The average latency of interactions is also reduced, which leads to faster response to the client and an improved user-perceived performance.
```
joanneong: "...will store the data and reuse **it** for identical..."

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,76 @@
...
...d server. The layers are made up by various [intermediaries](https://www.techopedia.com/definition/24378/web-intermediary-wbi) between the client and the server facilitating the processing tasks. Some examples of intermediaries are load-balancer, cache etc. The system elements have no knowledge of the things outside their own layers. For example, a client would not know if it is connected directly to the server or to an intermediary, and an intermediary would not know if it is connected to another intermediary. By limiting the scope of the system element, the complexity of the system is greatly reduced. The system becomes more modifiable as there are less dependencies between the system elements. By facilitating the processing tasks, intermediaries can enhance the server performance by reducing the redundant server processing. Moreover, as the intermediaries can carry out a wide range of tasks including [encryption](https://en.wikipedia.org/wiki/Encryption) and request modification and etc.(see "MEGs" in [this article](http://www.almaden.ibm.com/cs/wbi/doc/Architecture.html)), the organizations can use intermediaries to enhance their security and their control over the information.
```
joanneong: Is it the server facilitating the processing tasks or the intermediaries? What are you currently implying in your paragraph here?

joanneong: "...intermediaries can enhance the server performance by reducing ~the~ redundant server processing..."

joanneong: "...and their control over ~the~ information."

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,76 @@
...
+* Web intermediaries
+	* [Web intermediaries explained](http://www.almaden.ibm.com/cs/wbi/doc/Architecture.html)
+	* an important type of intermediaries: [proxy](https://en.wikipedia.org/wiki/Proxy_server#Filtering_of_encrypted_data)
+* Hypermedia as the engine of application state (HATEOAS)
+	* [A general wikipedia explanation](https://en.wikipedia.org/wiki/HATEOAS)
+	* [A more detailed explanation](https://ruben.verborgh.org/phd/hypermedia/#hypermedia-as-the-engine-of-application-state)
+* Code On Demand
+	* [a general wikipedia explanation](https://en.wikipedia.org/wiki/Code_on_demand)
+	* [a stackoverflow explanation](https://stackoverflow.com/questions/32094952/code-demand-constraint-for-restful-apis?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa)
+* [Video explaning REST (easy to get the big picture but not very precise)](https://www.youtube.com/watch?v=YCcAE2SCQ6k)
```
joanneong: Good range of links although the "Web intermediaries explained" one has already been referenced elsewhere in the chapter.

joanneong (changes requested):
Almost there!

contents/architecture/RESTArchitecturalStyle.md
wenmogu: ummm i dont really understand how it works tho...tried a few times and cannt get it work...
what I understand from the website is, you dont need the <a> tag and just simply put `#meaning-of-rest` at the point where I want to jump to `Meaning of REST`. Is my understanding wrong?

what i will do for now is just replaced the original name e.g. `MeaningOfRest` to `meaning-of-rest`...

wenmogu:
@joanneong ready for a third round of review 🌵

joanneong (approved):
@wenmogu You can either get one more person to review this, or request Prof to take a look at this. You might want to resolve the merge conflict first though...

wenmogu:
@damithc Prof can u help me take a look at this? thx!

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,76 @@
+# REST Architectural Style
+
+Author: [Wen Xin](https://github.com/wenmogu)
+
+## Definition:
+Representational State Transfer (REST) is the architecture style of the World Wide Web we are using today.
+
+## Why learn REST: 
+It is the set of guiding principles of the development of the Web (i.e. how the Web should be developed such that it can be efficient and widely applicable). It also serves as the model for Web protocols e.g. HTTP and the guiding principles for RESTful API. Hence, understanding REST will give us an idea of the rationale behind the design of the Web and let us have a better understanding of the key properties of RESTful API. The knowledge about REST would come in handy especially when we need to use other people’s RESTful APIs and implement our own RESTful APIs.
```
damithc: At this point the reader doesn't know what a RESTful API is.

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,76 @@
...
+
+Author: [Wen Xin](https://github.com/wenmogu)
+
+## Definition:
+Representational State Transfer (REST) is the architecture style of the World Wide Web we are using today.
+
+## Why learn REST: 
+It is the set of guiding principles of the development of the Web (i.e. how the Web should be developed such that it can be efficient and widely applicable). It also serves as the model for Web protocols e.g. HTTP and the guiding principles for RESTful API. Hence, understanding REST will give us an idea of the rationale behind the design of the Web and let us have a better understanding of the key properties of RESTful API. The knowledge about REST would come in handy especially when we need to use other people’s RESTful APIs and implement our own RESTful APIs. 
+
+## <a name="meaning-of-rest"></a>Meaning of REST:
```
damithc: There is no need to add this anchor manually. Markdown adds the anchor automatically for headings.

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,76 @@
...
+REST stands for “Representational State Transfer”. In this section, I will explain the meaning of this name.
```
damithc: Given the section heading, this sentence is redundant?
>In this section, I will explain the meaning of this name.

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,76 @@
...
+Before we understand the term “representational state”, we have to understand the term “resource”. In the context of REST, the term “resource” is used as the way to organize and represent data/information. Roy Fielding, the originator of REST, states in his paper that “any information that can be named can be a resource”, and the identification of resources is done by using unique identifiers of the resources i.e. unique name of the resources. The information represented by one resource can change at any time, but the name of the resource i.e. the identifier of the resource has to remain the same for identification purposes. For example, the student list of the course CS3281 can be named as “CS3281StudentList”, and hence can be a resource. The student list might change over time, e.g. from those who enrolled in Year 2017 to those who enrolled in Year 2018, but the identifier “CS3281StudentList” always refers to the student list regardless of its state/value.
```
damithc: se-edu/learningresources is a public repo. Readers are not guaranteed to know abut CS3281. Use a more general example.

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,76 @@
...
+
+Resource and the term “representational state” are closely related to the client-server model outlined as one of the rules of REST. Resource is the information stored at the server, whereas what the clients see when they access the resource are actually the “representations” of the resource at a specific point of time, i.e. the “representational state” of the resource. Giving the users the representation of the resource instead of the resources itself allows [dynamic binding](https://en.wikipedia.org/wiki/Late_binding) of the reference to a representation of the resource, and this enables the users to have access to and operate on this resource in the formats they want, such as JSON, HTML, XML etc. When the user wishes to make some changes to the resource, he/she can change the representation of this resource and send the representation back to server to update the resource. As such, the server is freed from managing different [application states](https://ruben.verborgh.org/phd/hypermedia/#the-statelessness-constraint-p-8) across requests.
```
damithc: Somewhere in this paragraph you need to bring in the term `transfer` to complete the definition.

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Why REST exists
+REST specifies the key designs of the Web. The design of the Web affects greatly how efficient, applicable and modifiable the Web can be. Hence, before understanding the REST principles, we need to understand the requirements of the Web i.e. what the Web is expected to be. 
+
+The Web is expected to be a place for convenient storing and sharing of information by its users. Hence, to make the ideal Web, we need to consider the properties of its users and their information. The target audience of the Web are people in different organizations all over the world connected via the Internet. Their organizations would have different requirements for their information (e.g. the authentication for access of information), and the machines they use to access the Web might be very different, with different operating systems and requesting different file formats. These people’s information also has a large variety in terms of the content type and the format. Hence, the Web needs to have the following properties:
```
damithc: If the content was based on some other source, remember to cite the source.

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,76 @@
...
+ * the Web should have a simple, uniform interface to present various types of information (e.g. video, audio, graphics, text etc.) 
+ * the Web should be efficient in transmitting information 
+ * the Web should be able to evolve
+ * the Web should not have a central entity to control the whole system, which means there is no control over the amount and the content of information outside the boundary of the organizations using the Web. Hence, 1) the Web elements should be able to cope with unexpected load 2) the Web should be able to communicate authentication data and authorization controls to sieve out the malicious information
+ * the Web should allow its element to undergo incremental changes for evolution
```
damithc: Where did you get this list? Cite the source?

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,76 @@
...
+ * the Web should have a simple, uniform interface to present various types of information (e.g. video, audio, graphics, text etc.) 
+ * the Web should be efficient in transmitting information 
+ * the Web should be able to evolve
+ * the Web should not have a central entity to control the whole system, which means there is no control over the amount and the content of information outside the boundary of the organizations using the Web. Hence, 1) the Web elements should be able to cope with unexpected load 2) the Web should be able to communicate authentication data and authorization controls to sieve out the malicious information
+ * the Web should allow its element to undergo incremental changes for evolution 
+
+REST is the abstraction of the design that satisfies the above requirements.
+
+## Details of REST
```
damithc: Rename to  `REST Requirements` or something like that?

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,76 @@
+# REST Architectural Style
+
+Author: [Wen Xin](https://github.com/wenmogu)
+
+## Definition:
```
damithc: No need for for `:` at the end of each heading.

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,76 @@
...
+ * Hypermedia as the engine of application state (HATEOAS): There should be hyperlinks embedded inside the representations given to the client, such that all the future actions that the client might take are within these representations. Hence, the client can interact with and navigate through the application without any prior knowledge of how to do so. Hence, the client and the server are more independent of each other. 
+
+### <a name="code-on-demand"></a>Code On Demand (Optional)
+The server can send a code snippet to the client to let the client execute. One example of this is the Javascript code sent along with the webpage in HTML. This constraint extends the client functionality and reduces the workload of the server by reducing the number of features to be implemented at the server. However, it reduces the visibility of the interactions between the client and the server and makes it more difficult to monitor the interactions. Hence, this constraint might be disabled in some implementations in REST style.
+
```
damithc: You can consider merging some of the sections to form these three main sections:
>## What is REST
>## Why REST
>## How REST Works

damithc (changes requested):
Very good effort in describing a difficult topic. Well done! Some suggestions added

contents/architecture/RESTArchitecturalStyle.md
wenmogu: I am actually more inclined to my current structure. After merging the sections, the structure will look like this:
## What is REST
### Definition
### Meaning of REST
### Background of REST (why REST exists)
## Why learn REST
## How REST Works

I feel that before the reader sees the reason why he should continue reading this (dry) article i.e. Why learn REST, he might give up halfway on trying to understand REST :( That's why i feel that I should give a one-liner introduction then (try to) hook the reader with a reason why he should continue

contents/architecture/RESTArchitecturalStyle.md
wenmogu: Emmm I feel that it is easy for the readers to misunderstand the term `REST Requirements` as they can also interpret it as `the requirements that REST needs to serve`. I think `Details of REST` is clear in its meaning as it tells the reader below is the content of REST.

contents/architecture/RESTArchitecturalStyle.md
wenmogu: This whole section is actually a very brief summary of the entire section 4.1 in Roy Fielding's paper... I will say something like the sentence below at the end of this section.  Does that look ok?

More details for the above requirements can be found in [Chapter 4.1 WWW Application Domain Requirements](https://www.ics.uci.edu/~fielding/pubs/dissertation/web_arch_domain.htm) in Roy Fielding's paper about REST.

contents/architecture/RESTArchitecturalStyle.md
wenmogu: I tried 
"## Meaning of REST"

and (see "[Meaning of REST]""(#meaning-of-rest)")

It doesnt work :(  Did i misunderstand anything

contents/architecture/RESTArchitecturalStyle.md
wenmogu: Yup I just added a one-liner explanation and embed a link at the term there. Hope that suffices :D

wenmogu (commented):
thx @damithc for the review! Emm but I face some problem with the markdown header anchor and citing the source for a whole section... i think i need some help on it

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,75 @@
...
+Author: [Wen Xin](https://github.com/wenmogu)
+
+## Definition
+Representational State Transfer (REST) is the architecture style of the World Wide Web we are using today.
+
+## Why learn REST 
+It is the set of guiding principles of the development of the Web (i.e. how the Web should be developed such that it can be efficient and widely applicable). It also serves as the model for Web protocols e.g. HTTP and the guiding principles for [RESTful API](https://openclassrooms.com/courses/build-your-web-projects-with-rest-apis/examples-of-rest-apis), the type of API that most of the companies use to provide their web services. Hence, understanding REST will give us an idea of the rationale behind the design of the Web and let us have a better understanding of the key properties of RESTful API. The knowledge about REST would come in handy especially when we need to use other people’s RESTful APIs and implement our own RESTful APIs. 
+
+## <a name="meaning-of-rest"></a>Meaning of REST
```
damithc: Can remove the explicit anchors. `see [this](#meaning-of-rest)` should work. Previously it didn't work because you had a colon at the end of the heading.

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,75 @@
...
+
+The quote below from Roy Fielding outlines what the name entails:
+> The name “Representational State Transfer” is intended to evoke an image of how a well-designed Web application behaves: a network of web pages (a virtual state-machine), where the user progresses through the application by selecting links (state transitions), resulting in the next page (representing the next state of the application) being transferred to the user and rendered for their use.
+
```
damithc: One concern I have is the headings such as _Meaning of X_, _Details of X_ etc. appear a bit "unpolished" (we don't usually see such headings in books). Slight tweak  of headings and content you can consider is given below. You can tweak further to make the flow smoother:

---

# REST Architectural Style

Author: [Wen Xin](https://github.com/wenmogu)

## Overview

REST stands for _Representational State Transfer_. It is the architecture style of the World Wide Web. In particular,
* It embodies a set of guiding principles the Web should follow such that it can be efficient and widely applicable
* It serves as the model for Web protocols e.g. HTTP
* It describes the nature of [RESTful APIs](https://openclassrooms.com/courses/build-your-web-projects-with-rest-apis/examples-of-rest-apis), the type of API that are widely used to provide/consume services over the Web

Let us learn a bit more about REST so that we can better understand the rationale behind the design of the Web and better design/consume RESTful APIs. 

## What exactly is REST?

### Resource

The first step to understand REST is to understand the term “resource”. In the context of REST, the term “resource” is used as the way to organize and represent data/information. Roy Fielding, the originator of REST, states in his paper that “any information that can be named can be a resource”, and the identification of resources is done by using unique identifiers of the resources i.e. unique name of the resources. The information represented by one resource can change at any time, but the name of the resource i.e. the identifier of the resource has to remain the same for identification purposes. For example, the student list of a computer science course with course code CS3281 can be named as “CS3281StudentList”, and hence can be a resource. The student list might change over time, e.g. from those who enrolled in Year 2017 to those who enrolled in Year 2018, but the identifier “CS3281StudentList” always refers to the student list regardless of its state/value.

As the combination of various types (video, audio, hyperlinks, text…) of information into one unit are becoming more common (e.g. in an online article teaching the reader how to use IntelliJ, there would be text describing the process, videos giving demonstrations and hyperlinks directing the reader to other articles), using an encompassing term to represent a unit of information makes sense as that’s how the user organizes information. Moreover, this abstraction saves us the trouble of distinguishing the types or implementation of information inside this unit when thinking of the process of accessing and transmitting this unit of information.

### Representational State

Resource and the term “representational state” are closely related to the client-server model outlined as one of the rules of REST. Resource is the information stored at the server, whereas what are given/transfered to the clients when they access the resource are actually the “representations” of the resource at a specific point of time, i.e. the “representational state” of the resource. 

### Representational State Transfer

Transfering to the users the representation of the resource instead of the resources itself allows [dynamic binding](https://en.wikipedia.org/wiki/Late_binding) of the reference to a representation of the resource, and this enables the users to have access to and operate on this resource in the formats they want, such as JSON, HTML, XML etc. When the user wishes to make some changes to the resource, he/she can change the representation of this resource and send the representation back to server to update the resource. As such, the server is freed from managing different [application states](https://ruben.verborgh.org/phd/hypermedia/#the-statelessness-constraint-p-8) across requests. 

The quote below from Roy Fielding outlines what the name entails:
> The name “Representational State Transfer” is intended to evoke an image of how a well-designed Web application behaves: a network of web pages (a virtual state-machine), where the user progresses through the application by selecting links (state transitions), resulting in the next page (representing the next state of the application) being transferred to the user and rendered for their use.

## How does REST Make the Web Better?

REST specifies the key designs of the Web. The design of the Web affects greatly how efficient, applicable and modifiable the Web can be. Hence, before understanding the REST principles, we need to understand the requirements of the Web i.e. what the Web is expected to be. 

The Web is expected to be a place for convenient storing and sharing of information by its users. Hence, to make the ideal Web, we need to consider the properties of its users and their information. The target audience of the Web are people in different organizations all over the world connected via the Internet. Their organizations would have different requirements for their information (e.g. the authentication for access of information), and the machines they use to access the Web might be very different, with different operating systems and requesting different file formats. These people’s information also has a large variety in terms of the content type and the format. Hence, the Web needs to have the following properties:
 * the Web should have a simple, uniform interface to present various types of information (e.g. video, audio, graphics, text etc.) 
 * the Web should be efficient in transmitting information 
 * the Web should be able to evolve
 * the Web should not have a central entity to control the whole system, which means there is no control over the amount and the content of information outside the boundary of the organizations using the Web. Hence, 1) the Web elements should be able to cope with unexpected load 2) the Web should be able to communicate authentication data and authorization controls to sieve out the malicious information
 * the Web should allow its element to undergo incremental changes for evolution 

REST is the abstraction of the design that satisfies the above requirements. More details for the above requirements can be found in [Chapter 4.1 WWW Application Domain Requirements](https://www.ics.uci.edu/~fielding/pubs/dissertation/web_arch_domain.htm) in Roy Fielding's paper about REST.

## REST as an Architectural Style

REST is an architectural style. Fielding defined architectural style as “a coordinated set of architectural constraints that restricts the roles/features of architectural elements and the allowed relationships among those elements within any architecture that conforms to that style”. 

After surveying some common architectures for network-based applications on how the architectural constraints induce their corresponding architectural properties, Fielding came up with REST. REST has 6 constraints which aim to induce the properties the Web should have. 

### <a name="client-server"></a>Client-Server
A system in REST style should separate the user interface concern from the data storage concern. As such, the server is freed from managing the user interface and the user interface is detached from the server. This separation of concern allows the server to evolve without impacting the user interface and makes upgrading the server easier. It also enables the system to have a uniform interface regardless of the different structures of data storage at the servers.

### <a name="stateless"></a>Stateless
A system in REST style should have stateless communications. This means that all the messages must be self-sufficient, containing all the necessary information to understand the message without referring to information outside the message. By making all the messages self-sufficient, the workload of the server is reduced as the server does not have to keep track of the state of application at the clients. The coupling between the server and the client is further reduced by the statelessness of the system, and thus the server can be scaled up and down according to the amount of workload (e.g. the number of requests from the client at certain point of time). Moreover, by eliminating cross referencing to other messages when interpreting one, this reduces the chances of error and increases reliability of the system. The statelessness of the system also allows the system to make use of intermediaries as the intermediaries in between the client and the server have all the necessary information to complete their tasks (see [Layered System](#layered-system)).  

### <a name="cache"></a>Cache
Data within the messages for communications between the server and the client should be indicated if it is cacheable. If cacheable, the caches at the elements along the line of communication (e.g. the client cache, the server cache, the proxy cache etc.) will store the data and reuse it if for identical requests later. The cacheability of information reduces the amount of interactions needed between the client and the server to access information, and thus improves the efficiency of the system. The average latency of interactions is also reduced, which leads to faster response to the client and an improved user-perceived performance.

### <a name="layered-system"></a>Layered System
There should be multiple layers between client and server. Between the client and the server, there are layers made up by various [intermediaries](https://www.techopedia.com/definition/24378/web-intermediary-wbi) facilitating the processing tasks. Some examples of intermediaries are load-balancer, cache etc. The system elements have no knowledge of the things outside their own layers. For example, a client would not know if it is connected directly to the server or to an intermediary, and an intermediary would not know if it is connected to another intermediary. By limiting the scope of the system element, the complexity of the system is greatly reduced. The system becomes more modifiable as there are less dependencies between the system elements. By facilitating the processing tasks, intermediaries can enhance the server performance by reducing redundant server processing. Moreover, as the intermediaries can carry out a wide range of tasks including [encryption](https://en.wikipedia.org/wiki/Encryption) and request modification and etc.(see "MEGs" in [this article](http://www.almaden.ibm.com/cs/wbi/doc/Architecture.html)), the organizations can use intermediaries to enhance their security and their control over information. 

### <a name="uniform-interface"></a>Uniform Interface
There should be a way for the server, the client and the intermediaries in the layers in between to communicate with each other. Hence, there should be a uniform interface in the system. The existence of the uniform interface is the foundation for the other 4 architectural constraints. Each component is encapsulated by the interface and hence become more independent of each other, allowing each to evolve independent of the rest. By having a uniform interface in the system, interactions between the layers can be monitored as the set of interactions are predefined. By allowing the interactions to be inspected by mediators (e.g., network firewalls), the security of the system is enhanced. However, the existence of the uniform interface might compromise the efficiency of the system as the information is transmitted in a standard format rather than catering to each component’s needs.

There are four sub-constraints which further specify the Uniform Interface constraint.
 * Identification of resources: as mentioned before (see [Meaning Of REST](#meaning-of-rest)), “resource” is an organization of information and the identifiers of the resources need to remain constant. An example to illustrate this constraint is [URI](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier). 
 * Manipulation of resources through representations: as mentioned before (see [Meaning Of REST](#meaning-of-rest)), the “representation” of the resource at one point i.e. the representation of the state of the resource, is what the users see and modify, instead of the resource itself. 
 * Self-descriptive messages: A message in a communication between the web components should contain all the information needed for the web components to understand its content. 
 * Hypermedia as the engine of application state (HATEOAS): There should be hyperlinks embedded inside the representations given to the client, such that all the future actions that the client might take are within these representations. Hence, the client can interact with and navigate through the application without any prior knowledge of how to do so. Hence, the client and the server are more independent of each other. 

### <a name="code-on-demand"></a>Code On Demand (Optional)
The server can send a code snippet to the client to let the client execute. One example of this is the Javascript code sent along with the webpage in HTML. This constraint extends the client functionality and reduces the workload of the server by reducing the number of features to be implemented at the server. However, it reduces the visibility of the interactions between the client and the server and makes it more difficult to monitor the interactions. Hence, this constraint might be disabled in some implementations in REST style.

## Useful Resources
* [Roy Fielding's paper which gave birth to REST](https://www.ics.uci.edu/~fielding/pubs/dissertation/fielding_dissertation.pdf)
* [History of hypermedia and REST explained](https://ruben.verborgh.org/phd/hypermedia/)
* Web intermediaries
	* [Web intermediaries explained](http://www.almaden.ibm.com/cs/wbi/doc/Architecture.html)
	* an important type of intermediaries: [proxy](https://en.wikipedia.org/wiki/Proxy_server#Filtering_of_encrypted_data)
* Hypermedia as the engine of application state (HATEOAS)
	* [A general wikipedia explanation](https://en.wikipedia.org/wiki/HATEOAS)
	* [A more detailed explanation](https://ruben.verborgh.org/phd/hypermedia/#hypermedia-as-the-engine-of-application-state)
* Code On Demand
	* [a general wikipedia explanation](https://en.wikipedia.org/wiki/Code_on_demand)
	* [a stackoverflow explanation](https://stackoverflow.com/questions/32094952/code-demand-constraint-for-restful-apis?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa)
* [Video explaning REST (easy to get the big picture but not very precise)](https://www.youtube.com/watch?v=YCcAE2SCQ6k)

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,75 @@
...
+## <a name="meaning-of-rest"></a>Meaning of REST
+REST stands for “Representational State Transfer”. Before we understand the term “representational state”, we have to understand the term “resource”. In the context of REST, the term “resource” is used as the way to organize and represent data/information. Roy Fielding, the originator of REST, states in his paper that “any information that can be named can be a resource”, and the identification of resources is done by using unique identifiers of the resources i.e. unique name of the resources. The information represented by one resource can change at any time, but the name of the resource i.e. the identifier of the resource has to remain the same for identification purposes. For example, the student list of a computer science course with course code CS3281 can be named as “CS3281StudentList”, and hence can be a resource. The student list might change over time, e.g. from those who enrolled in Year 2017 to those who enrolled in Year 2018, but the identifier “CS3281StudentList” always refers to the student list regardless of its state/value.
```
damithc: When you introduce a new term, you can use _italics_. Alternatively, use single quotes, but not double quotes.

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,75 @@
...
+
+### <a name="code-on-demand"></a>Code On Demand (Optional)
+The server can send a code snippet to the client to let the client execute. One example of this is the Javascript code sent along with the webpage in HTML. This constraint extends the client functionality and reduces the workload of the server by reducing the number of features to be implemented at the server. However, it reduces the visibility of the interactions between the client and the server and makes it more difficult to monitor the interactions. Hence, this constraint might be disabled in some implementations in REST style.
+
+## Useful Resources
+* [Roy Fielding's paper which gave birth to REST](https://www.ics.uci.edu/~fielding/pubs/dissertation/fielding_dissertation.pdf)
+* [History of hypermedia and REST explained](https://ruben.verborgh.org/phd/hypermedia/)
+* Web intermediaries
+	* [Web intermediaries explained](http://www.almaden.ibm.com/cs/wbi/doc/Architecture.html)
+	* an important type of intermediaries: [proxy](https://en.wikipedia.org/wiki/Proxy_server#Filtering_of_encrypted_data)
```
damithc: Why so much indentation?

damithc (changes requested):
Some tweaks to headings suggested.

wenmogu:
@damithc thx prof for the detailed review! yup I think u were right...the anchors r working now! I did some minor changes to the sub titles. I feel the structure u gave was indeed clearer and gives a better orientation to the readers than my original structure. 

Ready for another round of review thx!

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,88 @@
+# REST Architectural Style
+
+Author: [Wen Xin](https://github.com/wenmogu)
+
+## Overview 
+
+REST stands for *Representational State Transfer*. It is the architecture style of the World Wide Web. In particular,
+* It embodies a set of guiding principles the Web should follow such that it can be efficient and widely applicable.
+* It serves as the model for Web protocols e.g. HTTP.
+* It decribes the nature of [RESTful API](https://openclassrooms.com/courses/build-your-web-projects-with-rest-apis/examples-of-rest-apis),  the type of API that are widely used to provide/consume services over the Web.
```
damithc: typo: decribes

contents/architecture/RESTArchitecturalStyle.md
```diff
@@ -0,0 +1,88 @@
...
+
+The quote below from Roy Fielding outlines what the name entails:
+> The name “Representational State Transfer” is intended to evoke an image of how a well-designed Web application behaves: a network of web pages (a virtual state-machine), where the user progresses through the application by selecting links (state transitions), resulting in the next page (representing the next state of the application) being transferred to the user and rendered for their use.
+
+## REST defines what the Web is expected to be
```
damithc: This is fine but it fits better with its peer headings if you can phrase it as a question `How does REST Define the Web?`, or as a topic phrase e.g., `REST as an Abstraction of the Web`. Right now, it is an assertion/statement. While it is fine to use assertions/statements as headings, it is odd to have only one such heading while all other headings are of a different form.

damithc (changes requested):
a typo

damithc (approved):
LGTM 👍 
Good job persisting with this. REST is not easy to explain. This article is a significant step in that direction.

wenmogu:
thx @damithc for the guidance!

PR #48, opened 2018-03-13
Introduction to Go

craaaa:
Added a section that introduces Go; view the rendered version [here](https://github.com/craaaa/learningresources/blob/intro-to-go/contents/go/Go.md)!

@darrenwee -- since you did your lightning talk on Go, would appreciate if you could take a look!

### DIFF
--- a/contents/go/Go.md
+++ b/contents/go/Go.md
@@ -2,7 +2,7 @@
 
 Author(s): [Cara Leong](https://github.com/craaaa)
 
-- [Why Learn Go?](why-learn-go)
+- [Why Learn Go?](#why-learn-go)
 - [Go Features](#go-features)
 	- [Declaring Variables](#declaring-variables)
 	- [Concurrency](#concurrency)
--- a/contents/go/Go.md
+++ b/contents/go/Go.md
@@ -171,7 +171,7 @@ func useFile() {
 
 To deal with unexpected errors, Go also provides two mechanisms: `panic` and `recover`.
 
-- `panic` is similar to throwing an exception in other languages. An explicit call to `panic` on a function `F` stops the ordinary flow of execution of `F` at the point of the panic, executes any functions deferred by `F`, and returns to `F`'s caller as a call to `panic`. The `panic` recursively propagates up the call stack until all functions in the goroutine have returned, after which the program crashes. `panic` is used to fail fast on errors that cannot be handled gracefully.
+- `panic` is similar to throwing an exception in other languages. An explicit call to `panic` on a function `F` stops the ordinary flow of execution of `F` at the point of the panic, executes any functions deferred by `F`, and returns to `F`'s caller. To the caller, `F` behaves like a call to panic. It triggers a `panic` in `F`'s caller, which recursively propagates up the call stack until all functions in the goroutine have returned, after which the program crashes. `panic` is used to fail fast on errors that cannot be handled gracefully.
 - `recover` regains control of a panicking goroutine. Using `recover` is comparable to catching an exception in C++ or Java. When used inside a deferred function, a call to recover captures the value returned by `panic` and resumes normal execution.
 
 More information on error handling can be found on the [Go blog](https://blog.golang.org/error-handling-and-go) or [Go wiki](https://github.com/golang/go/wiki/PanicAndRecover).
--- a/contents/go/Go.md
+++ b/contents/go/Go.md
@@ -1,0 +1,141 @@
+# Introduction to Go
+
+Go (also known as `golang`) is a compiled, statically-typed, garbage-collected language that has special memory safety and concurrent programming features. Born out of frustration with the available languages (e.g. C, C++, Java) and environments for systems programming, Go was [conceptualized by programmers at Google](https://talks.golang.org/2012/splash.article) who sought to create a single language that was efficient to write, build and execute. Go also supports newer developments in computing such as multicore processors and network systems.
+
+Go is an [open source project](https://github.com/golang/go). Its source code may be useful reading for those interested in learning good practices, or simply to find out more about how the language was implemented.
+
+## Getting started
+
+Go provides its own [installation guide](https://golang.org/doc/install) and an interactive [ tour of Go](https://tour.golang.org/). These are useful and highly comprehensive resources for programmers looking to learn the syntax and style of Go. For those who prefer to read existing code examples, [Go by Example](https://gobyexample.com/) is a collection of code samples covering a wide variety of features in Go, and includes line-by-line explanations of the code. For those looking for a quick crash course on Go syntax, the Learn X In Y Minutes [Go cheatsheet](https://learnxinyminutes.com/docs/go/) may also be a good starting point.
+
+If you're unconvinced about Go, [the Go playground](https://play.golang.org/) can be used as a sandbox in which to write, build and execute code without installing Go on your machine.
+
+## Using Go
+As it builds on the foundations set by many popular and widely-used languages such as C, C++, Java and Python, much of Go's syntax draws from existing implementations and will be familiar to programmers looking to learn an additional language. However, Go also diverges explicitly from these other languages; listed below are some aspects of Go that may be unfamiliar to learners.
+
+### Arrays and slices
+Arrays are not often seen in Go code. This is because the size of a Go array must be declared at its creation, which limits how flexible an array can be.
+
+To create dynamically-sized arrays, Go introduces the concept of a slice. Slices are not arrays; rather, they are data structures that describe a piece of a separately-stored array. Slices are the topic of several Go blog entries; you can read more about [the internal layout of slices](https://blog.golang.org/go-slices-usage-and-internals) and [the mechanics of using slices](https://blog.golang.org/slices). Exercises on how to use slices are available in [A Tour of Go](https://tour.golang.org/moretypes/7).
+
+### `defer`, `panic` and `recover`
+In addition to traditional control flow mechanisms such as `if`, `for` and `switch`, Go also provides three other ways to alter the flow of a program: `defer`, `panic`, `recover`. These mechanisms can be conceived in terms of exception handling in Java and C++, but also have more general purpose uses.
+
+- `defer` pushes a function call to a list, and executes all functions on the list in last-in-first-out order after the surrounding function returns. It is frequently used for clean-up actions, such as to close files.
+- `panic` stops the ordinary flow of execution, executes any deferred functions, and returns to its caller as a call to `panic`. The `panic` continues up the stack until all functions in the goroutine have returned, after which the program crashes.
+- `recover` regains control of a panicking goroutine. When used inside a deferred function, a call to recover captures the value returned by `panic` and resumes normal execution.
+
+More information on these three mechanisms and their uses can be found on the [Go blog](https://blog.golang.org/defer-panic-and-recover).
+
+### Interfaces
+Although Go has types and methods and allows pseudo-object-oriented style of programming, type hierarchy does not exist in Go. Instead, Go uses interfaces to specify methods that types should implement, grouping types by functionality rather than using notions of inheritance. Types implement interfaces by implementing the methods in the interface, and do not need to explicitly specify which interfaces are implemented.
+
+In the example below, the `Rectangle` type implements the interface `TwoDimensional` by implementing the methods `area()` and `perim()` that are specified in the interface.
+
+```go
+type TwoDimensional interface {
+  area() float64
+  perim() float64
+}
+
+type Rectangle struct {
+  width, height float64
+}
+
+func (r Rectangle) area() {
+  return r.width * r.height
+}
+
+func (r Rectangle) perim() {
+  return r.width * 2 + r.height * 2
+}
+```
+
+One benefit of using a system where interface implementations need not be stated in the source code is that methods can be attached to types that you didn't write. In other words, you can extend a type to implement an interface without access to its source code by simply implementing the interface's method in your own code.
+
+Some resources to get started with Go interfaces include [this blog post](https://medium.com/golangspec/interfaces-in-go-part-i-4ae53a97479c) introducing Go interfaces, code examples on
+[how interfaces (including the empty interface) are used in practice](https://www.calhoun.io/how-do-interfaces-work-in-go/), this [comparison of Go's OOP style with that of other languages](https://flaviocopes.com/golang-is-go-object-oriented/) and [Go's official FAQ on OOP](https://golang.org/doc/faq#Is_Go_an_object-oriented_language).
+
+### Concurrency
+One of Go's special features is its support for concurrency. To this end, Go's standard library comes with two features that allow for easy and maintainable concurrency.
+
+#### Goroutines
+A goroutine is a lightweight thread that executes a function concurrently with its caller. A goroutine is launched by a `go` statement:
+
+```go
+func run() {
+  \\ does something
+}
+
+func main() {
+  go run()
+  executeOtherCommands()
+}
+```
+
+In the above example, using the `go` keyword launches a goroutine, which executes `run()` concurrently with `executeOtherCommands()`.
+
+Goroutines can also be started for anonymous functions:
+```go
+go func(msg string) {
+  fmt.Println(msg)
+}("a message")  // starts a goroutine that prints "a message"
+```
+
+A goroutine is not its own thread; instead, goroutines are dynamically multiplexed onto threads as required to keep them running. This makes them lightweight, so having a large number of goroutines is feasible. In practice, goroutines behave similarly to very cheap threads.
+
+#### Channels
+
+Channels are used to fulfill Go's philosophy on concurrent software: "don't communicate by sharing memory; share memory by communicating".
+
+Channels are typed conduits that allows goroutines to communicate with each other by sending and receiving messages. Before using a channel of a specific type, we must declare and `make` it:
+
+```go
+var c chan int
+c = make(chan int)
+
+//alternatively
+c := make(chan int)
+```
+
+Channels can transmit data of any type; thus, creating a channel that transmits channels (i.e. a `chan chan`) is theoretically possible and may even be [useful](http://tleyden.github.io/blog/2013/11/23/understanding-chan-chans-in-go/).
+
+Goroutines send and receive messages through a channel using the `<-` operator.
+
+```go
+ch <- v    // Send v to channel ch.
+v := <-ch  // Receive from ch, and
+           // assign value to v.
+```
+One useful way to think about sending and receiving data with the `<-` operator is that the data moves in the direction of the arrow.
+
+If you are interested in delving deeper into using Go's concurrency features extensively, Google developers have put out video presentations on Go's [basic](https://www.youtube.com/watch?v=f6kdp27TYZs) and [advanced concurrency patterns](https://www.youtube.com/watch?v=QDDwwePbDtw). This [code walkthrough](https://golang.org/doc/codewalk/sharemem/) provides an annotated example of how Go's memory-sharing principles can be applied in practice, .
+
+### Effective Go
+Formatting in Go is enforced by running `gofmt`, which will align your source code with the language-wide standard style of indentation and vertical alignment. Thus, given the following code:
+
+```go
+type T struct {
+    name string // name of the object
+    value int // its value
+}
+```
+Running `$ gofmt` in the same directory as the source file will line up comments and correct the source code's indentation:
+
+```go
+type T struct {
+    name    string // name of the object
+    value   int    // its value
+}
+```
+Variations on `gofmt` may be of use, and can be found in the [documentation](https://golang.org/cmd/gofmt/).
+
+Go also enforces good coding practices, for instance, by refusing to build projects that declare of unused variables or imports. Such enforcement, along with a clear, unified and extensive [treatise on coding conventions in Go](https://golang.org/doc/effective_go.html), have manifested in a reasonably stable Go coding style.
+
+## Useful resources
+Go's development team is heavily involved in documenting and growing the Go language and community. If you are keen to learn more about Go, here are some resources to help you get started:
+
+- [The Go FAQ](https://golang.org/doc/faq) - answers common questions about the language's history, usage, design and more
+- [Go's documentation](https://golang.org/doc/) - a good starting point, contains links to official information about Go
+- [The Go Blog](https://blog.golang.org/) - features news and in-depth articles about Go by the Go team and guests
+- [A list of reasons why Go is bad](https://github.com/ksimka/go-is-not-good) - informative counter-arguments against the use of Go; worth considering before you jump into the world of Go
--- a/contents/go/Go.md
+++ b/contents/go/Go.md
@@ -2,17 +2,19 @@
 
 Author(s): [Cara Leong](https://github.com/craaaa)
 
+- [What Is Go?](#what-is-go)
 - [Why Learn Go?](#why-learn-go)
-- [Go Features](#go-features)
+- [What Makes Go Special?](#what-makes-go-special)
 	- [Declaring Variables](#declaring-variables)
 	- [Concurrency](#concurrency)
 	- [Error Handling](#error-handling)
 	- [`defer`](#defer)
 	- [Interfaces](#interfaces)
 	- [Canonical Coding Style](#canonical-coding-style)
-- [Getting started](#getting-started)
-- [Useful Resources](#useful-resources)
+- [How Do I Start Using Go?](#how-do-i-start-using-go)
+	- [Useful Resources](#useful-resources)
 
+## What Is Go?
 Go (also known as `golang`) is a compiled, statically-typed, garbage-collected language that has special memory safety and concurrent programming features. Born out of frustration with the available languages (e.g. C, C++, Java) and environments for systems programming, Go was [conceptualized by programmers at Google](https://talks.golang.org/2012/splash.article) who sought to create a single language that was efficient to write, build and execute. Go also supports newer developments in computing such as multicore processors and network systems.
 
 ## Why Learn Go?
@@ -24,7 +26,7 @@ Go (also known as `golang`) is a compiled, statically-typed, garbage-collected l
 
 Of course, Go is [not a perfect language](https://github.com/ksimka/go-is-not-good). However, for some people, Go may be exactly the right language to pick up! If you're unconvinced about how you can learn and use Go, you can use [the Go playground](https://play.golang.org/) to write, build and execute code without installing Go on your machine.
 
-## Go Features
+## What Makes Go Special?
 As it builds on the foundations set by many popular and widely-used languages such as C, C++, Java and Python, much of Go's syntax draws from existing implementations and will be familiar to programmers looking to learn an additional language. However, Go also diverges explicitly from these other languages. Listed below are some features that make Go unique!
 
 ### Declaring Variables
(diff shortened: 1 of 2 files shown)


contents/go/Go.md
```diff
@@ -0,0 +1,143 @@
...
+  executeOtherCommands()
+}
+```
+
+In the above example, using the `go` keyword launches a goroutine, which executes `run()` concurrently with `executeOtherCommands()`.
+
+Goroutines can also be started for anonymous functions:
+```go
+go func(msg string) {
+  fmt.Println(msg)
```
darrenwee: Go uses hard tabs for indentation, not spaces.

contents/go/Go.md
```diff
@@ -0,0 +1,143 @@
...
+
+More information on these three mechanisms and their uses can be found on the [Go blog](https://blog.golang.org/defer-panic-and-recover).
+
+### Interfaces
+Although Go has types and methods and allows pseudo-object-oriented style of programming, type hierarchy does not exist in Go. Instead, Go uses interfaces to specify methods that types should implement, grouping types by functionality rather than using notions of inheritance. Types implement interfaces by implementing the methods in the interface, and do not need to explicitly specify which interfaces are implemented.
+
+In the example below, the `Rectangle` type implements the interface `TwoDimensional` by implementing the methods `area()` and `perim()` that are specified in the interface.
+
+```go
+type TwoDimensional interface {
```
darrenwee: Run `go fmt` on this code snippet? Doesn't look like the canonical style to me.

contents/go/Go.md
```diff
@@ -0,0 +1,143 @@
...
+```go
+ch <- v    // Send v to channel ch.
+v := <-ch  // Receive from ch, and
+           // assign value to v.
+```
+One useful way to think about sending and receiving data with the `<-` operator is that the data moves in the direction of the arrow.
+
+If you are interested in delving deeper into using Go's concurrency features extensively, Google developers have put out video presentations on Go's [basic](https://www.youtube.com/watch?v=f6kdp27TYZs) and [advanced concurrency patterns](https://www.youtube.com/watch?v=QDDwwePbDtw). This [code walkthrough](https://golang.org/doc/codewalk/sharemem/) provides an annotated example of how Go's memory-sharing principles can be applied in practice, .
+
+### Effective Go
```
darrenwee: Rename this section to something more relevant to formatting, like `### Canonical Code Style`.

contents/go/Go.md
```diff
@@ -0,0 +1,143 @@
...
+go func(msg string) {
+  fmt.Println(msg)
+}("a message")  // starts a goroutine that prints "a message"
+```
+
+A goroutine is not its own thread; instead, goroutines are dynamically multiplexed onto threads as required to keep them running. This makes them lightweight, so having a large number of goroutines is feasible. In practice, goroutines behave similarly to very cheap threads.
+
+#### Channels
+
+Channels are used to fulfill Go's philosophy on concurrent software: "don't communicate by sharing memory; share memory by communicating".
```
darrenwee: Specify that Go relies primarily on the message passing pattern to explain the quote in software engineering terms?

contents/go/Go.md
```diff
@@ -0,0 +1,143 @@
...
+In the above example, using the `go` keyword launches a goroutine, which executes `run()` concurrently with `executeOtherCommands()`.
+
+Goroutines can also be started for anonymous functions:
+```go
+go func(msg string) {
+  fmt.Println(msg)
+}("a message")  // starts a goroutine that prints "a message"
+```
+
+A goroutine is not its own thread; instead, goroutines are dynamically multiplexed onto threads as required to keep them running. This makes them lightweight, so having a large number of goroutines is feasible. In practice, goroutines behave similarly to very cheap threads.
```
darrenwee: You can also specify that goroutines start out with a very small stack (on the order of kBs instead of an OS thread in MBs); see [here](https://stackoverflow.com/questions/41906357/why-are-goroutines-much-cheaper-than-threads-in-other-languages)

contents/go/Go.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Concurrency
+One of Go's special features is its support for concurrency. To this end, Go's standard library comes with two features that allow for easy and maintainable concurrency.
+
+#### Goroutines
+A goroutine is a lightweight thread that executes a function concurrently with its caller. A goroutine is launched by a `go` statement:
+
+```go
+func run() {
+  \\ does something
```
darrenwee: A comment in Go uses `//` instead of `\\`

contents/go/Go.md
```diff
@@ -0,0 +1,143 @@
...
+- `panic` stops the ordinary flow of execution, executes any deferred functions, and returns to its caller as a call to `panic`. The `panic` continues up the stack until all functions in the goroutine have returned, after which the program crashes.
+- `recover` regains control of a panicking goroutine. When used inside a deferred function, a call to recover captures the value returned by `panic` and resumes normal execution.
+
+More information on these three mechanisms and their uses can be found on the [Go blog](https://blog.golang.org/defer-panic-and-recover).
+
+### Interfaces
+Although Go has types and methods and allows pseudo-object-oriented style of programming, type hierarchy does not exist in Go. Instead, Go uses interfaces to specify methods that types should implement, grouping types by functionality rather than using notions of inheritance. Types implement interfaces by implementing the methods in the interface, and do not need to explicitly specify which interfaces are implemented.
```
darrenwee: Use the magic phrase `composition over inheritance` since it's a pattern.

contents/go/Go.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Arrays and slices
+Arrays are not often seen in Go code. This is because the size of a Go array must be declared at its creation, which limits how flexible an array can be.
+
+To create dynamically-sized arrays, Go introduces the concept of a slice. Slices are not arrays; rather, they are data structures that describe a piece of a separately-stored array. Slices are the topic of several Go blog entries; you can read more about [the internal layout of slices](https://blog.golang.org/go-slices-usage-and-internals) and [the mechanics of using slices](https://blog.golang.org/slices). Exercises on how to use slices are available in [A Tour of Go](https://tour.golang.org/moretypes/7).
+
+### `defer`, `panic` and `recover`
+In addition to traditional control flow mechanisms such as `if`, `for` and `switch`, Go also provides three other ways to alter the flow of a program: `defer`, `panic`, `recover`. These mechanisms can be conceived in terms of exception handling in Java and C++, but also have more general purpose uses.
```
darrenwee: I think this section needs to be a lot more precise and rescoped.

Exception handling is kind of special in Go because you rely on multiple returns, e.g. some function returns `(int, int, String, err)` and we always check `if err != nil` to perform exception handling. `panic` is only used if `err != nil` _and_ we don't know what to do. Tl;dr `panic` and `recover` only describe a subset of exception handling
https://blog.golang.org/error-handling-and-go

You could talk about `defer` here, and move `panic` and `recover` into a separate section dedicated to exception handling. I think it's worth doing so because exception handling is quite different in Go compared to other languages.

You can also include an example of `defer` since it's not a common language feature (I think).

contents/go/Go.md
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func (r Rectangle) area() {
+  return r.width * r.height
+}
+
+func (r Rectangle) perim() {
+  return r.width * 2 + r.height * 2
+}
+```
```
darrenwee: Add an example, like `type Circle struct` which implements `area()` but not `perim()` to show a negative example.

contents/go/Go.md
```diff
@@ -0,0 +1,143 @@
...
+
+Go is an [open source project](https://github.com/golang/go). Its source code may be useful reading for those interested in learning good practices, or simply to find out more about how the language was implemented.
+
+## Getting started
+
+Go provides its own [installation guide](https://golang.org/doc/install) and an interactive [ tour of Go](https://tour.golang.org/). These are useful and highly comprehensive resources for programmers looking to learn the syntax and style of Go. For those who prefer to read existing code examples, [Go by Example](https://gobyexample.com/) is a collection of code samples covering a wide variety of features in Go, and includes line-by-line explanations of the code. For those looking for a quick crash course on Go syntax, the Learn X In Y Minutes [Go cheatsheet](https://learnxinyminutes.com/docs/go/) may also be a good starting point.
+
+If you're unconvinced about Go, [the Go playground](https://play.golang.org/) can be used as a sandbox in which to write, build and execute code without installing Go on your machine.
+
+## Using Go
```
darrenwee: Add a section on variable declaration, e.g. the difference between using `var`, `:=` for inference, etc. You can also cover the reversed adjective order, e.g. in Java we specify `String message` but in Go we say `message string`. https://blog.golang.org/gos-declaration-syntax

You can also cover pointers in the same section, since they're a thing in Go but their declaration syntax is also different.

contents/go/Go.md
```diff
@@ -0,0 +1,143 @@
...
+### Effective Go
+Formatting in Go is enforced by running `gofmt`, which will align your source code with the language-wide standard style of indentation and vertical alignment. Thus, given the following code:
+
+```go
+type T struct {
+    name string // name of the object
+    value int // its value
+}
+```
+Running `$ gofmt` in the same directory as the source file will line up comments and correct the source code's indentation:
```
darrenwee: I think if you're specifying a path (like current directory) you need to use `$ go fmt` instead. Can you double-check?

contents/go/Go.md
```diff
@@ -0,0 +1,143 @@
...
+### `defer`, `panic` and `recover`
+In addition to traditional control flow mechanisms such as `if`, `for` and `switch`, Go also provides three other ways to alter the flow of a program: `defer`, `panic`, `recover`. These mechanisms can be conceived in terms of exception handling in Java and C++, but also have more general purpose uses.
+
+- `defer` pushes a function call to a list, and executes all functions on the list in last-in-first-out order after the surrounding function returns. It is frequently used for clean-up actions, such as to close files.
+- `panic` stops the ordinary flow of execution, executes any deferred functions, and returns to its caller as a call to `panic`. The `panic` continues up the stack until all functions in the goroutine have returned, after which the program crashes.
+- `recover` regains control of a panicking goroutine. When used inside a deferred function, a call to recover captures the value returned by `panic` and resumes normal execution.
+
+More information on these three mechanisms and their uses can be found on the [Go blog](https://blog.golang.org/defer-panic-and-recover).
+
+### Interfaces
```
darrenwee: Since we're touching the OOP parts of Go, you can also add a section on how encapsulation works. https://code.tutsplus.com/tutorials/lets-go-object-oriented-programming-in-golang--cms-26540

contents/go/Go.md
```diff
@@ -0,0 +1,143 @@
...
+Goroutines send and receive messages through a channel using the `<-` operator.
+
+```go
+ch <- v    // Send v to channel ch.
+v := <-ch  // Receive from ch, and
+           // assign value to v.
+```
+One useful way to think about sending and receiving data with the `<-` operator is that the data moves in the direction of the arrow.
+
+If you are interested in delving deeper into using Go's concurrency features extensively, Google developers have put out video presentations on Go's [basic](https://www.youtube.com/watch?v=f6kdp27TYZs) and [advanced concurrency patterns](https://www.youtube.com/watch?v=QDDwwePbDtw). This [code walkthrough](https://golang.org/doc/codewalk/sharemem/) provides an annotated example of how Go's memory-sharing principles can be applied in practice, .
```
darrenwee: Include a minimum working example incorporating both goroutines and channels, e.g. asynchronously sum halves of an array.

contents/go/Go.md
```diff
@@ -0,0 +1,143 @@
...
+
+If you're unconvinced about Go, [the Go playground](https://play.golang.org/) can be used as a sandbox in which to write, build and execute code without installing Go on your machine.
+
+## Using Go
+As it builds on the foundations set by many popular and widely-used languages such as C, C++, Java and Python, much of Go's syntax draws from existing implementations and will be familiar to programmers looking to learn an additional language. However, Go also diverges explicitly from these other languages; listed below are some aspects of Go that may be unfamiliar to learners.
+
+### Arrays and slices
```
darrenwee: `Slices`

darrenwee (changes requested):
Looks pretty good. You'll need to run `go fmt` on all the code snippets, and I've included some comments on possible things you may want to cover because I think they're quite fundamental differences between Go and most other mainstream languages.

contents/go/Go.md
craaaa: Hm, I'm not sure I want to explicitly mention this, as the section isn't about OOP in general. There's quite a lot to be said about Go's approach to OOP, but I don't think that I want to cover that here as it's somewhat prickly. Moreover, I intended for this section to walk through potential painpoints for a new learner of Go to look out for, and I think Go's encapsulation method is pretty straightforward. Will add it to the list of links though!

craaaa:
Hey @darrenwee, thanks for your comments! I've made most of the changes you suggested; do check out the section and let me know what you think!

contents/go/Go.md
```diff
@@ -0,0 +1,248 @@
...
+//alternatively
+c := make(chan int)
+```
+
+Channels can transmit data of any type; thus, creating a channel that transmits channels (i.e. a `chan chan`) is theoretically possible and may even be [useful](http://tleyden.github.io/blog/2013/11/23/understanding-chan-chans-in-go/).
+
+Goroutines send and receive messages through a channel using the `<-` operator.
+
+```go
+	ch <- v   // Send v to channel ch.
```
darrenwee: Remove indent

darrenwee (approved):
Just the one comment, then good to Go :+1: :100:

darrenwee (approved):
LGTM

craaaa:
@darrenwee thanks for your help!
@damithc ready for review!

contents/go/Go.md
```diff
@@ -0,0 +1,248 @@
...
+
+Go is an [open source project](https://github.com/golang/go). Its source code may be useful reading for those interested in learning good practices, or simply to find out more about how the language was implemented.
+
+## Getting started
+
+Go provides its own [installation guide](https://golang.org/doc/install) and an interactive [ tour of Go](https://tour.golang.org/). These are useful and highly comprehensive resources for programmers looking to learn the syntax and style of Go. For those who prefer to read existing code examples, [Go by Example](https://gobyexample.com/) is a collection of code samples covering a wide variety of features in Go, and includes line-by-line explanations of the code. For those looking for a quick crash course on Go syntax, the Learn X In Y Minutes [Go cheatsheet](https://learnxinyminutes.com/docs/go/) may also be a good starting point.
+
+If you're unconvinced about Go, [the Go playground](https://play.golang.org/) can be used as a sandbox in which to write, build and execute code without installing Go on your machine.
```
tran-tien-dat: Shouldn't it be either `a sandbox to write` or `a sandbox in which you can write`?

contents/go/Go.md
```diff
@@ -0,0 +1,248 @@
...
+```go
+var name string = "John Smith"
+
+// alternatively
+var name = "John Smith"
+```
+
+When initializers are not used, simply declaring a variable allocates storage for the variable and initializes its value to that type's zero value. The zero value is `0` for numeric types, `false` for booleans, `""` for strings, `nil` for pointers, functions, interfaces, slices, channels, and maps.
+
+As an extension of type inference, the `:=` short assignment statement can be used inside a function to as a `var` declaration with implicit type:
```
tran-tien-dat: I think this sentence is not very clear for first-timer. I need to re-read it and look at the example to figure out what you mean. I think the problem is because you are using the passive voice which makes my brain wonders who is doing what. I suggest something like `In addition to omitting the type when it can be inferred, the programmer can also eliminate the keyword var in  the variable declaration by using the := short assignment statement`.

contents/go/Go.md
```diff
@@ -0,0 +1,248 @@
...
+```go
+name := "John Smith"
+```
+
+Thus, we see that there are several potential ways to declare a variable in Go. Choosing which style of variable declaration to use depends on how verbose a programmer wants to be in declaring the variable.
+
+Go's syntax for more complex types such as pointers, arrays and structs is also somewhat idiosyncratic, and can be explored in this [Go blog article](https://blog.golang.org/gos-declaration-syntax), or with the help of [this tutorial on pointers](http://www.golang-book.com/books/intro/8) and [this tutorial on structs](https://www.golang-book.com/books/intro/9).
+
+### Arrays and Slices
+Arrays are not often seen in Go code. This is because the size of a Go array must be declared at its creation, which limits how flexible an array can be. For instance, in the
```
tran-tien-dat: Unfinished  sentence?

contents/go/Go.md
```diff
@@ -0,0 +1,248 @@
...
+
+```go
+func foo() {
+	defer fmt.Println("This gets printed third")
+	defer fmt.Println("This gets printed second")
+	fmt.Println("This gets printed first")
+}
+```
+
+`defer` is frequently used for clean-up actions, such as to [close files](https://gobyexample.com/defer). Deferred functions run on panicking goroutines as well, which makes them useful for recovering from `panic`.
```
tran-tien-dat: So what is `panic`? Maybe mention that it will be explained later.

contents/go/Go.md
```diff
@@ -0,0 +1,248 @@
...
+
+An `error` variable represents any value that can describe itself as a string, by implementing the following interface:
+
+```go
+type error interface {
+	Error() string
+}
+```
+
+When calling a method that may return an error, we check if the returns `err != null` and handle the resulting error.
```
tran-tien-dat: What's the difference between `null` and `nil`? You mention `null` here but `nil` is used in the code.

contents/go/Go.md
```diff
@@ -0,0 +1,248 @@
...
+	    log.Fatal(err)
+	    return
+	}
+	// do something with the open *File f
+}
+```
+
+To deal with unexpected errors, Go also provides two mechanisms: `panic` and `recover`.
+
+- `panic` stops the ordinary flow of execution, executes any deferred functions, and returns to its caller as a call to `panic`. The `panic` continues up the stack until all functions in the goroutine have returned, after which the program crashes. `panic` is used to fail fast on errors that cannot be handled gracefully.
```
tran-tien-dat: This paragraph is really hard to digest. Had to re-read a few times to realize that you wants to describe a recursive mechanism. Maybe you can add some more explanation to make it easier to understand. Best is to have an example to illustrate.

Also, up till now I have assumed that `function` is the same as `goroutine` but from this sentence I suppose there is a difference. Ok I see that it is explained in Concurrency section. I would prefer that you give a definition, which could be brief, of every term the first time you use it though.

contents/go/Go.md
```diff
@@ -0,0 +1,248 @@
...
+- `recover` regains control of a panicking goroutine. When used inside a deferred function, a call to recover captures the value returned by `panic` and resumes normal execution.
```
tran-tien-dat: Just curious. Where does execution returns? The next line after the first `panic` that trigger all the panicking, or is it somewhere else?

contents/go/Go.md
```diff
@@ -0,0 +1,248 @@
...
+
+More information on error handling can be found on the [Go blog](https://blog.golang.org/error-handling-and-go).
+
+### Interfaces
+Although Go has types and methods and allows pseudo-object-oriented style of programming, type hierarchy does not exist in Go. Instead, Go uses interfaces to specify methods that types should implement, favouring composition over inheritance. Types implement interfaces by implementing the methods in the interface, and do not need to explicitly specify which interfaces are implemented.
```
tran-tien-dat: I would change `and` to `but`, because it is quite surprising for a person used to OOP paradigm to see this.

contents/go/Go.md
```diff
@@ -0,0 +1,248 @@
...
+
+In the example below, the `Rectangle` type implements the interface `TwoDimensional` by implementing the methods `area()` and `perim()` that are specified in the interface. Thus, instances of `Rectangle` can be used as arguments to `price`. 
+
+Meanwhile, although `Circle` implements `perim()`, it does not implement `area()`. Since it does not implement all the methods in the `TwoDimensional` interface, `Circle` does not implement `TwoDimensional`. Thus, instances of `Circle` cannot be used as arguments to `price`.
```
tran-tien-dat: Maybe add code below where instances of `Rectangle` and `Circle` are used and for `Circle` the compiler gives an error.

tran-tien-dat (commented):
Hi, I wanted to learn Go so I read your book chapter and left some comments. Hope that such comments from somebody without Go experience will be useful. Overall, the book chapter is very well-written. It's not just a list of syntactic constructs but focusing on highlighting interesting features of Go. Makes me wanna try it out 👍

damithc (commented):
I'm ok with the overall structure. Request a review when you have address Dat's comments.

craaaa:
Hi @tran-tien-dat, thanks for getting some fresh eyes on the page, and for giving such useful comments! I've made some changes to the document to make new terms more explicit (by moving sections around), and to clarify ambiguities that you pointed out.

@damithc ready for review!

tran-tien-dat (approved):
LGTM 💯

contents/go/Go.md
```diff
@@ -0,0 +1,261 @@
...
+In addition, the verbosity of a declaration statement in Go can vary. You do not need to declare the `variableType` of a new variable if you use an initializer, as the type of the variable will be inferred from its initialization. For instance, in the following example, the `string` type is optional:
+
+```go
+var name string = "John Smith"
+
+// alternatively
+var name = "John Smith"
+```
+
+When an initializer is not used, declaring a variable allocates storage for the variable and initializes its value to that type's zero value. The zero value is `0` for numeric types, `false` for booleans, `""` for strings, `nil` for pointers, functions, interfaces, slices, channels, and maps.
```
damithc: Can do without this sentence? Feels like its something the reader may not be interested at this stage.

contents/go/Go.md
```diff
@@ -0,0 +1,261 @@
...
+
+Go's syntax for more complex types such as pointers, arrays and structs is also somewhat idiosyncratic, and can be explored in this [Go blog article](https://blog.golang.org/gos-declaration-syntax), or with the help of [this tutorial on pointers](http://www.golang-book.com/books/intro/8) and [this tutorial on structs](https://www.golang-book.com/books/intro/9).
+
+### Arrays and Slices
+Arrays are not often seen in Go code. This is because the size of a Go array must be declared at its creation, which limits how flexible an array can be. For instance, in the 
+
+To create dynamically-sized arrays, Go introduces the concept of a slice. Slices are not arrays; rather, they are data structures that describe a piece of a separately-stored array. Slices are the topic of several Go blog entries; you can read more about [the internal layout of slices](https://blog.golang.org/go-slices-usage-and-internals) and [the mechanics of using slices](https://blog.golang.org/slices). Exercises on how to use slices are available in [A Tour of Go](https://tour.golang.org/moretypes/7).
+
```
damithc: Give a concrete example of slices?

contents/go/Go.md
```diff
@@ -0,0 +1,261 @@
+# Introduction to Go
+
+Author(s): [Cara Leong](https://github.com/craaaa)
```
damithc, line 3: Can add a table of contents? a ToC makes it easier to see the high-level structure.

contents/go/Go.md
```diff
@@ -0,0 +1,261 @@
...
+
+Go (also known as `golang`) is a compiled, statically-typed, garbage-collected language that has special memory safety and concurrent programming features. Born out of frustration with the available languages (e.g. C, C++, Java) and environments for systems programming, Go was [conceptualized by programmers at Google](https://talks.golang.org/2012/splash.article) who sought to create a single language that was efficient to write, build and execute. Go also supports newer developments in computing such as multicore processors and network systems.
+
+Go is an [open source project](https://github.com/golang/go). Its source code may be useful reading for those interested in learning good practices, or simply to find out more about how the language was implemented.
+
+## Getting started
```
damithc: Seems like this can be moved to the end of the chapter because at this point the reader has not yet decided whether to get started with Go.

contents/go/Go.md
```diff
@@ -0,0 +1,261 @@
...
+
+Go provides its own [installation guide](https://golang.org/doc/install) and an interactive [ tour of Go](https://tour.golang.org/). These are useful and highly comprehensive resources for programmers looking to learn the syntax and style of Go. For those who prefer to read existing code examples, [Go by Example](https://gobyexample.com/) is a collection of code samples covering a wide variety of features in Go, and includes line-by-line explanations of the code. For those looking for a quick crash course on Go syntax, the Learn X In Y Minutes [Go cheatsheet](https://learnxinyminutes.com/docs/go/) may also be a good starting point.
+
+If you're unconvinced about Go, you can use [the Go playground](https://play.golang.org/) to write, build and execute code without installing Go on your machine.
+
+## Using Go
```
damithc: Another way to frame this section is as `Go Features`. Then describe things that are somewhat unique to Go, but also mention why each feature matters i.e., how does it make Go better?

damithc (changes requested):
Looks good. Added a few comments.

craaaa:
@damithc Ready for review - made changes as suggested

contents/go/Go.md
```diff
@@ -0,0 +1,273 @@
...
+	- [Error Handling](#error-handling)
+	- [Interfaces](#interfaces)
+	- [Canonical Coding Style](#canonical-coding-style)
+- [Getting started](#getting-started)
+- [Useful Resources](#useful-resources)
+
+Go (also known as `golang`) is a compiled, statically-typed, garbage-collected language that has special memory safety and concurrent programming features. Born out of frustration with the available languages (e.g. C, C++, Java) and environments for systems programming, Go was [conceptualized by programmers at Google](https://talks.golang.org/2012/splash.article) who sought to create a single language that was efficient to write, build and execute. Go also supports newer developments in computing such as multicore processors and network systems.
+
+## Why Learn Go?
+Go is [not a perfect language](https://github.com/ksimka/go-is-not-good). However, here are some reasons why you might want to learn and use Go:
```
damithc: Put this at the end of this section rather than at the beginning?
>Go is [not a perfect language](https://github.com/ksimka/go-is-not-good).

contents/go/Go.md
```diff
@@ -0,0 +1,273 @@
...
+
+Firstly, Go is a language built for software engineers. As it was written by software engineers at Google, Go addresses and attempts to solve some of the  pain points that exist in many commonly-used languages. For instance, the language has easy dependency management, prides itself on fast builds and has many [easy to use debugging, testing, and code-vetting tools](https://medium.com/google-cloud/go-tooling-in-action-eca6882ff3bc). These features make Go an easy language to use for software development.
```
damithc: Alternative formatting:
>1. **Go is a language built for software engineers.** As it was written by software engineers at Google, Go addresses and attempts to solve some of the  pain points that exist in many commonly-used languages. For instance, the language has easy dependency management, prides itself on fast builds and has many [easy to use debugging, testing, and code-vetting tools](https://medium.com/google-cloud/go-tooling-in-action-eca6882ff3bc). These features make Go an easy language to use for software development.
>1. ...

contents/go/Go.md
```diff
@@ -0,0 +1,273 @@
...
+
+Go's syntax for more complex types such as pointers, arrays and structs is also somewhat idiosyncratic, and can be explored in this [Go blog article](https://blog.golang.org/gos-declaration-syntax), or with the help of [this tutorial on pointers](http://www.golang-book.com/books/intro/8) and [this tutorial on structs](https://www.golang-book.com/books/intro/9).
+
+### Concurrency
+One of Go's special features is its support for concurrency. To this end, Go's standard library comes with two features that allow for easy and maintainable concurrency.
```
damithc: Other language support concurrency too. Go has _better_ support?

contents/go/Go.md
```diff
@@ -0,0 +1,273 @@
...
+
+```go
+func foo() {
+	defer fmt.Println("This gets printed third")
+	defer fmt.Println("This gets printed second")
+	fmt.Println("This gets printed first")
+}
+```
+
+`defer` is frequently used for clean-up actions, such as to [close files](https://gobyexample.com/defer). Deferred functions run on [panicking goroutines](#error-handling) as well, which makes them useful for recovering from `panic`.
```
damithc: The mention of _panicking goroutines_  hurts the self-contained quality of this article because the reader has to go to another page before continuing to read this article. Give a quick definition (or avoid altogether)?

damithc (approved):
Looks good. 👍  A few cosmetic nits. I'll leave you to decide whether to follow the suggestions. Let me know when ready to merge.

contents/go/Go.md
craaaa: Decided to move this section further down in the list of Go features. The hope is that users unfamiliar with Go will have read the previous section on `panic`, which will teach them about what a panicking goroutine is, before they reach this section.

PR #54, opened 2018-04-03
Java: Add a section on Java Concurrency

vivekscl:
View the rendered version [here](https://github.com/se-edu/learningresources/blob/ff47f090d2b69cc3beeebbf811fb0fcc70e9e4de/contents/java/JavaConcurrency.md)

@darrenwee maybe could take a quick look :smile:


contents/java/JavaConcurrency.md
```diff
@@ -0,0 +1,149 @@
+# Java Concurrency
+
+Authors: Vivek Lakshmanan
+
+* [What is Java Concurrency](#what-isjava-concurrency)
```
darrenwee: Broken reference due to missing hyphen, should be `#what-is-java-concurrency`

contents/java/JavaConcurrency.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Creating and Starting Threads
+There are two ways to create a thread in Java:
+ 
+1. Extend the `Thread` class
+1. Implement the `Runnable` interface
+
+After which, override the `run()` method which contains the code that will be executed when the thread starts. 
+
+Depending on how you created the `Thread` you can either create the class that extends the `Thread` class or pass the class that implements the `Runnable` interface into the Thread constructor and then start it.
```
darrenwee: You missed monospace for the last `Thread`.

contents/java/JavaConcurrency.md
```diff
@@ -0,0 +1,149 @@
...
+
+When deciding which method to use to create a `Thread`, it is always advisable to implement the `Runnable` interface as this results in composition which will allow your code to be more loosely coupled as compared to inheritance. Furthermore, you can extend another class if need be.
```
Zhiyuan-Amos: You should write code here to show the difference in implementing `Runnable` interface vs not implementing it.

contents/java/JavaConcurrency.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Operations on Threads
+After creating and starting threads, you can carry out operations on them. There are several such operations you can use to manipulate threads:
+
+* Making a thread sleep for a given amount of time. This is useful for pausing the execution of a certain thread for a certain amount of time to free up resources for other threads. The sleeping thread can be interrupted by another thread, which can be a way of communicating that the other thread is done with it's work allowing the current thread to start.
```
Zhiyuan-Amos: Just wondering, what's the difference between using a lock for the threads vs handling `InterruptedException`? It seems that handling `InterruptedException` isn't a good way of "communicating that the other thread is done with it's work allowing the current thread to start" because exceptions ought to be exceptional; it shouldn't be part of control flow.

contents/java/JavaConcurrency.md
```diff
@@ -0,0 +1,149 @@
...
+```java
+try {
+    thread2.join();
+} catch (InterruptedException e) {
+    // handle interruption
+}
+```
+
+### Thread Interference
+Since threads share the resources of the process they exists in, there will inevitably be conflicts when using shared resources due to the unpredictable nature of threads. When threads try to write to the same resource, thread interference occurs. To illustrate this problem, here's a sequence of execution for two threads, A and B, that increment and decrement a counter variable respectively:
```
Zhiyuan-Amos: exist~~s~~

contents/java/JavaConcurrency.md
```diff
@@ -0,0 +1,149 @@
...
+
+1. Thread A retrieves the value of counter and increments it to a value of 1
+2. Thread B then retrieves the values of counter, which is still 0, and decrements it to a value of -1
+3. Thread A stores the value of 1 to the counter variable
+4. Thread B stores the value of -1 to the counter variable and overwrites the previous value stored by Thread A
```
Zhiyuan-Amos, line 97: I think you can use images to illustrate this, which will make it easier to understand.

contents/java/JavaConcurrency.md
```diff
@@ -0,0 +1,149 @@
...
+
+Without thread interference, the expected value of the counter variable would be 0, since one thread increments it while the other decrements it. But with thread interference, the value of counter is simply the value written by the last thread. This is due to the unpredictable nature of threads as there is no way to know when the operating system switches between the two threads. 
+
+To solve this issue of interference, the keyword `synchronized` is used to ensure that method or block of code can only be accessed by one thread at a time. This is done through the use of the intrinsic lock system put in place by Java. Simply put, each object has it's own intrinsic lock which:
```
Zhiyuan-Amos: the term `lock` needs to be explained.

contents/java/JavaConcurrency.md
```diff
@@ -0,0 +1,149 @@
...
+
+![Synchronized block](http://www.javalearningacademy.com/wp-content/uploads/2015/10/Synchronized_Code_Block_And_Thread_Execution.jpg)
+
+For a deeper look, see the [Java Synchronisation](java/JavaSynchronization.md) section. 
+
+### Executors
+While it is easy to create one or two threads and run them, it becomes a problem when your application requires creating 20 or 30 threads for running tasks concurrently. This is where the `Executors` class comes in. It helps you with:
+
+1. **Thread Creation**<br>
+It provides various methods for creating threads, more specifically a pool of threads, uses to run tasks concurrently.
```
Zhiyuan-Amos: > more specifically a pool of threads, uses to run tasks concurrently.

The word "uses" seem out of place.

contents/java/JavaConcurrency.md
```diff
@@ -0,0 +1,149 @@
...
+
+1. **Thread Management**<br>
+It manages the life cycle of the threads in the thread pool. You don’t need to worry about whether the threads in the thread pool are active or busy or dead before submitting a task for execution.
+
+1. **Task submission and execution**<br>
+It provides methods for submitting tasks for execution in the thread pool, and also allows you to decide when the tasks will be executed. For example, a task can be submitted to be executed instantly, scheduled to be executed later or even executed periodically. 
+
+The `Executors` class provides convenient factory methods for creating the `ExecutorService` class. This class manages the lifecycle of tasks in various ways by assigning a `ThreadPool` to it. Different thread pools manage your tasks in different ways along with their own advantages and disadvantages. Some of these include:
```
Zhiyuan-Amos: Is there a diagram to show how `Executors`, `ExecutorService` and `ThreadPool` are linked? :P

contents/java/JavaConcurrency.md
```diff
@@ -0,0 +1,149 @@
...
+* `FixedThreadPool` -  A thread pool with a fixed number of threads. If all threads are active when a new task is submitted, they will wait in queue until a thread becomes available. This can come in handy when you know the exactly how many threads you need, though that may be tricky by itself.
+
+* `ThreadPoolExecutor` - This thread pool implementation adds the ability to configure parameters such as the maximum number of threads it can hold and how long to keep extra threads alive.
+ 
+A simple example of using the `Executors` class is shown below where after passing in the task to be executed, it is automatically managed by the `ExecutorService`. For a more detailed look at the workflow of the `ExecutorService`, see this [in-depth tutorial](https://www.callicoder.com/java-executor-service-and-thread-pool-tutorial/).  
+```java
+ExecutorService executor = Executors.newFixedThreadPool(10); 
+executor.execute(() -> {  
+    System.out.println("Hello " + Thread.currentThread().getName());
+});
```
Zhiyuan-Amos: You can show the output of this code

contents/java/JavaConcurrency.md
```diff
@@ -0,0 +1,149 @@
...
+ 
+1. Extend the `Thread` class
+1. Implement the `Runnable` interface
+
+After which, override the `run()` method which contains the code that will be executed when the thread starts. 
+
+Depending on how you created the `Thread` you can either create the class that extends the `Thread` class or pass the class that implements the `Runnable` interface into the Thread constructor and then start it.
+
+When deciding which method to use to create a `Thread`, it is always advisable to implement the `Runnable` interface as this results in composition which will allow your code to be more loosely coupled as compared to inheritance. Furthermore, you can extend another class if need be.
+
```
LuLechuan: Maybe you can provide a code snippet or links to some examples of implementing threads using the `Thread` class and `Runnable` respectively.

contents/java/JavaConcurrency.md
```diff
@@ -0,0 +1,149 @@
...
+As the saying goes, _there is no free lunch_. While concurrency provides great benefits as mentioned above, it does come with several issues such as:
+
+* **A more complex design**<br>
+Due to the creation of multiple threads and the problem of thread interference, the design becomes more complex as much more work is done to manage these threads such that no problem occurs when the application runs.  
+
+* **Harder debugging and testing process**<br>
+The unpredictable nature of threads result in errors that can be hard to detect, reproduce and fix as these errors don't crop up consistently like normal errors do. The execution order is indeterminate and it's hard to see the flow of the program.
+
+* **Context switching overhead**<br>
+When a CPU switches from executing one thread to executing another, the CPU needs to save the local data, program pointer etc. of the current thread, and load the local data, program pointer etc. of the next thread to execute, making this process of context switching very expensive.
```
Zhiyuan-Amos: There's too much specifics here e.g. `local data, program pointer`. Since this resource is meant to be read by newbies, we should keep it simple by finding a general term that encompasses all these things.

contents/java/JavaConcurrency.md
```diff
@@ -0,0 +1,149 @@
...
+    System.out.println("Hello " + Thread.currentThread().getName());
+});
+
+```
+
+## Disadvantages of Java Concurrency
+As the saying goes, _there is no free lunch_. While concurrency provides great benefits as mentioned above, it does come with several issues such as:
+
+* **A more complex design**<br>
+Due to the creation of multiple threads and the problem of thread interference, the design becomes more complex as much more work is done to manage these threads such that no problem occurs when the application runs.
```
Zhiyuan-Amos: You can use a concrete example to show how code looks more complex when using concurrency.

contents/java/JavaConcurrency.md
```diff
@@ -0,0 +1,149 @@
...
+```
+
+## Disadvantages of Java Concurrency
+As the saying goes, _there is no free lunch_. While concurrency provides great benefits as mentioned above, it does come with several issues such as:
+
+* **A more complex design**<br>
+Due to the creation of multiple threads and the problem of thread interference, the design becomes more complex as much more work is done to manage these threads such that no problem occurs when the application runs.  
+
+* **Harder debugging and testing process**<br>
+The unpredictable nature of threads result in errors that can be hard to detect, reproduce and fix as these errors don't crop up consistently like normal errors do. The execution order is indeterminate and it's hard to see the flow of the program.
```
Zhiyuan-Amos: The first sentence is quite similar to the 2nd sentence; the reason why threads have "unpredictable nature" is because the "execution order is indeterminate".

LuLechuan (commented):
Hi @vivekscl, good work in explaining concurrency. This document is very clear and easy to follow even for someone never take CS2106 like me. However, perhaps it will be better if you can provide some code snippets to show how to implement it using the `Thread` class and  `Runnable` interface respectively.

contents/java/JavaConcurrency.md
LuLechuan: Agreed

contents/java/JavaConcurrency.md
```diff
@@ -0,0 +1,149 @@
...
+The following resources are the many in-depth tutorials that will help you get a better grasp of concurrency in Java.
+
+* [Detailed seven part series on concurrency in Java](https://baptiste-wicht.com/posts/2010/05/java-concurrency-part-1-threads.html)
+* [High-level overview of Concurrency in Java by Vogella](http://www.vogella.com/tutorials/JavaConcurrency/article.html#concurrency)
+* [Concurrency best practices](https://www.javacodegeeks.com/2015/09/concurrency-best-practices.html)
+* [Java 101: Understanding Java threads](https://www.javaworld.com/article/2074217/java-concurrency/java-101--understanding-java-threads--part-1--introducing-threads-and-runnables.html)
+* [Java Concurrency / Multithreading Tutorial by Jenkov](http://tutorials.jenkov.com/java-concurrency/index.html)
+* [Java Concurrency Tutorial by HowToDoInJava](https://howtodoinjava.com/java-concurrency-tutorial/)
+* [Java Concurrency Tutorial by tutorialspoint](https://www.tutorialspoint.com/java_concurrency/concurrency_overview.htm)
+* [Easy to understand and quick Java 8 Concurrency Tutorial](http://winterbe.com/posts/2015/04/07/java8-concurrency-tutorial-thread-executor-examples/)
```
darrenwee: There are four concurrency tutorials listed here. Is anyone of them better than the other? I would remove like at least two of them and just link to a useful one(s) instead of linking to redundant material.

contents/java/JavaConcurrency.md
```diff
@@ -0,0 +1,149 @@
...
+    * [Creating and Starting Threads](#creating-and-starting-threads)
+    * [Operations on Threads](#operations-on-threads)
+    * [Thread Interference](#thread-interference)
+    * [Executors](#executors)
+* [Disadvantages of Java Concurrency](#disadvantages-of-java-concurrency)
+* [Resources](#resources)
+    * [Further Reading](#further-reading)
+
+## What is Java Concurrency
+_Concurrency_ is the ability to run several programs or several parts of a program **out-of-order**, in an **interleaved fashion**. Simply put, if a program is running concurrently, the program execution will execute one part of the program, pause it, execute another part and repeat. This is different from _Parallelism_ in which parts of the program are executed at the **same time**.
```
darrenwee: This paragraph is technically correct, but I think you should disambiguate concurrency from parallelism as separate properties. You can be concurrent and not parallel, not concurrent and parallel, neither concurrent nor parallel or both concurrent and parallel. (see [this](https://stackoverflow.com/a/36604522))

contents/java/JavaConcurrency.md
```diff
@@ -0,0 +1,149 @@
...
+
+Java Concurrency enables you to use perform tasks using multi-threading in your code and therefore:
```
darrenwee: Related to the above comment, since you use the term `multi-threading` here and that concurrency can `improve performance` and `take advantage of multi-core processors`. This would imply using both concurrency _and_ parallelism at the same time. I feel that the two concepts are not well separated in this part on what concurrency allows you to do.

I think that you should explain the distinction between concurrency and parallelism at the start, but then talk about concurrency in the rest of the chapter like as if concurrency implies parallelism. What do you think?

darrenwee (changes requested):
The others have also highlighted some of the same concerns that I have!

contents/java/JavaConcurrency.md
vivekscl: >I think that you should explain the distinction between concurrency and parallelism at the start

This sounds good! As you mentioned above, I haven't separated them properly, so having a clear distinction would improve clarity. 

>but then talk about concurrency in the rest of the chapter like as if concurrency implies parallelism. What do you think?

Since this chapter is about concurrency, I think it's better to replace `take advantage of multi-core processors` with `better resource utilization`. So now, `using both concurrency and parallelism at the same time` is not implied which is more appropriate since that's not what concurrency in Java allows you to do.

On a side note, the term `multi-threading` still falls under concurrency as long as we clearly state that it is done by a single processor right?

darrenwee: Ah yes, that is technically correct re. `multi-threading`. I am also guilty of conflating it with `concurrency + parallelism` :sweat_smile: 

> Since this chapter is about concurrency, I think it's better to replace `take advantage of multi-core processors` with `better resource utilization`. So now, `using both concurrency and parallelism at the same time` is not implied which is more appropriate since that's not what concurrency in Java allows you to do.

Yeah, I think that's okay. It's stuff like not letting the core go idle for stuff like waiting on I/O.

contents/java/JavaConcurrency.md
vivekscl: But then again, wouldn't having complex code be hard to understand for newbies?

contents/java/JavaConcurrency.md
vivekscl: Ahh, my bad :sweat_smile: . A lock should be a better way of `communicating that the other thread is done with it's work allowing the current thread to start`. The handling of the `InterruptedException` should only done in exceptional cases and not for communication. `Thread.sleep` is instead **an efficient means of making processor time available to the other threads and freeing up resources**.

vivekscl:
Thanks @LuLechuan, @Zhiyuan-Amos and @darrenwee for the thorough reviews! Really appreciate it! :smile: 
I've made the necessary changes :+1: 
[Rendered version](https://github.com/vivekscl/learningresources/blob/5404284a0ecea5cf2c5d670c6166da881e47eaf8/contents/java/JavaConcurrency.md)

contents/java/JavaConcurrency.md
Zhiyuan-Amos: If you put comments to demarcate each segment, then it wouldn't be :P

contents/java/JavaConcurrency.md
```diff
@@ -0,0 +1,177 @@
...
+    * [Thread Interference](#thread-interference)
+    * [Executors](#executors)
+* [Disadvantages of Java Concurrency](#disadvantages-of-java-concurrency)
+* [Resources](#resources)
+    * [Further Reading](#further-reading)
+
+## What is Java Concurrency
+_Concurrency_ is the ability to run several programs or several parts of a program **out-of-order**, in an **interleaved fashion**. Simply put, if a program is running concurrently, the processor will execute one part of the program, pause it, execute another part and repeat. 
+
+As such , Java Concurrency enables you to perform tasks using multi-threading in your code and therefore:
```
Zhiyuan-Amos: Extra space after `As such`.

contents/java/JavaConcurrency.md
```diff
@@ -0,0 +1,177 @@
...
+1. improving performance
```
Zhiyuan-Amos: Not necessary if you only have 1 processor since each part of the program has a lesser execution time.

contents/java/JavaConcurrency.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Understanding Threads
+First off, a process is simply a program in execution. It contains at least one thread and has it's own memory space.
+
+A thread: 
+* is a lightweight process that exists within a process 
+* is an independent path of execution through program code 
+* has its own call stack and can access shared data of other threads in the same process. 
+
+When multiple threads execute code concurrently, it gives rise to multi-threading.
```
Zhiyuan-Amos: This is parallelism right? :P Since the focus is on concurrency and not parallelism, then you should either remove this statement, or you should explicitly mention that this is parallelism.

contents/java/JavaConcurrency.md
```diff
@@ -0,0 +1,177 @@
...
+* **Harder debugging and testing process**<br>
+The unpredictable nature of threads result in errors that can be hard to detect, reproduce and fix as these errors don't crop up consistently like normal errors do.
+
+* **Context switching overhead**<br>
+When a CPU switches from executing one thread to executing another, the CPU needs to save the state of the current thread, and load the state of the next thread to execute, making this process of context switching very expensive.
+
+## Resources
+The following resources are the many in-depth tutorials that will help you get a better grasp of concurrency in Java.
+
+* [Detailed seven part series on concurrency in Java](https://baptiste-wicht.com/posts/2010/05/java-concurrency-part-1-threads.html)
```
Zhiyuan-Amos: As mentioned by 1 of the previous reviewers, you may want to point out what's the unique point of each article so that readers will know which articles they want to read out of the many articles that you are providing.

Zhiyuan-Amos (commented):
In the `ExecutorService` image, why are there 3 arrows pointing to the thread pool when there's 4 threads inside :P

I'll pass on the next review; busy :P

vivekscl:
>In the ExecutorService image, why are there 3 arrows pointing to the thread pool when there's 4 threads inside

I believe it means the 3 tasks are sent randomly to one these threads? So each arrow represents a task? But the image describes the concept well so I guess it's ok to keep it? :smile:

Zhiyuan-Amos:
> I believe it means the 3 tasks are sent randomly to one these threads? So each arrow represents a task? But the image describes the concept well so I guess it's ok to keep it? 😄

Oh I see, ok! 👍

darrenwee (approved):

damithc (approved):
LGTM 👍

PR #56, opened 2018-04-04
Add Chapter: Introduction to Functional Programming

crphang:
A high level introduction to Functional Programming.

### DIFF
--- a/contents/functionalProgramming/intro.md
+++ b/contents/functionalProgramming/intro.md
@@ -8,7 +8,7 @@ Functional programming is a programming paradigm that treats computation as the 
 
 ## Functional Programming Languages
 
-Functional programming is simply a paradigm and needs to be implemented by programming languages. Below are various [languages](https://en.wikipedia.org/wiki/List_of_programming_languages_by_type#Functional_languages) that have explicit support for functional programming paradigm such as:
+Functional programming is simply a paradigm and needs to be implemented by programming languages. Below are various [languages](https://en.wikipedia.org/wiki/List_of_programming_languages_by_type#Functional_languages) that have explicit support for the functional programming paradigm such as:
 
 - Haskell
 - Clojure
@@ -17,7 +17,7 @@ Functional programming is simply a paradigm and needs to be implemented by progr
 - Scala
 - ML (Meta Language) family of languages
 
-While functional programming can be implemented by languages like Java, the languages listed above encourage functional programming paradigm such as [pure functions](#pure-functions) or even enforce them in the case of Haskell.
+While functional programming can be implemented by languages like Java, the languages listed above encourage the functional programming paradigm such as [pure functions](#pure-functions) or even enforce them in the case of Haskell.
 
 ## Concepts in Functional Programming
 
@@ -64,7 +64,7 @@ To know more about immutability in functional languages, you can take a look at:
 
 #### Recursion
 
-Pure functional languages does not have loop constructs that procedural languages does. This is because `Iteration` usually involves state mutation per iteration. Since functional programming avoids state changes, `Recursion` is a commonly used [technique](https://www.quora.com/Why-dont-pure-functional-programming-languages-provide-a-loop-construct) to replace `Iteration`.
+Pure functional languages do not have loop constructs like imperative languages. This is because `Iteration` usually involves state mutation per iteration. Since functional programming avoids state changes, `Recursion` is a commonly used [technique](https://www.quora.com/Why-dont-pure-functional-programming-languages-provide-a-loop-construct) to replace `Iteration`.
 
 Hence, to be able to write functional languages effectively, it means being able to replace Iteration with Recursion. Here are some guides to help you on that:
 
@@ -101,7 +101,7 @@ The common question when learning about functional programming is what's the poi
 
 ## Disadvantages of Functional Programming
 
-Functional programming is not a new concept and has been around since the 1950s. Even though it is gaining in popularity today, it is not the predominant programming paradigm used in software applications today. Despite the stated benefits of Functional programming, there are some downsides of it that can help explain why it is not the mainstream programming paradigm:
+Functional programming is not a new concept and has been around since the 1950s. Even though it is gaining in popularity today, it is not the predominant programming paradigm used in software applications today. Despite the stated benefits of functional programming, there are some downsides of it that can help explain why it is not the mainstream programming paradigm:
 
 - Lack of imperative data structures. One good example is the [lack of hash map](https://stackoverflow.com/questions/6793259/how-does-one-implement-hash-tables-in-a-functional-language) which is an important performant dictionary.
 - Functional programming can be [slower](https://www.quora.com/Do-functional-programming-languages-always-run-slower-than-imperative-language) than optimal Imperative programming for reasons such as data copying due to data immutability.
--- a/contents/functionalProgramming/intro.md
+++ b/contents/functionalProgramming/intro.md
@@ -8,7 +8,7 @@ Functional programming is a programming paradigm that treats computation as the 
 
 ## Functional Programming Languages
 
-Functional programming is simply a paradigm and needs to be implemented by programming languages. Below are various [languages](https://en.wikipedia.org/wiki/List_of_programming_languages_by_type#Functional_languages) that have explicit support for the functional programming paradigm such as:
+Functional programming is simply a paradigm and needs to be implemented by programming languages. Below are various [languages](https://en.wikipedia.org/wiki/List_of_programming_languages_by_type#Functional_languages) that have explicit support for the functional programming paradigm:
 
 - Haskell
 - Clojure
@@ -23,7 +23,7 @@ While functional programming can be implemented by languages like Java, the lang
 
 #### Pure Functions
 
-Pure functions is one of the concepts in functional programming. To call a function `pure`, it needs to satisfy two conditions:
+_Pure functions_ is one of the concepts in functional programming. To call a function _pure_, it needs to satisfy two conditions:
 
 1. [Idempotence](https://en.wikipedia.org/wiki/Idempotence) - The function should return the same output for the same input when invoked arbitrarily many times.
 1. No side effects - The function should not cause side effects like mutating global state.
@@ -41,18 +41,18 @@ Contrast this with an impure function:
 ```python
 # Mutate state
 y=0
-def addXToY(x):
+def add_to_y(x):
     global y
     y = y + x
     return y
 ```
-We see that the impure function violates both conditions. Running `addXToY` multiple times with the same input X would give different value.
+We see that the impure function violates both conditions. Running `add_to_y` multiple times with the same input X would give different value.
 
 #### Immutability
 
-Another concept in functional programming is immutability. Immutability prevents an object's state to be change after it is created. When we assign `x=1`, it does not make mathematical sense to have this expression `x:=x+1`.
+Another concept in functional programming is immutability. Immutability prevents an object's state to be change after it is created.
 
-Having immutability built into functional programming languages like Haskell also helps to ensure that functions are `pure` because mutable variables and states can introduce side-effects.
+Having immutability built into functional programming languages like Haskell helps ensure that functions are _pure_ because mutable variables and states can introduce side-effects.
 
 To know more about immutability in functional languages, you can take a look at:
 
@@ -64,7 +64,7 @@ To know more about immutability in functional languages, you can take a look at:
 
 #### Recursion
 
-Pure functional languages do not have loop constructs like imperative languages. This is because `Iteration` usually involves state mutation per iteration. Since functional programming avoids state changes, `Recursion` is a commonly used [technique](https://www.quora.com/Why-dont-pure-functional-programming-languages-provide-a-loop-construct) to replace `Iteration`. An example of replacing state-mutating iterative code to a pure functional recusive code is shown below:
+Pure functional languages do not have loop constructs like imperative languages. This is because `Iteration` usually involves state mutation per iteration. Because functional programming avoids state changes, `Recursion` is a commonly used [technique](https://www.quora.com/Why-dont-pure-functional-programming-languages-provide-a-loop-construct) to replace `Iteration`. An example of replacing state-mutating iterative code to a pure functional recusive code is shown below:
 
 ```Python
 # Iterative
@@ -88,7 +88,7 @@ Hence, to be able to write functional languages effectively, it means being able
 
 #### Higher Order Functions
 
-Higher Order Functions are functions that either take functions as parameters, return a function or both. As functions are first class citizens in functional programming languages, this allows [functions to be passed around like objects and values](https://en.wikipedia.org/wiki/Functional_programming#First-class_and_higher-order_functions).
+Higher Order Functions are functions that either take functions as parameters, return a function, or both. As functions are first class citizens in functional programming languages, this allows [functions to be passed around like objects and values](https://en.wikipedia.org/wiki/Functional_programming#First-class_and_higher-order_functions).
 
 One example of Higher Order Functions is `map`, which takes in a function and a collection of objects.
 
@@ -108,8 +108,8 @@ There are useful functions and techniques that are based off Higher Order Functi
 
 The common question when learning about functional programming is what's the point of having us going through all the trouble of having pure functions, state immutability and absence of loops. Turns out, there are various positive implications of functional programming:
 
-- [More Bug-Free Code](https://www.quora.com/Are-software-written-using-Functional-Programming-less-buggy-more-robust-and-stable) for example famously in Haskell, [if it compiles it usually works](https://wiki.haskell.org/Why_Haskell_just_works).
-- [Concurrency](https://softwareengineering.stackexchange.com/questions/293851/what-is-it-about-functional-programming-that-makes-it-inherently-adapted-to-para). Having state immutability and absence of side effects can allow for bug-free concurrent code with less concerns of race conditions on mutable objects.
+- [More Bug-Free Code](https://www.quora.com/Are-software-written-using-Functional-Programming-less-buggy-more-robust-and-stable). For example famously in Haskell, [if it compiles it usually works](https://wiki.haskell.org/Why_Haskell_just_works).
+- [Safer Concurrency](https://softwareengineering.stackexchange.com/questions/293851/what-is-it-about-functional-programming-that-makes-it-inherently-adapted-to-para). Having state immutability and absence of side effects can allow for concurrent code with less concerns of race conditions on mutable objects.
 - [And many more](https://alvinalexander.com/scala/fp-book/benefits-of-functional-programming) such as simpler unit testing as functions are guaranteed not to have any side effects.
 
 ## Disadvantages of Functional Programming
@@ -124,7 +124,7 @@ Some of these reasons help in part explain why 'impure' functional languages lik
 
 ## Guides on Functional Programming
 
-Functional programming can be a very different programming paradigm and it definitely takes time to practice to be an expert at it. To learn more about functional programming, I would recommend looking at these amazing guides for a deeper understanding:
+Functional programming can be a very different programming paradigm and it definitely takes time to practice to be an expert at it. To learn more about functional programming, take a look at these amazing guides for a deeper understanding:
 
 - A good [overview of functional programming](https://en.wikipedia.org/wiki/Functional_programming)
 - A six-part [introduction to functional programming](https://medium.com/@cscalfani/so-you-want-to-be-a-functional-programmer-part-1-1f15e387e536)
--- a/contents/functionalProgramming/intro.md
+++ b/contents/functionalProgramming/intro.md
@@ -100,7 +100,7 @@ The common question when learning about Functional Programming is what's the poi
 
 ## Guides on Functional Programming
 
-Functional Programming is a new paradigm and it definitely takes time to practice to be an expert at it. To learn more about Functional Programming, I would recommend looking at these amazing guides for a deeper understanding:
+Functional Programming can be a very different programming paradigm and it definitely takes time to practice to be an expert at it. To learn more about Functional Programming, I would recommend looking at these amazing guides for a deeper understanding:
 
 - [Introduction to Functional Programming](https://medium.com/@cscalfani/so-you-want-to-be-a-functional-programmer-part-1-1f15e387e536)
 - [Functional Programming Principles course](https://www.coursera.org/learn/progfun1)
--- a/contents/functionalProgramming/intro.md
+++ b/contents/functionalProgramming/intro.md
@@ -1,0 +1,107 @@
+# An Introduction to Functional Programming
+
+Authors: [Phang Chun Rong](https://github.com/crphang)
+
+## What is Functional Programming
+
+Functional Programming is a programming paradigm designed based on mathematical functions and avoid mutating state and data. Unlike procedure programming languages like C, Java and Python, Functional Programming languages are [declarative](https://en.wikipedia.org/wiki/Declarative_programming). Even though this demands a shift in mindset from most programmers who are used to imperative form of programming, Functional Programming paradigm is still [increasingly popular](https://blog.appdynamics.com/engineering/the-most-popular-programming-languages-for-2017/) due to its [advantages](#advantages-of-functional-programming).
+
+## Functional Programming Languages
+
+Functional Programming is simply a paradigm and needs to be implemented by programming languages. Below are various [languages](https://en.wikipedia.org/wiki/List_of_programming_languages_by_type#Functional_languages) that have explicit support for Functional Programming paradigm such as:
+
+- Haskell
+- Clojure
+- Elixir
+- Elm
+- Scala
+
+While functional programming can be implemented by languages like Java, the languages listed above encourage Functional Programming paradigm such as [pure functions](#pure-functions) or even enforce them in the case of Haskell.
+
+## Concepts in Functional Programming
+
+#### Pure Functions
+
+Pure functions is one of concepts in functional programming. To call a function `pure`, it needs to satisfy two conditions:
+
+1. [Idempotence](https://en.wikipedia.org/wiki/Idempotence) - The function should return the same output for the same input.
+1. No Side Effects - The function should not cause side effects like mutating global state.
+
+To illustrate an example of a pure function. Consider this simple example:
+
+```python
+# Pure Function
+def add(x, y):
+    return x + y
+```
+
+Contrast this with an impure function:
+
+```python
+# Mutate state
+y=0
+def addXToY(x):
+    global y
+    y = y + x
+    return y
+```
+We see that the impure function violates both conditions. Running `addXToY` multiple times with the same input X would give different value.
+
+#### Immutability
+
+Another concept in Functional Programming is immutability. When we assign `x=1`, it does not make mathematical sense to have this expression `x:=x+1`.
+
+Having immutability built into Functional Programming languages like Haskell also helps to ensure that functions are `pure` because mutable variables and states can introduce side-effects.
+
+To know more about immutability in functional languages, you can take a look at:
+
+- [Immutability in Haskell](https://mmhaskell.com/blog/2017/1/9/immutability-is-awesome)
+- [Immutability in Elm](http://elmprogramming.com/immutability.html)
+- [Why does immutability enable Functional Programming](https://stackoverflow.com/questions/12207757/why-do-immutable-objects-enable-functional-programming)
+
+## Techniques in Functional Programming
+
+#### Recursion
+
+Pure functional languages does not have loop constructs that procedural languages does. This is because `Iteration` usually involves state mutation per iteration. Since Functional Programming avoids state changes, `Recursion` is a commonly used [technique](https://www.quora.com/Why-dont-pure-functional-programming-languages-provide-a-loop-construct) to replace `Iteration`.
+
+Hence, to be able to write functional languages effectively, it means being able to replace Iteration with Recursion. Here are some guides to help you on that:
+
+- [Introduction to Recursion](https://www.topcoder.com/community/data-science/data-science-tutorials/an-introduction-to-recursion-part-1/)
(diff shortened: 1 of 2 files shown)


crphang:
@LiHaoTan Since your chapter on Rust has got to do with Functional Programming, would appreciate if you can take a look at this PR!

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,107 @@
...
+Functions are first class citizen in Functional Programming languages, this mean that functions can be passed around like objects and values. 
+
+One example of Higher Order Functions is `map`, which takes in a function and a collection of objects.
+
+```python
+length_of_names = map(len, ['Alex', 'Thomas', 'John'])
+print(length_of_names) # [4,6,4]
+```
+
+The function creates a new collection, run the function on each element of the given collection, insert the new values and return the new collection.
```
tran-tien-dat: This explanation of map is not very clear for a beginner. It doesn't even mention the name `map` so it may not be obvious to the reader. It also repeats words like `function`, `collection` many times, and each instance of the word refers to a different function/collection. I suggest something like: "`map` takes in 2 parameters, the first one is a function and the second is a collection. `map` returns a new collection whose elements are the result of applying the given function on the elements of the given collection". Even better, I think you can refer to the example you quoted, e.g. "In the example, `map` is given a function `len` and a collection `['Alex', 'Thomas', 'John']`. It returns `[len('Alex'), len('Thomas'), len('John')]`."

tran-tien-dat (commented):
Hi Chun Rong. Couldn't sleep yet so I read your book chapter. Nice introduction to functional programming. Would be more comprehensive if you can also talk about the disadvantages of functional programming, or when would we not want to use functional programming.

crphang:
@tran-tien-dat Thanks a lot for the review! I've added the changes.

tran-tien-dat (approved):
Looks great 👍

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,117 @@
...
+
+## What is Functional Programming
+
+Functional Programming is a programming paradigm designed based on mathematical functions and avoid mutating state and data. Unlike procedure programming languages like C, Java and Python, Functional Programming languages are [declarative](https://en.wikipedia.org/wiki/Declarative_programming). Even though this demands a shift in mindset from most programmers who are used to imperative form of programming, Functional Programming paradigm is still [increasingly popular](https://blog.appdynamics.com/engineering/the-most-popular-programming-languages-for-2017/) due to its [advantages](#advantages-of-functional-programming).
+
+## Functional Programming Languages
+
+Functional Programming is simply a paradigm and needs to be implemented by programming languages. Below are various [languages](https://en.wikipedia.org/wiki/List_of_programming_languages_by_type#Functional_languages) that have explicit support for Functional Programming paradigm such as:
+
+- Haskell
```
LiHaoTan: Actually, I'm wondering why these selected languages? Like Haskell is famous, Clojure got more popular, Elixir works on the Erlang VM, Elm translates to Javascript. But it seems the ML family of languages ought to deserve a mention too?

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,117 @@
+# An Introduction to Functional Programming
+
+Authors: [Phang Chun Rong](https://github.com/crphang)
+
+## What is Functional Programming
+
+Functional Programming is a programming paradigm designed based on mathematical functions and avoid mutating state and data. Unlike procedure programming languages like C, Java and Python, Functional Programming languages are [declarative](https://en.wikipedia.org/wiki/Declarative_programming). Even though this demands a shift in mindset from most programmers who are used to imperative form of programming, Functional Programming paradigm is still [increasingly popular](https://blog.appdynamics.com/engineering/the-most-popular-programming-languages-for-2017/) due to its [advantages](#advantages-of-functional-programming).
```
LiHaoTan: Just a nit but I think it should be Functional **programming**, not capitalized (across the document).

Also, I think Wikipedia's definition reads better:

> that treats computation as the evaluation of mathematical functions 

> designed based on mathematical functions

The second one seems more vague?

Also I think instead of **procedural**, we should use **imperative**.

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,117 @@
...
+- Elm
+- Scala
+
+While functional programming can be implemented by languages like Java, the languages listed above encourage Functional Programming paradigm such as [pure functions](#pure-functions) or even enforce them in the case of Haskell.
+
+## Concepts in Functional Programming
+
+#### Pure Functions
+
+Pure functions is one of concepts in functional programming. To call a function `pure`, it needs to satisfy two conditions:
```
LiHaoTan: ... one of **the** ...

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,117 @@
...
+
+1. [Idempotence](https://en.wikipedia.org/wiki/Idempotence) - The function should return the same output for the same input.
+1. No Side Effects - The function should not cause side effects like mutating global state.
```
LiHaoTan: Just like functional programming, I don't think **Side Effects** should be capitalized.

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,117 @@
...
+
+To illustrate an example of a pure function. Consider this simple example:
```
LiHaoTan: Not sure if we should mention ...this simple example **in Python**:

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,117 @@
...
+Hence, to be able to write functional languages effectively, it means being able to replace Iteration with Recursion. Here are some guides to help you on that:
+
+- [Introduction to Recursion](https://www.topcoder.com/community/data-science/data-science-tutorials/an-introduction-to-recursion-part-1/)
+- [Recursion in Functional Programming](https://dzone.com/articles/functional-programming-recursion)
+
+#### Higher Order Functions
+
+> Higher Order Functions are functions that take functions as parameters, return functions or both.
+
+Functions are first class citizen in Functional Programming languages, this mean that functions can be passed around like objects and values.
```
LiHaoTan: **first class citizens**

And practically I don't think it makes much difference differentiating higher order and first-class functions but I think a brief note and link they are different will help prevent confusion?

https://en.wikipedia.org/wiki/Functional_programming#First-class_and_higher-order_functions

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,117 @@
...
+There are useful functions and techniques that are based off Higher Order Functions in Functional Programming.
+
+- [Currying](https://hackernoon.com/javascript-and-functional-programming-currying-pt-4-96e3230782ab)
+- [Map, Reduce and Filter](https://medium.freecodecamp.org/higher-order-functions-in-javascript-d9101f9cf528) 
+
+## Advantages of Functional Programming
+
+The common question when learning about Functional Programming is what's the point of having us going through all the trouble of having pure functions, state immutability and absence of loops. Turns out, there are various positive implications of Functional Programming:
+
+- [More Bug-Free Code](https://www.quora.com/Are-software-written-using-Functional-Programming-less-buggy-more-robust-and-stable)
```
LiHaoTan: Maybe we can strengthen this with the infamous Haskell:

[More Bug-Free Code](https://www.quora.com/Are-software-written-using-Functional-Programming-less-buggy-more-robust-and-stable), for example famously in Haskell [if it compiles it usually works](https://wiki.haskell.org/Why_Haskell_just_works).

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,117 @@
...
+
+- Lack of imperative data structures. One good example is the [lack of hash map](https://stackoverflow.com/questions/6793259/how-does-one-implement-hash-tables-in-a-functional-language) which is an important performant dictionary.
+- Functional Programming can be [slower](https://www.quora.com/Do-functional-programming-languages-always-run-slower-than-imperative-language) than optimal Imperative Programming for reasons such as data copying due to data immutability.
+- And various [other concerns](http://flyingfrogblog.blogspot.sg/2016/05/disadvantages-of-purely-functional.html)
+
+Some of these reasons help explain why 'impure' Functional languages like Scala is much more popular today than Haskell. Scala is a general purpose language that has interoperability with Java and has support for both popular Object Oriented Programming and Functional Programming paradigm. This establishes a [middle ground](https://cacm.acm.org/magazines/2014/4/173220-unifying-functional-and-object-oriented-programming-with-scala/fulltext) for programmers new to Functional Programming.
```
LiHaoTan: Just a nit: ... in part explain... 

because Scala's popularity probably also comes from a variety of other factors such as being on the JVM and the pitch as a Java alternative with the built-in rich concurrency support.

On the other hand, Haskell probably has some concerns such as the complicated type system.

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Guides on Functional Programming
+
+Functional Programming can be a very different programming paradigm and it definitely takes time to practice to be an expert at it. To learn more about Functional Programming, I would recommend looking at these amazing guides for a deeper understanding:
+
+- [Introduction to Functional Programming](https://medium.com/@cscalfani/so-you-want-to-be-a-functional-programmer-part-1-1f15e387e536)
```
LiHaoTan: I think these are very high quality curated resources, nice! (I haven't took that course myself though just knew that it is very famous 😆 ). I also could even see clearly the different intentions. I'm suggesting that maybe we make it even better, something like this:

* An introduction to functional programming, only a three-part series!
* If you are hungry for more, take the excellent functional programming principles course
* And to help with the mindset shift for functional programming, take a look at a practical guide on how to translate an imperative to functional style

Also, I think maybe we should include the [wikipedia](https://en.wikipedia.org/wiki/Functional_programming) page too because this is one of the times it has an excellent overview of functional programming.

LiHaoTan (commented):
I think this is an overall very nice introduction to functional programming that addresses the relevant points. 

Just mostly some nits and suggestions

crphang:
@LiHaoTan added the changes. Can I have another review?

contents/functionalProgramming/intro.md
```diff
@@ -4,11 +4,11 @@ Authors: [Phang Chun Rong](https://github.com/crphang)
...
 
 ## Functional Programming Languages
 
-Functional programming is simply a paradigm and needs to be implemented by programming languages. Below are various [languages](https://en.wikipedia.org/wiki/List_of_programming_languages_by_type#Functional_languages) that have explicit support for Functional programming paradigm such as:
+Functional programming is simply a paradigm and needs to be implemented by programming languages. Below are various [languages](https://en.wikipedia.org/wiki/List_of_programming_languages_by_type#Functional_languages) that have explicit support for functional programming paradigm such as:
```
LiHaoTan: Should it be ...explicit support for the functional programming paradigm...?

contents/functionalProgramming/intro.md
```diff
@@ -17,7 +17,7 @@ Functional programming is simply a paradigm and needs to be implemented by progr
 - Scala
 - ML (Meta Language) family of languages
 
-While functional programming can be implemented by languages like Java, the languages listed above encourage Functional programming paradigm such as [pure functions](#pure-functions) or even enforce them in the case of Haskell.
+While functional programming can be implemented by languages like Java, the languages listed above encourage functional programming paradigm such as [pure functions](#pure-functions) or even enforce them in the case of Haskell.
```
LiHaoTan: Should it be ...explicit support for the functional programming paradigm...?

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,119 @@
...
+The common question when learning about functional programming is what's the point of having us going through all the trouble of having pure functions, state immutability and absence of loops. Turns out, there are various positive implications of functional programming:
+
+- [More Bug-Free Code](https://www.quora.com/Are-software-written-using-Functional-Programming-less-buggy-more-robust-and-stable) for example famously in Haskell, [if it compiles it usually works](https://wiki.haskell.org/Why_Haskell_just_works).
+- [Concurrency](https://softwareengineering.stackexchange.com/questions/293851/what-is-it-about-functional-programming-that-makes-it-inherently-adapted-to-para)
+- [And Many More](https://alvinalexander.com/scala/fp-book/benefits-of-functional-programming)
+
+## Disadvantages of Functional Programming
+
+Functional programming is not a new concept and has been around since the 1950s. Even though it is gaining in popularity today, it is not the predominant programming paradigm used in software applications today. Despite the stated benefits of Functional programming, there are some downsides of it that can help explain why it is not the mainstream programming paradigm:
```
LiHaoTan: You missed the capitalization `...stated benefits of Functional programming...` here

contents/functionalProgramming/intro.md
```diff
@@ -50,32 +50,32 @@ We see that the impure function violates both conditions. Running `addXToY` mult
...
-- [Why does immutability enable Functional programming](https://stackoverflow.com/questions/12207757/why-do-immutable-objects-enable-functional-programming)
+- [Why does immutability enable functional programming](https://stackoverflow.com/questions/12207757/why-do-immutable-objects-enable-functional-programming)
 
 ## Techniques in Functional Programming
 
 #### Recursion
 
-Pure functional languages does not have loop constructs that procedural languages does. This is because `Iteration` usually involves state mutation per iteration. Since Functional programming avoids state changes, `Recursion` is a commonly used [technique](https://www.quora.com/Why-dont-pure-functional-programming-languages-provide-a-loop-construct) to replace `Iteration`.
+Pure functional languages does not have loop constructs that procedural languages does. This is because `Iteration` usually involves state mutation per iteration. Since functional programming avoids state changes, `Recursion` is a commonly used [technique](https://www.quora.com/Why-dont-pure-functional-programming-languages-provide-a-loop-construct) to replace `Iteration`.
```
LiHaoTan: Suggestion: Pure functional languages do not have loop constructs like imperative languages.

LiHaoTan (approved):
Looks good! Just some small suggestions

crphang:
@damithc Ready for review!

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,119 @@
+# An Introduction to Functional Programming
+
+Authors: [Phang Chun Rong](https://github.com/crphang)
+
+## What is Functional Programming
+
+Functional programming is a programming paradigm that treats computation as the evaluation of mathematical function and avoids mutating state and variables. Unlike imperative programming languages like C, Java and Python, functional programming languages are [declarative](https://en.wikipedia.org/wiki/Declarative_programming). Even though this demands a shift in mindset from most programmers who are used to imperative form of programming, functional programming paradigm is still [increasingly popular](https://blog.appdynamics.com/engineering/the-most-popular-programming-languages-for-2017/) due to its [advantages](#advantages-of-functional-programming).
```
damithc: This paragraph assumes the reader knows the meaning of _imperative_ and _mutating state_, but not the meaning of _functional programming_. That is a bit unrealistic. People may be used to imperative languages but they might not know that they are imperative languages. On a related note, languages like Python support multiple paradigms; it's not correct to present them as of a certain paradigm.

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,119 @@
...
+
+While functional programming can be implemented by languages like Java, the languages listed above encourage the functional programming paradigm such as [pure functions](#pure-functions) or even enforce them in the case of Haskell.
+
+## Concepts in Functional Programming
+
+#### Pure Functions
+
+Pure functions is one of the concepts in functional programming. To call a function `pure`, it needs to satisfy two conditions:
+
+1. [Idempotence](https://en.wikipedia.org/wiki/Idempotence) - The function should return the same output for the same input.
```
damithc: This definition need to mention 'multiple invocations' somewhere.

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,119 @@
...
+```
+We see that the impure function violates both conditions. Running `addXToY` multiple times with the same input X would give different value.
+
+#### Immutability
+
+Another concept in functional programming is immutability. When we assign `x=1`, it does not make mathematical sense to have this expression `x:=x+1`.
+
+Having immutability built into functional programming languages like Haskell also helps to ensure that functions are `pure` because mutable variables and states can introduce side-effects.
+
+To know more about immutability in functional languages, you can take a look at:
```
damithc: May be explain immutability a bit more rather than point to external pages immediately? Reading external links should be optional.

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,119 @@
...
+- [Immutability in Elm](http://elmprogramming.com/immutability.html)
+- [Why does immutability enable functional programming](https://stackoverflow.com/questions/12207757/why-do-immutable-objects-enable-functional-programming)
+
+## Techniques in Functional Programming
+
+#### Recursion
+
+Pure functional languages do not have loop constructs like imperative languages. This is because `Iteration` usually involves state mutation per iteration. Since functional programming avoids state changes, `Recursion` is a commonly used [technique](https://www.quora.com/Why-dont-pure-functional-programming-languages-provide-a-loop-construct) to replace `Iteration`.
+
+Hence, to be able to write functional languages effectively, it means being able to replace Iteration with Recursion. Here are some guides to help you on that:
```
damithc: Give an example before pointing to external resources?

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,119 @@
...
+
+- [Introduction to Recursion](https://www.topcoder.com/community/data-science/data-science-tutorials/an-introduction-to-recursion-part-1/)
+- [Recursion in functional programming](https://dzone.com/articles/functional-programming-recursion)
+
+#### Higher Order Functions
+
+> Higher Order Functions are functions that take functions as parameters, return functions or both.
```
damithc: This phrasing is a bit unclear. For example, If a function returns a function but does not take a function as a parameter, is it sill a higher order function? Did you mean this?
>Higher Order Functions take functions as parameters and/or return functions as the result.

damithc: In fact, this need not be given as a quote, especially if you are not citing a source. It can be given as normal text.

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,119 @@
...
+- [Currying](https://hackernoon.com/javascript-and-functional-programming-currying-pt-4-96e3230782ab)
+- [Map, Reduce and Filter](https://medium.freecodecamp.org/higher-order-functions-in-javascript-d9101f9cf528) 
+
+## Advantages of Functional Programming
+
+The common question when learning about functional programming is what's the point of having us going through all the trouble of having pure functions, state immutability and absence of loops. Turns out, there are various positive implications of functional programming:
+
+- [More Bug-Free Code](https://www.quora.com/Are-software-written-using-Functional-Programming-less-buggy-more-robust-and-stable) for example famously in Haskell, [if it compiles it usually works](https://wiki.haskell.org/Why_Haskell_just_works).
+- [Concurrency](https://softwareengineering.stackexchange.com/questions/293851/what-is-it-about-functional-programming-that-makes-it-inherently-adapted-to-para)
+- [And Many More](https://alvinalexander.com/scala/fp-book/benefits-of-functional-programming)
```
damithc: Here too, give a bit more details. Otherwise the reader is forced to go read the external link to understand what you mean.

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,119 @@
...
+
+- Lack of imperative data structures. One good example is the [lack of hash map](https://stackoverflow.com/questions/6793259/how-does-one-implement-hash-tables-in-a-functional-language) which is an important performant dictionary.
+- Functional programming can be [slower](https://www.quora.com/Do-functional-programming-languages-always-run-slower-than-imperative-language) than optimal Imperative programming for reasons such as data copying due to data immutability.
+- And various [other concerns](http://flyingfrogblog.blogspot.sg/2016/05/disadvantages-of-purely-functional.html)
+
+Some of these reasons help in part explain why 'impure' functional languages like Scala is much more popular today than Haskell. Scala is a general purpose language that has interoperability with Java and has support for both object-oriented and functional programming paradigms. This establishes a [middle ground](https://cacm.acm.org/magazines/2014/4/173220-unifying-functional-and-object-oriented-programming-with-scala/fulltext) for programmers new to functional programming.
+
```
damithc: There was no mention of OOP until this point. The reader will wonder how it fits into the whole picture.

damithc (changes requested):
Looks good 👍 
Some suggestions added. I'll leave you to decide which ones to follow.

crphang:
@damithc Changes are added, ready for another round of review.

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,132 @@
...
+Authors: [Phang Chun Rong](https://github.com/crphang)
+
+## What is Functional Programming
+
+Functional programming is a programming paradigm that treats computation as the evaluation of mathematical function and avoids mutating state and variables. Unlike [imperative programming](https://en.wikipedia.org/wiki/Imperative_programming) languages like C, functional programming languages are [declarative](https://en.wikipedia.org/wiki/Declarative_programming). Even though this demands a shift in mindset from most programmers who are used to imperative form of programming, functional programming paradigm is still [increasingly popular](https://blog.appdynamics.com/engineering/the-most-popular-programming-languages-for-2017/) due to its [advantages](#advantages-of-functional-programming).
+
+## Functional Programming Languages
+
+Functional programming is simply a paradigm and needs to be implemented by programming languages. Below are various [languages](https://en.wikipedia.org/wiki/List_of_programming_languages_by_type#Functional_languages) that have explicit support for the functional programming paradigm such as:
```
damithc: `such as` is redundant?

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,132 @@
...
+- Scala
+- ML (Meta Language) family of languages
+
+While functional programming can be implemented by languages like Java, the languages listed above encourage the functional programming paradigm such as [pure functions](#pure-functions) or even enforce them in the case of Haskell.
+
+## Concepts in Functional Programming
+
+#### Pure Functions
+
+Pure functions is one of the concepts in functional programming. To call a function `pure`, it needs to satisfy two conditions:
```
damithc: It might help if you use italics here.
>_Pure functions_ is one of the concepts ...

damithc: using code font for `pure` is not appropriate here. Consider `_pure_` or `'pure'`

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,132 @@
...
+def add(x, y):
+    return x + y
+```
+
+Contrast this with an impure function:
+
+```python
+# Mutate state
+y=0
+def addXToY(x):
```
damithc: Use a more pythonic name? e.g., `add_to_y(x)`

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,132 @@
...
+    global y
+    y = y + x
+    return y
+```
+We see that the impure function violates both conditions. Running `addXToY` multiple times with the same input X would give different value.
+
+#### Immutability
+
+Another concept in functional programming is immutability. Immutability prevents an object's state to be change after it is created. When we assign `x=1`, it does not make mathematical sense to have this expression `x:=x+1`.
```
damithc: > it does not make mathematical sense to have this expression `x:=x+1`

While this is true, I'm not sure that equates to an explanation or even a justification of immutability. The sentence  sounds like this equation doesn't make sense, mathematically. i.e., the syntax is nonsensical

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,132 @@
...
+    for number in listOfNumbers:
+        sum += number
+    return sum
+
+# Recursive
+def getSumOfListRec(listOfNumbers):
+    if len(listOfNumbers) == 0:
+        return 0
+    return listOfNumbers[0] + getSumOfListRec(listOfNumbers[1:])
+```
```
damithc: Give as two code segments. Otherwise this code will be rejected by Python?

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,132 @@
...
+
+Hence, to be able to write functional languages effectively, it means being able to replace Iteration with Recursion. Here are some guides to help you on that:
+
+- [Introduction to Recursion](https://www.topcoder.com/community/data-science/data-science-tutorials/an-introduction-to-recursion-part-1/)
+- [Recursion in functional programming](https://dzone.com/articles/functional-programming-recursion)
+
+#### Higher Order Functions
+
+Higher Order Functions are functions that either take functions as parameters, return a function or both. As functions are first class citizens in functional programming languages, this allows [functions to be passed around like objects and values](https://en.wikipedia.org/wiki/Functional_programming#First-class_and_higher-order_functions).
```
damithc: Use oxford comma. ` xxx, yyy, or zzz`.

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,132 @@
...
+
+- [Immutability in Haskell](https://mmhaskell.com/blog/2017/1/9/immutability-is-awesome)
+- [Immutability in Elm](http://elmprogramming.com/immutability.html)
+- [Why does immutability enable functional programming](https://stackoverflow.com/questions/12207757/why-do-immutable-objects-enable-functional-programming)
+
+## Techniques in Functional Programming
+
+#### Recursion
+
+Pure functional languages do not have loop constructs like imperative languages. This is because `Iteration` usually involves state mutation per iteration. Since functional programming avoids state changes, `Recursion` is a commonly used [technique](https://www.quora.com/Why-dont-pure-functional-programming-languages-provide-a-loop-construct) to replace `Iteration`. An example of replacing state-mutating iterative code to a pure functional recusive code is shown below:
```
damithc: Prefer `Because` over `Since` (latter has two meanings)

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,132 @@
...
+There are useful functions and techniques that are based off Higher Order Functions in Functional programming.
+
+- [Currying](https://hackernoon.com/javascript-and-functional-programming-currying-pt-4-96e3230782ab)
+- [Map, Reduce and Filter](https://medium.freecodecamp.org/higher-order-functions-in-javascript-d9101f9cf528)
+
+## Advantages of Functional Programming
+
+The common question when learning about functional programming is what's the point of having us going through all the trouble of having pure functions, state immutability and absence of loops. Turns out, there are various positive implications of functional programming:
+
+- [More Bug-Free Code](https://www.quora.com/Are-software-written-using-Functional-Programming-less-buggy-more-robust-and-stable) for example famously in Haskell, [if it compiles it usually works](https://wiki.haskell.org/Why_Haskell_just_works).
```
damithc: >[More Bug-Free Code](https://www.quora.com/Are-software-written-using-Functional-Programming-less-buggy-more-robust-and-stable) for example famously in Haskell, [if it compiles it usually works](https://wiki.haskell.org/Why_Haskell_just_works).

Punctuation missing after `More Bug-Free Code`?

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,132 @@
...
+- [Concurrency](https://softwareengineering.stackexchange.com/questions/293851/what-is-it-about-functional-programming-that-makes-it-inherently-adapted-to-para). Having state immutability and absence of side effects can allow for bug-free concurrent code with less concerns of race conditions on mutable objects.
```
damithc: Don't equate _fewer bugs_ with _bug-free_ :-p

damithc: `Concurrency` is not an advantage. Imperative languages can have concurrency too.

contents/functionalProgramming/intro.md
```diff
@@ -0,0 +1,132 @@
...
+- Functional programming can be [slower](https://www.quora.com/Do-functional-programming-languages-always-run-slower-than-imperative-language) than optimal Imperative programming for reasons such as data copying due to data immutability.
+- And various [other concerns](http://flyingfrogblog.blogspot.sg/2016/05/disadvantages-of-purely-functional.html)
+
+Some of these reasons help in part explain why 'impure' functional languages like Scala is much more popular today than Haskell. Scala is a general purpose language that has interoperability with Java and has support for both imperative and functional programming paradigms. This establishes a [middle ground](https://cacm.acm.org/magazines/2014/4/173220-unifying-functional-and-object-oriented-programming-with-scala/fulltext) for programmers new to functional programming.
+
+## Guides on Functional Programming
+
+Functional programming can be a very different programming paradigm and it definitely takes time to practice to be an expert at it. To learn more about functional programming, I would recommend looking at these amazing guides for a deeper understanding:
```
damithc: Avoid using `I`. Reason: in future, more authors can be added.

damithc (changes requested):
Looks good. Some nits.

contents/functionalProgramming/intro.md
crphang: @damithc, can you clarify this?

damithc: I meant show it like this:

----
Iterative:
```Python
def getSumOfList(listOfNumbers):
    sum = 0
    for number in listOfNumbers:
        sum += number
    return sum
```
Recursive:
```Python
def getSumOfList(listOfNumbers):
    if len(listOfNumbers) == 0:
        return 0
    return listOfNumbers[0] + getSumOfList(listOfNumbers[1:])
```

---

At first glance I thought they have the same function name, but on closer inspection they don't. That means the code itself is fine.
But they are alternates --> unlikely to appear in the same code file.

crphang: I ran the code and it works. The functions are named differently as well.

damithc (commented):
Looks like my comment didn't go through earlier.

crphang:
@damithc Changes are added. Ready for another round of review.

damithc (approved):
LGTM 👍

PR #60, opened 2018-04-13
Rvalue references and move semantics in C++

yamidark:
Add an introduction to rvalue references and move semantics in C++.
View the rendered version [here](https://github.com/yamidark/learningresources/blob/f58b66a8942e4e65b1cc3e2287c5e18d1335cc47/contents/c%2B%2B/rvalue.md).


yamidark:
@vivekscl  Mind helping me take a look at this if your free?

contents/c++/rvalue.md
```diff
@@ -0,0 +1,145 @@
...
+    * [Reference Semantics](#reference-semantics)
+* [Move Semantics](#move-semantics)
+    * [rvalues and lvalues references](#rvalues-and-lvalues-references)
+    * [Using rvalue references](#using-rvalue-references)
+    * [Move Semantics](#move-semantics)
+* [Rvalue anti-pattern](#rvalue-anti-pattern)
+* [Resources](#resources)
+
+## Background
+C++ is a general-purpose programming language that was designed for resource-constrained and large systems, high performance and efficiency. The language has been extended and improved after the past decades, with new standards being released periodically every few years. One such standard, the [C++11](https://isocpp.org/wiki/faq/cpp11) standard, brought about many changes and improvements, that it made C++ feel like a different language altogether. One such feature that furthur improves on the performance on the language is the introduction of *rvalue* references and *Move Semantics*.
```
vivekscl: For the first sentence, I think a better phrasing would be:
`C++ is a general-purpose programming language designed to provide high performance and efficiency for resource-constrained and large systems.` The language has been extended and improved **consistently**, with new standards being released periodically ~~every few years~~. `The C++11 is one such standard that` brought about many changes and improvements ~~, that it made C++ feel like a different language altogether~~. One such feature of `C++11`**(Look at note below)**  that furthuer improves on the performance on the language is the introduction of *rvalue* references and *Move Semantics*.

**Note:**
Is the feature from C++11? If so, it should be referenced as `C++11` here.

contents/c++/rvalue.md
```diff
@@ -0,0 +1,145 @@
+# Rvalue references and Move Semantics
+
+Authors: [Tan Jun An](https://github.com/yamidark)
+
+* [Background](#background)
+    * [Value Semantics](#value-semantics)
+    * [Reference Semantics](#reference-semantics)
+* [Move Semantics](#move-semantics)
+    * [rvalues and lvalues references](#rvalues-and-lvalues-references)
```
vivekscl: Broken link. Probably because the title at the bottom, **Rvalue and Lvalue references** as the R and L capitalized.

contents/c++/rvalue.md
```diff
@@ -0,0 +1,145 @@
...
+    * [Using rvalue references](#using-rvalue-references)
+    * [Move Semantics](#move-semantics)
+* [Rvalue anti-pattern](#rvalue-anti-pattern)
+* [Resources](#resources)
+
+## Background
+C++ is a general-purpose programming language that was designed for resource-constrained and large systems, high performance and efficiency. The language has been extended and improved after the past decades, with new standards being released periodically every few years. One such standard, the [C++11](https://isocpp.org/wiki/faq/cpp11) standard, brought about many changes and improvements, that it made C++ feel like a different language altogether. One such feature that furthur improves on the performance on the language is the introduction of *rvalue* references and *Move Semantics*.
+
+Before we move on to talk about Move Semantics and rvalue references, we will first briefly go through the 2 other modes available in C++, *Value Semantics* and *References Semantics*.
```
vivekscl: I think there should be a reason why you are explaining *Value Semantics* and *References Semantics* before moving on to Move Semantics and rvalue references. For instance:
`To understand why and how Move Semantics and rvalue references are used, it's important to understand the two other modes available in C++, Value Semantics and References Semantics.`

contents/c++/rvalue.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Value Semantics
+Value (or copy) semantics is given to users by default when we declare arugments with only the type name (e.g. `int`, `string`). It is the programming style where we are only concerned about the values stored in the objects, rather than the object itself. As such, we will always create an extra copy of the value whenever we pass it to a function (also known as [pass-by-value](http://www.learncpp.com/cpp-tutorial/72-passing-arguments-by-value/)) or when constructing a new object and variable. This ensures that each object (or function) will have their own copy value to use, without having to concern themselves with their originator.
```
vivekscl: >given to users by default

Hmm, maybe this could be phrased better? I don't think programming languages give the users a semantic but rather it's a concept right? So in this case, I think you can just remove this part instead.

>when we declare arugments

Is it only when we declare arguments or when we declare variables as well? Also, it's spelled incorrectly, it should be **arguments**.

>with only the type name

I think `data type` is a better description as compared to `type name`. 

> values stored in the objects, rather than the object itself

In this context, objects refer to the memory location instead of the object in OOP right? So in that case, it should be defined properly.

contents/c++/rvalue.md
```diff
@@ -0,0 +1,145 @@
...
+
+Some of the advantages of using Value Semantics is that it ensures that we no memory management issue, as we won't have any [dangling references](https://www.quora.com/What-is-dangling-reference) to objects that may not exist and no [memory leaks](https://www.geeksforgeeks.org/what-is-memory-leak-how-can-we-avoid/).
```
vivekscl: I think you can organise this section of advantages and disadvantages better. Like:
Some advantages of Value Semantics include:
* No memory management issues because...
* Ensure referential transparency...

contents/c++/rvalue.md
```diff
@@ -0,0 +1,145 @@
...
+
+Another seemingly odd benefit of Value Semantics is that it may speedup programs. If we require accessing the value many times, having your own local copy of the value may be faster than having a pointer to the value and dereferencing it each time. This is especially true in C++ as it encourages Value Semantics, with optimization techniques such as [copy elision](http://en.cppreference.com/w/cpp/language/copy_elision) to help passing-by-value be faster.
+
+However, [1 major flaw](https://www.quora.com/What-are-the-drawbacks-of-pass-by-value-result) of Value Semantics and passing-by-value is having poor performance and scability. When we wish to only read the values without modifying it, creating an additional copy just for this purpose is an unnecessary computation, especially if we scale up to larger objects.
```
vivekscl: This sentence is hard to understand. Why is a copy created when reading a value? And what does `scale up to larger objects` mean?

contents/c++/rvalue.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Reference Semantics
+Reference (or pointer) semantics is another mode available for users in C++. C++ allows users to declare [pointers](http://www.cplusplus.com/doc/tutorial/pointers/) and [references](https://www.geeksforgeeks.org/references-in-c/) to the address of the value. As such, we can pass around these pointers to functions and objects, and all of them will refer to and use the same value (address).
```
vivekscl: >is another mode available for users

Hmm, it's more of a choice available to users rather than a mode right?

>allows users to declare pointers and references to the address of the value

This can be phrased like so:
`allows users to declare pointers and references that point to the address of a value`

and all of them will refer to and use the same ~~value~~ **address**

contents/c++/rvalue.md
```diff
@@ -0,0 +1,145 @@
...
+
+One large advantage of using Reference Semantics is that it allows us pass around large objects without having to create an additional copy, thus improving on the performance and scability of the program. It also allows objects to interact with each other as they can have pointers to and modify the same value.
```
vivekscl: Can organize this section better like the advantages section of Value Semantics as mentioned above.

contents/c++/rvalue.md
```diff
@@ -0,0 +1,145 @@
...
+
+However, Reference Semantics also comes with its own problems. In a multi-threaded environment, having such object pointing to the same value will lead to unnecessary data races and un-deterministic behaviour, and requires extra work in trying to synchronize the values between objects. More details of this can be found in the [Java Concurrency](java/JavaConcurrency.md) and [Java Synchronization](java/JavaSynchronization.md).
```
vivekscl: >having such object 

Having a reference or object? And do you mean many references or just a single one?

**indeterminant** behaviour which requires extra work to synchronize values between objects

I'm not sure what `values between objects` mean. Maybe it could be phrased better?

contents/c++/rvalue.md
```diff
@@ -0,0 +1,145 @@
...
+Reference Semantics and also lead to unintended behaviour if we modify the values in 1 object, which in turn affects the other objects which have pointers to the same values as well, an example shown in this [article](http://www.drdobbs.com/cpp/optimization-calling-by-value-or-by-refe/232400151).
```
vivekscl: Reference Semantics **can** also lead to unintended behaviour. **If** we modify the value~~s~~ **of an** object, ~~which in turn~~ **it** affects ~~the~~ other objects **that** have pointers to the same value~~s~~ as well

contents/c++/rvalue.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Move Semantics
+As mentioned above, 1 major flaw of Value Semantics is that we will create a copy of the value, which can computationally expensive if this value is a large object. However, we may also not want to use Reference Semantics due to its different problems as discussed above. As such, in order to continue using Value Semantics without creating this unnecessary copy, C++ introduced a new mode, Move Semantics.
+
+### Rvalue and Lvalue references
+To understand how Move Semantics work in C++, we first need to know about the distinction between rvalue and lvalue references. 1 simple line to describe the differencce between these 2 is:
```
vivekscl: `it is important to distinguish` between rvalue and lvalue references. ~~1 simple line to describe the differencce between these 2 is:~~

contents/c++/rvalue.md
```diff
@@ -0,0 +1,145 @@
...
+```
+lvalue = rvalue
+```
+where lvalue (left value) is on the left side of the equation, and rvalue (right value) is on the right.
```
vivekscl: **From the example above, the** lvalue (left value) is on the left side of the equation...

Not sure if this sentence is required though. It just states what the code looks like.

contents/c++/rvalue.md
```diff
@@ -0,0 +1,145 @@
...
+### Using rvalue references
+Rvalue references allows us to distinguish an lvalue from an rvalue. In C++11, we can declare an rvalue reference using the `&&` operator:
+```cpp
+int &&rvalue = 55;
+```
+
+We can also convert an lvalue to a rvalue using the `std::move` function:
+```cpp
+int lvalue = 99;
+int &&rvalue2 = std::move(lvalue); // std::move converts an lvalue to a rvalue
```
vivekscl: Not sure if the comment here is required since it is already explained above. :thinking:

contents/c++/rvalue.md
```diff
@@ -0,0 +1,145 @@
...
+}
+
+int main() {
+    int x = 5;
+    print(x); // lvalue method used
+    print(10); // rvalue method used.
+}
+```
+
+Using of these functions overload with rvalue parameters, which take on temporary objects, helps us to write more efficient program using Move Semantics!
```
vivekscl: helps **in** writing more efficient program**s** using Move Semantics!

contents/c++/rvalue.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Move Semantics
+The main usage of rvalue references is that it allows us to create *move* constructor and *move* assignments, instead of *copy* constructor and *copy* assignments by default. Since rvalues are typically temporary objects, we can just *move* the value instead of *copying* it, thus helping us to save on memory consumption and improving on the performance on our program as well!
```
vivekscl: Rvalue references allows us to create *move* constructor**s** and *move* assignments, instead of *copy* constructor**s** and *copy* assignments by default. Since rvalues are typically temporary objects, we can just *move* the value instead of *copying* it, thus **reducing** memory consumption and improving performance.

contents/c++/rvalue.md
```diff
@@ -0,0 +1,145 @@
...
+
+One example implementation of a move constructor and move assignment is:
```
vivekscl: ...move assignment is **shown below**:

contents/c++/rvalue.md
```diff
@@ -0,0 +1,145 @@
...
+
+Foo& operator=(Foo&& other) { // move assignment
+    x = std::move(other.x);
+    y = std::move(other.y);
+    z = std::move(other.z);
+    return *this;
+}
+```
+
+In this move constructor and assignment, the contents of the `other` parameter is *moved* into the object, and the contents in `other` is destroyed afterwards. No additional memory allocation occurs, and the *move* operation is done quickly by a few assignment of address operations, leading to a more performance and memory efficient program.
```
vivekscl: No additional memory allocation **is required** and the *move* operation is done quickly by a few assignment**s** of address operations, leading to a **faster** and more memory efficient program.

contents/c++/rvalue.md
```diff
@@ -0,0 +1,145 @@
...
+
+Along with Move Semantics in C++11, the [STL library](https://www.geeksforgeeks.org/the-c-standard-template-library-stl/) functions also defines the overloaded *move* functions for its container classes, thus we can take advantage of these Move Semantics by simply supplying rvalues, without the need to redefine the classes ourselves.
```
vivekscl: Along with Move Semantics in C++11, the STL library ~~functions also defines the~~ **provides** overloaded move functions for its container classes, thus we can take advantage of these Move Semantics by simply supplying rvalues, without the need to redefine the classes ourselves

Maybe it's better to explain what the container classes here refer to.

contents/c++/rvalue.md
```diff
@@ -0,0 +1,145 @@
...
+    _y = y;
+}
+
+Foo(std::string&& x, std::string&& y) { // move constructor
+    _x = x;
+    _y = y;
+}
+```
+
+As we can see, the *move* constructor defines each of its parameter to be rvalues. In this case, if we were to call the constructor using a mixture of both, lvalue for `x` and rvalue for `y`, the copy constructor instead will be called! This is because the copy constructor and take both lvalues and rvalues, both the *move* constructor can only take rvalues. As such, we may think we have made use of Move Semantics to optimize our program, but that may not be the case!
```
vivekscl: >This is because the copy constructor and take both lvalues and rvalues, both the *move* constructor can only take rvalues.

This sentence is rather hard to understand as well.

contents/c++/rvalue.md
```diff
@@ -0,0 +1,145 @@
...
+
+1 solution to this problem would be to overload the constructor for each combination of rvalue parameters possible. However, this is not feasible, especially when the function have many parameters, as we would require `2^n` overload functions where `n` is the number of parameters, leading to too much [boilerplate code](https://www.quora.com/What-is-boilerplate-code), reducing code quality, and increasing memory consumption and compiliation time.
```
vivekscl: **One** solution to this problem would be to overload the constructor for **every possible** combination of **the** rvalue parameters. However, this is not feasible, especially when the function **has** many parameters, as we would require 2^n overload functions where n is the number of parameters, leading to too much boilerplate code, reducing code quality, and increasing memory consumption and compiliation time.

This sentence is rather long. Should be broken up into separate sentences.

vivekscl (changes requested):
Overall, the content is detailed with a good explanation of concepts! However, I'm not very familiar with C++, so I only have a few suggestions regarding the content. On the other hand, the structure and phrasing need to be improved upon. 

On a side note, I don't think the word `we` should be used throughout the chapter as this chapter is trying to provide information from an objective point of view. Also, consider using words instead of numbers i.e. `two` instead of `2`.

Do consider spell checking the entire chapter as there's quite a few spelling mistakes.

There are quite a few abstract terms that are not explained either such as `container classes`. Providing definitions for them would help.

There is a need to improve phrasing as well. Many concepts are hard to understand due to the somewhat awkward phrasings and redundancy in sentences.

The resources section needs some work as well. The current namings of the links need to be more descriptive.

contents/c++/rvalue.md
yamidark: >Hmm, maybe this could be phrased better? I don't think programming languages give the users a semantic but rather it's a concept right?

Hmm, maybe we can move this to the end of the paragraph, something like:
`By default, C++ uses this mode if we declare variables with only the data type` ?

>In this context, objects refer to the memory location instead of the object in OOP right? So in that case, it should be defined properly.

Do you mean to mention `memory address` instead? Because I think its easier to understand and compare the concept to OOP if we mention `objects`.

contents/c++/rvalue.md
yamidark: >his sentence is hard to understand. Why is a copy created when reading a value? 

Hmm, maybe I should add `If we pass the value to a function that only reads the value in the object, the extra copy created is actually unnecessary....`?

>And what does scale up to larger objects mean?

Oh, phrasing seems abit weird, should instead say `objects of larger size`?

contents/c++/rvalue.md
yamidark: >Maybe it's better to explain what the container classes here refer to.

Perhaps I can give a few examples here `(e.g. vector, list, set)`?

contents/c++/rvalue.md
yamidark: >This sentence is rather hard to understand as well.

Whoops, quite a few errors in this 1 sentence alone (also couldn't understand what I was writing) xD.

contents/c++/rvalue.md
vivekscl: >Hmm, maybe we can move this to the end of the paragraph, something like:
By default, C++ uses this mode if we declare variables with only the data type ?

Not sure if `C++ uses this mode` sounds appropriate. Is `Value Semantics` a mode? If it is then it should be fine :+1: 

>Do you mean to mention memory address instead? Because I think its easier to understand and compare the concept to OOP if we mention objects.

That is true. So how does this sound instead?
`rather than the object reference`?

contents/c++/rvalue.md
vivekscl: >Hmm, maybe I should add If we pass the value to a function that only reads the value in the object, the extra copy created is actually unnecessary....?

That's much clearer :smile: 

>Oh, phrasing seems abit weird, should instead say objects of larger size?

Sounds good.

contents/c++/rvalue.md
vivekscl: Yup, that would be better :+1:

yamidark:
@vivekscl Thanks for the thorough review! I have made the changes requested, do let me know if any parts are still not clear.

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+    * [Reference Semantics](#reference-semantics)
+* [Move Semantics](#move-semantics)
+    * [rvalue and lvalue references](#rvalue-and-lvalue-references)
+    * [Using rvalue references](#using-rvalue-references)
+    * [Move Semantics](#move-semantics)
+* [Rvalue anti-pattern](#rvalue-anti-pattern)
+* [Resources](#resources)
+
+## Background
+*C++* is a general-purpose programming language designed to provide high performance and efficiency for resource-constrained and large systems. The language has been extended and improved after the past decades, with new standards being released periodically. One such standard, the [`C++11`](https://isocpp.org/wiki/faq/cpp11) standard, brought about many changes and improvements, that it made C++ feel like a different language altogether. One such feature that furthur improves on the performance of the language is the introduction of *rvalue* references and *Move Semantics*.
```
vivekscl: This paragraph could be more concise like so:
`The language has since been extended and improved upon, with new standards being released periodically. One such standard, the C++11 standard,  improved the performance of the language with features such as rvalue references and Move Semantics.`

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+
+Some of the advantages of Value Semantics include:
+* No memory management issue, because there won't be any [dangling references](https://www.quora.com/What-is-dangling-reference) to objects that may not exist, nor any [memory leaks](https://www.geeksforgeeks.org/what-is-memory-leak-how-can-we-avoid/).
+* Ensures [referential transparency](https://en.wikipedia.org/wiki/Referential_transparency). Having our own copy of the object ensures changing the values inside one object will not affect the original object. This is especially useful in a [multi-threaded](https://stackoverflow.com/questions/1313062/what-is-a-multithreaded-application) environment, as it removes the need for synchronization of the object's values, allowing the program to run faster.
```
vivekscl: ~~Ensures~~ **Referential Transparency**. Having our own copy of the object ensures **that** changing the values of **the copied object** will not affect the original object.

...as it removes the need **to synchronise** the object's values, allowing the program to run faster.

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+
+In order to better understand the benefits of using Move Semantics, it is important to first understand the two other modes already available in C++, *Value Semantics* and *References Semantics*.
+
+### Value Semantics
+Value (or copy) semantics is the programming style where users are only concerned about the values stored in the objects, rather than the object itself. As such, an extra copy of the object will always be created whenever it is passed to a function, (also known as [pass-by-value](http://www.learncpp.com/cpp-tutorial/72-passing-arguments-by-value/)) or when constructing or assigning a new object. This ensures that each object declared (or function) will have their own copied value to use, without having to concern themselves with their originator. By default, C++ uses this mode if variables are declared with only the data type.
+
+Some of the advantages of Value Semantics include:
+* No memory management issue, because there won't be any [dangling references](https://www.quora.com/What-is-dangling-reference) to objects that may not exist, nor any [memory leaks](https://www.geeksforgeeks.org/what-is-memory-leak-how-can-we-avoid/).
```
vivekscl: **Less memory management issues, as there...**

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+* Ensures [referential transparency](https://en.wikipedia.org/wiki/Referential_transparency). Having our own copy of the object ensures changing the values inside one object will not affect the original object. This is especially useful in a [multi-threaded](https://stackoverflow.com/questions/1313062/what-is-a-multithreaded-application) environment, as it removes the need for synchronization of the object's values, allowing the program to run faster.
+* Speedup. If the function requires accessing of the value many times, having your own local copy of the object may be faster than having a pointer to the object and dereferencing it each time. This is especially true in C++ which encourages Value Semantics, with optimization techniques such as [copy elision](http://en.cppreference.com/w/cpp/language/copy_elision) in the compiler to help passing-by-value be faster.
```
vivekscl: **Better performance**. If the function requires accessing of the value many times, having **a** local copy

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+
+However, Value Semantics has [one major flaw](https://www.quora.com/What-are-the-drawbacks-of-pass-by-value-result):
+* Poor performance and scalability. If the function called is a read-only function, creating an additional copy will be an unnecessary consumption of memory and runtime. This is especially true when scaling up to pass around objects of larger size, as more memory and runtime is needed to create each copy.
+
+### Reference Semantics
+Reference (or pointer) semantics is another choice available for use in C++. C++ allows users to declare [pointers](http://www.cplusplus.com/doc/tutorial/pointers/) and [references](https://www.geeksforgeeks.org/references-in-c/) that point to the address of the object. As such, we can pass around these reference address to functions and declaration of other pointers, and all of them will refer to and use the same object and address.
```
vivekscl: Reference (or pointer) semantics is another choice available for use in C++ **which** allows...

As such, we can pass **these** to functions...

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+
+Some advantages of Reference Semantics include:
+* Improve performance and scalability. By passing around pointers and references to the same object, functions can just use the values in that object directly without having to create an extra copy, thus overcoming the major flaw of Value Semantics.
```
vivekscl: Improve`d`

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+* Allows interaction between objects and functions. Since the same object can be referred to by the different pointers, they can modify the same address values in different areas.
```
vivekscl: ~~Allows~~ Interaction`s` between objects and functions

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+
+However, Reference Semantics also comes with its own problems, which include:
+* Indeterminant behaviour in a multi-threaded environment. Since each thread will all be pointing to the same object, this will lead to unnecessary data races and indeterminant behaviour as these same objects are modified concurrently in each thread. This in turn requires extra work to synchronize the values between objects, which also slows down the program. More details of this can be found in the resources [Java Concurrency](java/JavaConcurrency.md) and [Java Synchronization](java/JavaSynchronization.md).
```
vivekscl: this will lead to unnecessary data races and indeterminant behaviour as **they** are...

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Reference Semantics
+Reference (or pointer) semantics is another choice available for use in C++. C++ allows users to declare [pointers](http://www.cplusplus.com/doc/tutorial/pointers/) and [references](https://www.geeksforgeeks.org/references-in-c/) that point to the address of the object. As such, we can pass around these reference address to functions and declaration of other pointers, and all of them will refer to and use the same object and address.
+
+Some advantages of Reference Semantics include:
+* Improve performance and scalability. By passing around pointers and references to the same object, functions can just use the values in that object directly without having to create an extra copy, thus overcoming the major flaw of Value Semantics.
+* Allows interaction between objects and functions. Since the same object can be referred to by the different pointers, they can modify the same address values in different areas.
+
+However, Reference Semantics also comes with its own problems, which include:
```
vivekscl: However, Reference Semantics also comes with its own problems, **such as**:

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+* Indeterminant behaviour in a multi-threaded environment. Since each thread will all be pointing to the same object, this will lead to unnecessary data races and indeterminant behaviour as these same objects are modified concurrently in each thread. This in turn requires extra work to synchronize the values between objects, which also slows down the program. More details of this can be found in the resources [Java Concurrency](java/JavaConcurrency.md) and [Java Synchronization](java/JavaSynchronization.md).
+* Unintended behaviour. If users are not careful, they can modify the values in the object using one pointer or reference, which will then be reflected by another pointer to the same object, which the user may not have intended. A good example of this problem can be found in this [article](http://www.drdobbs.com/cpp/optimization-calling-by-value-or-by-refe/232400151).
```
vivekscl: Doesn't this point overlap with the `Indeterminant behaviour`?

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Move Semantics
+From the summary above, there are many benefits for using Value Semantics. However, it's one major flaw is that a copy of the object will always be created, which can computationally expensive if this object is of a large size. However, Reference Semantics may also not be preferred due to its different problems as discussed above. As such, in order to continue gaining the benefits of Value Semantics while overcoming it's major flaw, `C++11` introduced a new mode to users, Move Semantics.
```
vivekscl: This paragraph has sentences that are fragmented. 

For instance, you briefly mention that Reference Semantics has flaws and then move on to say Move Semantics mitigates the flaws of Value Semantics without linking it to Reference Semantics. 

So i'm not sure what this paragraph is about. Are you trying to say the Move Semantics mitigate the flaws of both Value and Reference Semantics or just Value Semantics?

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+
+### rvalue and lvalue references
+To understand how Move Semantics work in C++, it is important to distinguish between rvalue and lvalue references.
+```
+lvalue = rvalue
+```
+From the line above, lvalue (left value) basically refers to values that are addressable, whileas rvalue (right value) are temporary objects or values which are used only on the right side of an assignment expression. More details and classification of these 2 values can be found [here](http://www.bogotobogo.com/cplusplus/C11/4_C11_Rvalue_Lvalue.php).
```
vivekscl: basically refers to values that are addressable, `while` rvalue

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+}
+
+int main() {
+    int x = 5;
+    print(x); // lvalue method used
+    print(10); // rvalue method used.
+}
+```
+
+Using of these functions overload with rvalue parameters, which take on temporary objects, helps in writing more efficient program using Move Semantics!
```
vivekscl: Can be more concise here. Also, there are some typos here as well. Could re-written as:
**Usage** of function overloading with rvalue parameters, help in writing more efficient program`s` using Move Semantics!

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Move Semantics
+The main usage of rvalue references is that it allows us to create *move* constructor and *move* assignments, instead of *copy* constructor and *copy* assignments by default. Since rvalues are typically temporary objects, we can just *move* the value instead of *copying* it, thus reducing on memory consumption and improving on the performance!
```
vivekscl: reducing ~~on~~ memory consumption and improving ~~on the~~ performance!

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+    _x = x;
+    _y = y;
+}
+
+Foo(std::string&& x, std::string&& y) { // move constructor 3
+    _x = x;
+    _y = y;
+}
+```
+However, this is not feasible as we would require `2^n` overload functions, where `n` is the number of parameters. This results in large amount of [boilerplate code](https://www.quora.com/What-is-boilerplate-code), which in turn reduces code quality, and also contributes to increased memory consumption and compilation time.
```
vivekscl: results in **increased** boilerplate code, which in turn reduces code quality, and ~~also contributes to~~ increases memory consumption and compilation time.

vivekscl (changes requested):
Overall the content is much clearer now. What's left is to improve the language and phrasing a bit more and it should be fine :+1:

contents/c++/rvalue.md
yamidark: Hmm, does sound so...
Guess I will remove this section then :P

contents/c++/rvalue.md
yamidark: Hmmm, I'm kind of trying to say that we might want to continue using Value Semantics because of its benefits, and Move Semantics help to mitigate its major flaw.
Perhaps I should not even mention Reference Semantics here?

contents/c++/rvalue.md
yamidark: Actually, `indeterminant` behaviour is different from `unintended`, since `indeterminant` means we can't determine what the result until we run the program, and it may produce different results each time.
But even if the result is `unintended`, in a sequential program the result should be deterministic.

contents/c++/rvalue.md
yamidark: Would this sound better:
`As such, in order to continue gaining the benefits Value Semantics have over Reference Semantics while overcoming it's major flaw, `C++11` introduced a new mode to users, Move Semantics.`

vivekscl: Yup, this sounds better

contents/c++/rvalue.md
vivekscl: Ohh, then is it better to say `Unintended side-effects`?

yamidark: I'm thinking of writing this as `Harder to debug`, since if the behaviour is unintended, users will have a harder time trying to spot where the side-effected come from, especially if it is changed in distant place in the program?

yamidark: Is this better?
`Hardy to debug. If users are not careful, they may modify the values in the object using one pointer, which will then be reflected by another pointer to the same object in a different part of the program, which can confuse users as they may not know where and why the values are changing.`

vivekscl: Sounds good to me :+1:

yamidark:
@vivekscl I have made the changes requested. Could you help me review this again?

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+
+In order to better understand the benefits of using Move Semantics, it is important to first understand the two other modes already available in C++, *Value Semantics* and *References Semantics*.
+
+### Value Semantics
+Value (or copy) semantics is the programming style where users are only concerned about the values stored in the objects, rather than the object itself. As such, an extra copy of the object will always be created whenever it is passed to a function, (also known as [pass-by-value](http://www.learncpp.com/cpp-tutorial/72-passing-arguments-by-value/)) or when constructing or assigning a new object. This ensures that each object declared (or function) will have their own copied value to use, without having to concern themselves with their originator. By default, C++ uses this mode if variables are declared with only the data type.
+
+Some of the advantages of Value Semantics include:
```
vivekscl: Some ~~of the~~ advantages of Value Semantics include

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+* Lesser memory management issue, as there won't be any [dangling references](https://www.quora.com/What-is-dangling-reference) to objects that may not exist, nor any [memory leaks](https://www.geeksforgeeks.org/what-is-memory-leak-how-can-we-avoid/).
```
vivekscl: Lesser memory management issue`s`

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+* [Referential transparency](https://en.wikipedia.org/wiki/Referential_transparency). Having our own copy of the object ensures that changing the values inside the copied object will not affect the original object. This is especially useful in a [multi-threaded](https://stackoverflow.com/questions/1313062/what-is-a-multithreaded-application) environment, as it removes the need to synchronise the object's values, allowing the program to run faster.
+* Better performance. If the function requires accessing of the object's value many times, having a local copy of the object may be faster than having a pointer to the object and dereferencing it each time. This is especially true in C++ which encourages Value Semantics, with optimization techniques such as [copy elision](http://en.cppreference.com/w/cpp/language/copy_elision) in the compiler to help passing-by-value be faster.
```
vivekscl: in the compiler **which speeds up the process of passing-by-value.**

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Reference Semantics
+Reference (or pointer) semantics is another choice available for use in C++, which allows users to declare [pointers](http://www.cplusplus.com/doc/tutorial/pointers/) and [references](https://www.geeksforgeeks.org/references-in-c/) that point to the address of the object. As such, we can pass these reference address to functions and declaration of other pointers, and all of them will refer to and use the same object and address.
+
+Some advantages of Reference Semantics include:
+* Improved performance and scalability. By passing around pointers and references to the same object, functions can just use the values in that object directly without having to create an extra copy, thus overcoming the major flaw of Value Semantics.
+* Interactions between objects and functions. Since the same object can be referred to by the different pointers, they can modify the same address values in different areas of the program.
```
vivekscl: Since the same object can be referred to by ~~the~~ different pointers, **its value can be modified** in different areas of the program.

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+
+However, Reference Semantics also comes with its own problems, such as:
+* Indeterminant behaviour. In a multi-threaded environment, since each thread will all be pointing to the same object, this will lead to unnecessary data races and indeterminant behaviour as they are modified concurrently in each thread. This in turn requires extra work to synchronize the values between objects, which also slows down the program. More details of this can be found in the resources [Java Concurrency](java/JavaConcurrency.md) and [Java Synchronization](java/JavaSynchronization.md).
+* Hardy to debug. If users are not careful, they may modify the values in the object using one pointer, which will then be reflected by another pointer to the same object in a different part of the program, which can confuse users as they may not know where and why the values are changing. A good example of this problem can be found in this [article](http://www.drdobbs.com/cpp/optimization-calling-by-value-or-by-refe/232400151).
```
vivekscl: Hard~~y~~ to debug

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Move Semantics
+From the summary above, there are many benefits for using Value Semantics. However, it's one major flaw is that a copy of the object will always be created, which can computationally expensive if this object is of a large size. Reference Semantics may also not be preferred due to its different problems as discussed above. As such, in order to continue gaining the benefits Value Semantics have over Reference Semantics while overcoming it's major flaw, `C++11` introduced a new mode to users, Move Semantics.
```
vivekscl: However, it's one major flaw is that a copy of the object will always be created, which can **be** computationally expensive...

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+    int x = 5;
+    print(x); // lvalue method used
+    print(10); // rvalue method used.
+}
+```
+
+Usage of function overloading with rvalue parameters, which take on temporary objects, helps in writing more efficient programs using Move Semantics!
+
+### Move Semantics
+The main usage of rvalue references is that it allows us to create *move* constructor and *move* assignments, instead of *copy* constructor and *copy* assignments by default. Since rvalues are typically temporary objects, we can just *move* the value instead of *copying* it, thus reducing  memory consumption and improving on performance!
```
vivekscl: and improving ~~on~~ performance!

vivekscl (approved):
Just some minor nits. After this you can request a review from prof :+1:

yamidark:
@vivekscl Thanks for helping me in reviewing this!
@damithc Ready for your review.

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+    * [Using rvalue references](#using-rvalue-references)
+    * [Move Semantics](#move-semantics)
+* [rvalue anti-pattern](#rvalue-anti-pattern)
+* [Resources](#resources)
+
+## Background
+*C++* is a general-purpose programming language designed to provide high performance and efficiency for resource-constrained and large systems. The language has since been extended and improved upon, with new standards being released periodically. One such standard, the C++11 standard, improved the performance of the language with features such as *rvalue references* and *Move Semantics*.
+
+In order to better understand the benefits of using Move Semantics, it is important to first understand the two other modes already available in C++, *Value Semantics* and *References Semantics*.
+
```
damithc: > ... as *rvalue references* and *Move Semantics*.

> ... available in C++, *Value Semantics* and *References Semantics*.

Of the 4 terms mentioned, any reason why the first one is all lower case?

Also, try to use `_italic_` and `**bold**` so that it is easy to see the difference between the two.

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+Some advantages of Reference Semantics include:
+* Improved performance and scalability. By passing around pointers and references to the same object, functions can just use the values in that object directly without having to create an extra copy, thus overcoming the major flaw of Value Semantics.
+* Interactions between objects and functions. Since the same object can be referred to by different pointers, its value can be modified in different areas of the program.
+
+However, Reference Semantics also comes with its own problems, such as:
+* Indeterminant behaviour. In a multi-threaded environment, since each thread will all be pointing to the same object, this will lead to unnecessary data races and indeterminant behaviour as they are modified concurrently in each thread. This in turn requires extra work to synchronize the values between objects, which also slows down the program. More details of this can be found in the resources [Java Concurrency](java/JavaConcurrency.md) and [Java Synchronization](java/JavaSynchronization.md).
```
damithc: unexplained term: _data race_

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+* Harder to debug. If users are not careful, they may modify the values in the object using one pointer, which will then be reflected by another pointer to the same object in a different part of the program, which can confuse users as they may not know where and why the values are changing. A good example of this problem can be found in this [article](http://www.drdobbs.com/cpp/optimization-calling-by-value-or-by-refe/232400151).
+
+## Move Semantics
+From the summary above, there are many benefits for using Value Semantics. However, it's one major flaw is that a copy of the object will always be created, which can be computationally expensive if this object is of a large size. Reference Semantics may also not be preferred due to its different problems as discussed above. As such, in order to continue gaining the benefits Value Semantics have over Reference Semantics while overcoming it's major flaw, `C++11` introduced a new mode to users, Move Semantics.
```
damithc: `it's one` --> `its one` ?

contents/c++/rvalue.md
```diff
@@ -0,0 +1,169 @@
...
+
+### rvalue and lvalue references
```
damithc: `reference` -> `Reference`?
Check other headings as well.

damithc (changes requested):
Good work 👍 
Just a few cosmetic nits

yamidark:
@damithc Thanks for the feedback! Made the changes requested, and ready for another round of review.

contents/c++/rvalue.md
```diff
@@ -0,0 +1,170 @@
...
+    z = std::move(other.z);
+    return *this;
+}
+```
+
+In this move constructor and assignment, the contents of the `other` parameter is *moved* into the object, and the contents in `other` is destroyed afterwards. No additional memory allocation is required, and the *move* operation is done quickly by a few assignment of address operations, leading to a faster and more memory efficient program.
+
+Along with Move Semantics in `C++11`, the [STL library](https://www.geeksforgeeks.org/the-c-standard-template-library-stl/) provides the overloaded *move* functions for its container classes (e.g. `vector`, `list`, `set`), thus we can take advantage of these Move Semantics by simply supplying rvalues, without the need to redefine the classes ourselves.
+
+## Rvalue anti-pattern
```
damithc: `Anti-Pattern`?

contents/c++/rvalue.md
```diff
@@ -0,0 +1,170 @@
+# Rvalue References and Move Semantics
+
+Authors: [Tan Jun An](https://github.com/yamidark)
+
+* [Background](#background)
+    * [Value Semantics](#value-semantics)
+    * [Reference Semantics](#reference-semantics)
```
damithc: Indentation here should be like this (not 4 spaces):
```
* [Background](#background)
  * [Value Semantics](#value-semantics)
```

damithc (changes requested):
Couple of nits I spotted upon a quick scan.

yamidark:
@damithc Made the changes requested, and ready for another round of review.

damithc (approved):
LGTM 👍

PR #62, opened 2018-07-28
Fix styling

pyokagan:


pyokagan:
@damithc Heh nice, I see Netlify has already been set up.

damithc (approved):
Nice 👍 Thanks for doing this.
Hope it wasn't a lot of work to figure out the MarkBind stuff. We haven't optimized the author-friendliness aspect to our target levels yet.

damithc:
@pyokagan do you want to keep the commits or shall I squash?

pyokagan:
@damithc 

Nah, took about 1 hour tops. I copied most of the stuff from se-book/website-base.

Keep the commits please.

pyokagan:
btw, I needed to do some workarounds with strange MarkBind defaults in https://github.com/se-edu/learningresources/pull/62/commits/d8f206f04b5565dba134066dbf379109bc4683da . cc @yamgent

damithc:
Deployed.

pyokagan:
> Hope it wasn't a lot of work to figure out the MarkBind stuff. We haven't optimized the author-friendliness aspect to our target levels yet.

Oh, I realized I probably should have left some usability feedback for MarkBind:

* I found MarkBind constantly clearing the console kind of annoying, esp for non-daemon usage like `markbind build`.
* I locally installed MarkBind (i.e. in the `node_modules` directory). MarkBind tried to convert the Markdown documents inside the `node_modules` directory. I tried to ignore it by adding `node_modules/*` to the `site.json`, but it didn't seem to work? In the end I hacked my way out by doing the following: (Note, I didn't read the user guide)
    ```diff
    commit f1f32e4a2b4c4793af8707476e853b6b8562fbd2
    Author: Paul Tan <pyokagan@pyokagan.name>
    Date:   Sat Jul 28 13:55:21 2018 +0800

        TMP

    diff --git a/site.json b/site.json
    index fcbb30b..43ea59e 100644
    --- a/site.json
    +++ b/site.json
    @@ -5,11 +5,21 @@
         "_markbind/logs/*",
         "_site/*",
         "site.json",
    -    "*.md"
    +    "*.md",
    +    "node_modules/*"
       ],
       "pages": [
         {
    -      "glob": "**/*.md"
    +      "glob": "contents/**/*.md"
    +    },
    +    {
    +        "glob": "index.md"
    +    },
    +    {
    +        "glob": "GuidelinesForContributors.md"
    +    },
    +    {
    +        "glob": "Contact.md"
         }
       ],
       "deploy": {
    ```

* `node` printed the following on `markbind serve`, which is quite scary :-P 
    ```
    (node:25288) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 error listeners added. Use emitter.setMaxListeners() to increase limit
    ```
* The `navbar` variable is used in almost all the markdown documents. However, when I modified the navbar in `_markbind/variables.md`, none of the documents got rebuilt. In the end I just resorted to Ctrl-C-ing `markbind serve` and restarting it each time.
* A progress indicator for `markbind build` would be nice.

pyokagan:
Also feels kind of obtuse to be adding `head: head.md` to every document as in https://github.com/se-edu/learningresources/pull/62/commits/90f6492143925fe799428f52684c7e94ce780a05 , can it be set in the `site.json`?

damithc:
Thanks for the comments @pyokagan Good points.
Hope to fix them in a future release (focus for this summer was reader-friendliness. Hope to switch focus to author-friendliness in the coming semester). Your comments will be useful at that time.

@yamgent Please create issues as necessary, when you have time.

yamgent:
> I found MarkBind constantly clearing the console kind of annoying, esp for non-daemon usage like markbind build.

Yes, I agree (I use the terminal constantly, and the clearing of console just seems to be counter-intuitive). Does the big logo printed at the beginning annoys you too as well? :P

> I locally installed MarkBind (i.e. in the node_modules directory). MarkBind tried to convert the Markdown documents inside the node_modules directory. I tried to ignore it by adding node_modules/* to the site.json, but it didn't seem to work? In the end I hacked my way out by doing the following: (Note, I didn't read the user guide)

Not sure what the "hack" here is, because the change that you made is actually correct.

> node printed the following on markbind serve, which is quite scary :-P
> ```
> (node:25288) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 error listeners added. Use emitter.setMaxListeners() to increase limit
> ```

This looks quite familiar, will look into it. Does it always happen for you?

> The navbar variable is used in almost all the markdown documents. However, when I modified the navbar in _markbind/variables.md, none of the documents got rebuilt. In the end I just resorted to Ctrl-C-ing markbind serve and restarting it each time.
>
> A progress indicator for markbind build would be nice.

👌 

> Also feels kind of obtuse to be adding head: head.md to every document as in 90f6492 , can it be set in the site.json?

We will be reconsidering the design of this in the coming semester (`footer` and `site-nav` currently suffers the same problem).

damithc:
> node printed the following on markbind serve, which is quite scary :-P

Happened to me several times. In my case it was always caused by a mistake in my nunjucks code e.g., an array not closed properly.

pyokagan:
@yamgent 
> Not sure what the "hack" here is, because the change that you made is actually correct.

I was surprised that the `ignore` configuration did not affect `pages`.

(Looking at the User Guide now, I see that there is no way to exclude files from a `pages` glob?)

pyokagan:
> This looks quite familiar, will look into it. Does it always happen for you?

Yeah it always happens on this repo, even from the start. Dunno if its different depending on node version, I'm using `v9.11.2`.

yamgent:
> I was surprised that the `ignore` configuration did not affect `pages`.

It is not supposed to. `pages` affects what files are getting rendered from markdown to html. `ignore` affects what files are being copied over to the final built site.

I know it can get quite confusing. I also found it confusing when I first started with the MarkBind project, but apparently no one else was disturbed by this problem, so... ¯\\\_(ツ)\_/¯

> (Looking at the User Guide now, I see that there is no way to exclude files from a `pages` glob?)

Nope. This problem is raised in https://github.com/MarkBind/markbind/issues/228.

PR #71, opened 2019-02-09
Add introduction to React

### LINKED ISSUE #66: Introduction to React
Outline:
- Overview
- Advantages
- Disadvantages
- Getting started
- Other resources for learning

amad-person:
Added an overview, pros and cons, and resources to get started with React.
Preview it [here](https://deploy-preview-71--learningresources.netlify.com/contents/javascript/javascript-framework-react).

Fixes #66.


contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,115 @@
...
+#### React is Declarative
+
+With React's <tooltip content="Declarative programming focuses on what the program should accomplish without specifying how the program should achieve the result.">declarative</tooltip> approach to building user interfaces, you can painlessly create interaction-rich applications. 
+You can build these interfaces without touching the DOM directly, and even have an event system that doesn't need you to interact with actual DOM events. 
+
+This is unlike the traditional <tooltip content="Imperative programming focuses on explicitly describing how a program operates.">imperative</tooltip> approach of building applications. An example would be looking up elements in the DOM using jQuery, where you have to tell the browser what to do and how to update the elements whenever data changes.  
+
+You can see React's declarative approach in action in the sandbox below:
+
+<iframe src="https://codesandbox.io/embed/p564px0j47?fontsize=12" style="width:100%; height:400px; border:0; border-radius: 4px; overflow:hidden;" sandbox="allow-modals allow-forms allow-popups allow-scripts allow-same-origin"></iframe>
```
amad-person: This is an editable CodeSandbox embed (if it is not clear).

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,115 @@
+<frontmatter>
+  title: ReactJs
```
damithc: Should this be React? If not, match with the page heading below.

damithc (commented):
Good start. Some prelim comments:
* Give an unbiased view. e.g., avoid marketing phrases such as `has taken the world of front-end development by storm`
* Provide more info about where React fits into the whole landscape. e.g., other competing alternatives

After that, you can proceed to request a peer review.

contents/javascript/Javascript-framework-React.md
amad-person: Yep, I missed that. I'll change it.

amad-person:
> Give an unbiased view. e.g., avoid marketing phrases such as `has taken the world of front-end development by storm`

Noted! 

> Provide more info about where React fits into the whole landscape. e.g., other competing alternatives

Can I link to articles, or is it better if I add a short section myself?

damithc:
> Can I link to articles, or is it better if I add a short section myself?

It is better to add a brief description and link to more info.

amad-person:
Okay, thank you!

amad-person:
Ready for review.

amad-person:
@amrut-prabhu @RonakLakhotia may I get a review? Thanks!

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,134 @@
...
+
+### What is React?
+
+**React** is an open-source JavaScript <tooltip content="React only handles the view component of your application. It is not an MVC framework!">library</tooltip> for building composable user interfaces. It was developed by Facebook and released to the world in 2013, and is now used in a huge number of popular applications, including Facebook and Instagram.
+
+<box type="tip">
+    If you have the <a href="https://github.com/facebook/react-devtools">React DevTools</a> extension installed in your browser, you can inspect the webpage of either of these applications and observe that a lot of React components are being used by them! 
+</box>
+
+React abstracts away the [DOM](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Introduction) from you, giving you a simpler programming model and better performance. It can also render on the server using Node, and it can power native mobile applications using [React Native](https://facebook.github.io/react-native/).
```
amrut-prabhu: >It can also render on the server using Node

Not exactly clear about what this means 😅

amrut-prabhu: I think it would be good to start a new sentence/paragraph and separate React native from React.

So, you can say something like `A slight variation of React is React Native...used for building native apps for Android and iOS...`
Also, React Native is a framework (vs React, which is a library)

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Why React?
```
amrut-prabhu: I see what you're doing here, but it may be better to say "Why use React?" or "Why `React`"

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,134 @@
...
+
+React allows developers to create large web applications which can change data, without reloading the page. The main purpose of React is to be fast, scalable, and simple.
```
amrut-prabhu: > applications which can change data

Something along these lines instead?
`applications in which the data in specific parts/components? of the page can be changed without the need to reload the entire page`

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,134 @@
...
+React abstracts away the [DOM](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Introduction) from you, giving you a simpler programming model and better performance. It can also render on the server using Node, and it can power native mobile applications using [React Native](https://facebook.github.io/react-native/).
+
+### Why React?
+
+React allows developers to create large web applications which can change data, without reloading the page. The main purpose of React is to be fast, scalable, and simple. 
+Some of its benefits are described in detail below.
+
+#### React is Declarative
+
+With React's <tooltip content="Declarative programming focuses on what the program should accomplish without specifying how the program should achieve the result.">declarative</tooltip> approach to building user interfaces, you can painlessly create interaction-rich applications.
```
amrut-prabhu: `painlessly` sounds like a marketing term 😅

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Why React?
+
+React allows developers to create large web applications which can change data, without reloading the page. The main purpose of React is to be fast, scalable, and simple. 
+Some of its benefits are described in detail below.
+
+#### React is Declarative
+
+With React's <tooltip content="Declarative programming focuses on what the program should accomplish without specifying how the program should achieve the result.">declarative</tooltip> approach to building user interfaces, you can painlessly create interaction-rich applications. 
+You can build these interfaces without touching the DOM directly, and even have an event system that doesn't need you to interact with actual DOM events.
```
amrut-prabhu: without ~touching~ **manipulating** the DOM directly, and even have an event system that doesn't ~need~ **require** you to?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,134 @@
...
+
+React allows developers to create large web applications which can change data, without reloading the page. The main purpose of React is to be fast, scalable, and simple. 
+Some of its benefits are described in detail below.
+
+#### React is Declarative
+
+With React's <tooltip content="Declarative programming focuses on what the program should accomplish without specifying how the program should achieve the result.">declarative</tooltip> approach to building user interfaces, you can painlessly create interaction-rich applications. 
+You can build these interfaces without touching the DOM directly, and even have an event system that doesn't need you to interact with actual DOM events. 
+
+This is unlike the traditional <tooltip content="Imperative programming focuses on explicitly describing how a program operates.">imperative</tooltip> approach of building applications. An example would be looking up elements in the DOM using jQuery, where you have to tell the browser what to do and how to update the elements whenever data changes.
```
amrut-prabhu: Add a link to jQuery?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,134 @@
...
+You can build these interfaces without touching the DOM directly, and even have an event system that doesn't need you to interact with actual DOM events. 
+
+This is unlike the traditional <tooltip content="Imperative programming focuses on explicitly describing how a program operates.">imperative</tooltip> approach of building applications. An example would be looking up elements in the DOM using jQuery, where you have to tell the browser what to do and how to update the elements whenever data changes.  
+
+You can see React's declarative approach in action in the sandbox below:
+
+<iframe src="https://codesandbox.io/embed/p564px0j47?fontsize=12" style="width:100%; height:400px; border:0; border-radius: 4px; overflow:hidden;" sandbox="allow-modals allow-forms allow-popups allow-scripts allow-same-origin"></iframe>
+
+The sandbox above is editable. You can play around with the application `state` and see how the view automatically gets updated.
```
amrut-prabhu: I think it would be good to have this in a `tip` or `info` box instead

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,134 @@
...
+#### React is Declarative
+
+With React's <tooltip content="Declarative programming focuses on what the program should accomplish without specifying how the program should achieve the result.">declarative</tooltip> approach to building user interfaces, you can painlessly create interaction-rich applications. 
+You can build these interfaces without touching the DOM directly, and even have an event system that doesn't need you to interact with actual DOM events. 
+
+This is unlike the traditional <tooltip content="Imperative programming focuses on explicitly describing how a program operates.">imperative</tooltip> approach of building applications. An example would be looking up elements in the DOM using jQuery, where you have to tell the browser what to do and how to update the elements whenever data changes.  
+
+You can see React's declarative approach in action in the sandbox below:
+
+<iframe src="https://codesandbox.io/embed/p564px0j47?fontsize=12" style="width:100%; height:400px; border:0; border-radius: 4px; overflow:hidden;" sandbox="allow-modals allow-forms allow-popups allow-scripts allow-same-origin"></iframe>
```
amrut-prabhu: A few suggestions:
- Add a comment explaining what `App` means/refers to
- >   // ~State is where the data in your application lives.~

   State represents the data of the `App` component
- > // This method specifies what to render as the view.

   render as the HTML that is displayed on the page
- Add a comment in styles.css to explain that `.App` is the `class` specified under `className` in index.js?
- Remove **unnecessary** stuff from index.html and package.json. Only keep the bare minimum required to compile the app. I think you should do this to make things clearer to the reader. You don't expect them to download the example to get started with React anyways, right?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,134 @@
...
+Some of its benefits are described in detail below.
+
+#### React is Declarative
+
+With React's <tooltip content="Declarative programming focuses on what the program should accomplish without specifying how the program should achieve the result.">declarative</tooltip> approach to building user interfaces, you can painlessly create interaction-rich applications. 
+You can build these interfaces without touching the DOM directly, and even have an event system that doesn't need you to interact with actual DOM events. 
+
+This is unlike the traditional <tooltip content="Imperative programming focuses on explicitly describing how a program operates.">imperative</tooltip> approach of building applications. An example would be looking up elements in the DOM using jQuery, where you have to tell the browser what to do and how to update the elements whenever data changes.  
+
+You can see React's declarative approach in action in the sandbox below:
```
amrut-prabhu: in the sandbox (**with comments explaining the code**) below?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,134 @@
...
+
+Any updates to your application will first be made to the virtual DOM. Then, React will compare the virtual DOM with the actual DOM using a [diffing algorithm](https://reactjs.org/docs/reconciliation.html) to find out what changes have to be made to the actual DOM. 
+Finally, React figures out how to update the actual DOM in the most effective way possible to guarantee minimum update time. This results in better performance and a cleaner user experience, even with data-intensive applications.
+
+Those are the three main benefits of React. Other advantages are as follows:
+ - React only allows data to flow downwards (one-way data binding), which makes your application easier to debug.
+ - It is a constantly developing open-source library that is backed by Facebook, so you can be sure of getting community support.
+ - If your application is large, you can manage its state conveniently using [Redux](https://redux.js.org/introduction/getting-started). Note that Redux can be used with any other view library, but it is mostly used with React. Hence, there are more resources for learning React + Redux.
+
+### Why not React?
```
amrut-prabhu: Something like `Why not to use React` instead?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,134 @@
...
+
+Like any other library, React has its disadvantages, which are as follows:
+ - The high pace of development means that you would need to regularly relearn how to do things.
+ - The huge number of tools you can integrate with React don't always have comprehensive documentation.
+ - React is just a UI library. As React only allows one-way data binding, you would need to use [Flux](https://github.com/facebook/flux), a new application architecture introduced by Facebook that favours unidirectional data flow. 
+ However, if you prefer working with the MVC pattern, you should consider a framework that lets you use it.
+
+#### React and other Competing Alternatives
+
+There are a lot of JavaScript frameworks and libraries that you can use to build your next web application. Some popular alternatives to React are [Angular](https://angular.io/) and [Vue](https://vuejs.org/).
```
amrut-prabhu: Link to the corresponding book chapters instead of external websites?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,134 @@
...
+
+How do you decide which one to use? Here are some resources to help you choose between them:
+- [React, Angular, Vue: What they can do and which one is for you](https://blog.teamtreehouse.com/react-angular-vue) - guidelines for choosing which technology to learn.
+- [Angular vs Vue vs React](https://www.codeinwp.com/blog/angular-vs-vue-vs-react/) - in addition to comparing the three technologies, this article aims to give a general structure for comparing JavaScript frameworks and libraries. Hence, you can use this structure to choose between any new frameworks that may arrive in the future.
+- [State of JS 2018: Front-end Frameworks](https://2018.stateofjs.com/front-end-frameworks/overview/) - survey comparing the average salaries, company size, developer satisfaction, etc. for the most used JavaScript front-end technologies as of 2018.
+
+React is best suited for projects that have a lot of user interactions that manipulate the DOM extensively.
```
amrut-prabhu: Bold `user interactions`?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,134 @@
...
+
+How do you decide which one to use? Here are some resources to help you choose between them:
+- [React, Angular, Vue: What they can do and which one is for you](https://blog.teamtreehouse.com/react-angular-vue) - guidelines for choosing which technology to learn.
+- [Angular vs Vue vs React](https://www.codeinwp.com/blog/angular-vs-vue-vs-react/) - in addition to comparing the three technologies, this article aims to give a general structure for comparing JavaScript frameworks and libraries. Hence, you can use this structure to choose between any new frameworks that may arrive in the future.
+- [State of JS 2018: Front-end Frameworks](https://2018.stateofjs.com/front-end-frameworks/overview/) - survey comparing the average salaries, company size, developer satisfaction, etc. for the most used JavaScript front-end technologies as of 2018.
+
+React is best suited for projects that have a lot of user interactions that manipulate the DOM extensively. 
+You should choose React if you want the flexibility to choose your application's stack, as it is not opinionated.
```
amrut-prabhu: Add a tooltip explaining what `opinionated` means

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,134 @@
...
+ - It is a constantly developing open-source library that is backed by Facebook, so you can be sure of getting community support.
+ - If your application is large, you can manage its state conveniently using [Redux](https://redux.js.org/introduction/getting-started). Note that Redux can be used with any other view library, but it is mostly used with React. Hence, there are more resources for learning React + Redux.
+
+### Why not React?
+
+Like any other library, React has its disadvantages, which are as follows:
+ - The high pace of development means that you would need to regularly relearn how to do things.
+ - The huge number of tools you can integrate with React don't always have comprehensive documentation.
+ - React is just a UI library. As React only allows one-way data binding, you would need to use [Flux](https://github.com/facebook/flux), a new application architecture introduced by Facebook that favours unidirectional data flow. 
+ However, if you prefer working with the MVC pattern, you should consider a framework that lets you use it.
```
amrut-prabhu: > consider a framework that lets you use it

Can you link to any examples?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,134 @@
...
+
+With components written in JavaScript, you can easily pass data through your application and keep state out of the DOM. You can read more about components and props in the React [documentation](https://reactjs.org/docs/components-and-props.html). 
+
+#### React is Fast
+
+Web applications can have a lot of user interaction and data updates. Extensive DOM manipulation can be a performance bottleneck in these applications as the DOM is tree-structured - simple changes at the top can trigger the entire application to get updated. 
+
+React solves this problem by using a virtual DOM. The virtual DOM is a JavaScript object that is kept in the memory of your application. In other words, it is a virtual representation of the actual DOM.
+
+Any updates to your application will first be made to the virtual DOM. Then, React will compare the virtual DOM with the actual DOM using a [diffing algorithm](https://reactjs.org/docs/reconciliation.html) to find out what changes have to be made to the actual DOM.
```
amrut-prabhu: Use this link instead?
https://reactjs.org/docs/reconciliation.html#the-diffing-algorithm

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,134 @@
...
+#### React is Fast
+
+Web applications can have a lot of user interaction and data updates. Extensive DOM manipulation can be a performance bottleneck in these applications as the DOM is tree-structured - simple changes at the top can trigger the entire application to get updated. 
+
+React solves this problem by using a virtual DOM. The virtual DOM is a JavaScript object that is kept in the memory of your application. In other words, it is a virtual representation of the actual DOM.
+
+Any updates to your application will first be made to the virtual DOM. Then, React will compare the virtual DOM with the actual DOM using a [diffing algorithm](https://reactjs.org/docs/reconciliation.html) to find out what changes have to be made to the actual DOM. 
+Finally, React figures out how to update the actual DOM in the most effective way possible to guarantee minimum update time. This results in better performance and a cleaner user experience, even with data-intensive applications.
+
+Those are the three main benefits of React. Other advantages are as follows:
```
amrut-prabhu: Would it be better to have another subheading `Other advantages/benefits` for these?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,134 @@
...
+ - React only allows data to flow downwards (one-way data binding), which makes your application easier to debug.
+ - It is a constantly developing open-source library that is backed by Facebook, so you can be sure of getting community support.
+ - If your application is large, you can manage its state conveniently using [Redux](https://redux.js.org/introduction/getting-started). Note that Redux can be used with any other view library, but it is mostly used with React. Hence, there are more resources for learning React + Redux.
```
amrut-prabhu: > Hence, there are more resources for learning React + Redux.

I don't think this is an advantage of React per se since there are other alternatives if you want to use other frameworks. If you want to use Redux however, React may be a better choice due to reason you mentioned

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,134 @@
...
+
+React solves this problem by using a virtual DOM. The virtual DOM is a JavaScript object that is kept in the memory of your application. In other words, it is a virtual representation of the actual DOM.
+
+Any updates to your application will first be made to the virtual DOM. Then, React will compare the virtual DOM with the actual DOM using a [diffing algorithm](https://reactjs.org/docs/reconciliation.html) to find out what changes have to be made to the actual DOM. 
+Finally, React figures out how to update the actual DOM in the most effective way possible to guarantee minimum update time. This results in better performance and a cleaner user experience, even with data-intensive applications.
+
+Those are the three main benefits of React. Other advantages are as follows:
+ - React only allows data to flow downwards (one-way data binding), which makes your application easier to debug.
+ - It is a constantly developing open-source library that is backed by Facebook, so you can be sure of getting community support.
```
amrut-prabhu: I think you can add something like `Since it is a constantly developing open-source library that is backed by Facebook, it is constantly being updated with new features/performance optimisations?`.
Otherwise, the first statement doesn't add any value. In fact, it is kinda mentioned as a disadvantage below

amrut-prabhu (changes requested):
Good work! The chapter seems very useful.

Here are some comments and suggestions

contents/javascript/Javascript-framework-React.md
amad-person: Server-side rendering. I'll add in a link that explains it, it would be clearer then.

contents/javascript/Javascript-framework-React.md
amad-person: Agreed, I'll change it.

contents/javascript/Javascript-framework-React.md
amad-person: That's a good point, I'll put 

> The sandbox above is editable. You can play around with the application `state` and see how the view automatically gets updated. 

in an `info` box.

contents/javascript/Javascript-framework-React.md
amad-person: > Add a comment explaining what App means/refers to

Do you mean I should mention that it is the main view - where everything in the web application will be rendered?

> State represents the data of the `App` component

I haven't explained what a `component` is in the article yet, so I thought I shouldn't use the word without defining it first. What do you think?

> // render as the HTML that is displayed on the page

Noted, I'll make the comment clearer.

> Add a comment in styles.css to explain that `.App` is the class specified under className in index.js?

Okay!

> Remove unnecessary stuff from `index.html` and `package.json`. 

Good point, I'll do that.

contents/javascript/Javascript-framework-React.md
amad-person: > in the sandbox **(with comments explaining the code)** below

Not sure how useful adding this will be, because the reader will be able to see that for themselves? Comments are used to explain the code, after all.

contents/javascript/Javascript-framework-React.md
amad-person: As of now, the book chapter on Angular hasn't been added yet. 
I could link to the chapter on `Vue` though.

contents/javascript/Javascript-framework-React.md
amad-person: Good catch, thanks!

contents/javascript/Javascript-framework-React.md
amad-person: You're right, I'll update the statement to `constantly being updated with new features/performance optimisations`.

contents/javascript/Javascript-framework-React.md
amad-person: > I think it would be good to start a new sentence/paragraph and separate React native from React.

Although there is merit in elaborating on React Native, I'm wondering if this would take the focus away from the main topic (React).

amad-person:
@amrut-prabhu thanks for your comprehensive review! I've addressed your comments, please take a look if the updates are fine.

contents/javascript/Javascript-framework-React.md
amrut-prabhu: Fair enough, you can either mention it as an aside (in an info box) or you can leave the structure the way it is. However, I think you should at least mention something along the lines of `... using React Native, which is an extension of React.`

It's your call

contents/javascript/Javascript-framework-React.md
amrut-prabhu: > Do you mean I should mention that it is the main view - where everything in the web application will be rendered?

Yep, something like that. See what I've written below ↓ as well

> I haven't explained what a component is in the article yet, so I thought I shouldn't use the word without defining it first. What do you think?

You've actually name-dropped it in the intro tip box and line 35 of this sandbox. Maybe, you can mention it for `App` as well, for completeness and direct them to look at the `React is Component-Based` section for details?

I guess it's fine if you don't too, since it is mentioned in the next sandbox

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,140 @@
...
+
+Any updates to your application will first be made to the virtual DOM. Then, React will compare the virtual DOM with the actual DOM using a [diffing algorithm](https://reactjs.org/docs/reconciliation.html#the-diffing-algorithm) to find out what changes have to be made to the actual DOM. 
+Finally, React figures out how to update the actual DOM in the most effective way possible to guarantee minimum update time. This results in better performance and a cleaner user experience, even with data-intensive applications.
+
+#### Other Advantages of React
+
+Besides the three main benefits explained above, React has the following advantages:
+ - React only allows data to flow downwards (one-way data binding), which makes your application easier to debug.
+ - As it is an open-source library backed by Facebook, it is constantly being updated with new features and performance optimizations.
+ - If you want to manage the state of your application using [Redux](https://redux.js.org/introduction/getting-started), React might be a better choice than other libraries / frameworks as there are more resources for learning React + Redux.
```
amrut-prabhu: Minor nit
Use `libraries and frameworks` instead of `libraries / frameworks`?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,140 @@
...
+
+### When to not use React
```
amrut-prabhu: The points below do not match this title. You'll need to either change this title or restructure the way the points are presented

amrut-prabhu (changes requested):
@amad-person I've replied to your comments in the previous review as well.

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,140 @@
...
+
+### What is React?
+
+**React** is an open-source JavaScript <tooltip content="React only handles the view component of your application. It is not an MVC framework!">library</tooltip> for building composable user interfaces. It was developed by Facebook and released to the world in 2013, and is now used in a huge number of popular applications, including Facebook and Instagram.
+
+<box type="tip">
+    If you have the <a href="https://github.com/facebook/react-devtools">React DevTools</a> extension installed in your browser, you can inspect the webpage of either of these applications and observe that a lot of React components are being used by them! 
+</box>
+
+React abstracts away the [DOM](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Introduction) from you, giving you a simpler programming model and better performance.
```
marvinchin: I was wondering if it might be useful to add a tooltip explaining what the DOM is, to allow a reader completely new to frontend development to have a rough idea of what is happening without having to leave the article 🙂

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,140 @@
...
+### Why use React?
+
+React allows developers to create interactive web applications where data in specific parts of a page can be changed without reloading the entire page. 
+The main purpose of React is to be fast, scalable, and simple. 
+Some of its benefits are described in detail below.
+
+#### React is Declarative
+
+With React's <tooltip content="Declarative programming focuses on what the program should accomplish without specifying how the program should achieve the result.">declarative</tooltip> approach to building user interfaces, you can create interaction-rich applications. 
+You can build these interfaces without manipulating the DOM directly, and even have an event system that doesn't require you to interact with actual DOM events.
```
marvinchin: > even have an event system that doesn't require you to interact with actual DOM events. 

Perhaps you might want to explain why working with synthetic events is better than native events to illustrate this point

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,140 @@
...
+
+You can see React's declarative approach in action in the sandbox below:
+
+<iframe src="https://codesandbox.io/embed/p564px0j47?fontsize=12" style="width:100%; height:400px; border:0; border-radius: 4px; overflow:hidden;" sandbox="allow-modals allow-forms allow-popups allow-scripts allow-same-origin"></iframe>
+
+<box type="info">
+    The sandbox above is editable. You can play around with the application <code>state</code> and see how the view automatically gets updated. 
+</box>
+
+Notice that with React, you just need to specify *what* you want to do when the application data changes. React will efficiently update and render just the right views without you having to interact with the DOM.
```
marvinchin: I feel like it should be what you want to *show* rather than *do*, since *do* feels a little imperative to me 😅

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,140 @@
...
+- [Angular vs Vue vs React](https://www.codeinwp.com/blog/angular-vs-vue-vs-react/) - in addition to comparing the three technologies, this article aims to give a general structure for comparing JavaScript frameworks and libraries. Hence, you can use this structure to choose between any new frameworks that may arrive in the future.
+- [State of JS 2018: Front-end Frameworks](https://2018.stateofjs.com/front-end-frameworks/overview/) - survey comparing the average salaries, company size, developer satisfaction, etc. for the most used JavaScript front-end technologies as of 2018.
+
+React is best suited for projects that have a lot of **user interactions** that manipulate the DOM extensively. 
+You should choose React if you want the flexibility to choose your application's stack, as it is not <tooltip content="An opiniated framework guides or locks you into their preferred way of doing things.">opinionated</tooltip>.
+Additionally, if you need a mobile version for your app, React Native will make the development process considerably faster as most of the React code in your web version can be reused.
```
marvinchin: Not sure if this is helpful in the context of the article, but from anecdotal experience unfortunately the idea of "write once deploy everywhere" is much harder than it sounds 🙁

marvinchin (commented):
I think this is a great introduction to React 🙂 Just added a couple of comments and suggestions for your consideration!

contents/javascript/Javascript-framework-React.md
amad-person: Thanks for catching that!

contents/javascript/Javascript-framework-React.md
amad-person: Sure!

contents/javascript/Javascript-framework-React.md
amad-person: True...not sure if I should add a note of caution, or remove this statement altogether. What do you think?

amrut-prabhu (approved):

amrut-prabhu:
@amad-person  Some suggestions regarding the sandboxes.

For the 1st sandbox's `styles.css`:

- > Styles for div with class = App

  Styles for div with class = `"`App`"`

- >  The class name is defined in index.js#23 using the JSX keyword className. 

  The class name is defined in index.js#`24` using the JSX keyword `"`className`"`.


For the 2nd sandbox:
- index.js#20:   Passing data to the component is done via `"`props`"`.

- And add this to index.js#30?
  `// This will render App in the div with id as "root" defined in public/index.html.`

- For styles.css: 
  ...defined in index.js#`22` using...

amad-person:
@damithc requesting your review.

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,142 @@
...
+  pageNav: 3
+</frontmatter>
+
+{{ navbar | safe }}
+
+<div class="website-content">
+
+# React
+
+Author: [Aadyaa Maddi](https://github.com/amad-person)
```
damithc: Make this line bold?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,142 @@
...
+
+Reviewers: [Amrut Prabhu](https://github.com/amrut-prabhu), [Marvin Chin](https://github.com/marvinchin)
+
+### What is React?
+
+**React** is an open-source JavaScript <tooltip content="React only handles the view component of your application. It is not an MVC framework!">library</tooltip> for building composable user interfaces. It was developed by Facebook and released to the world in 2013, and is now used in a huge number of popular applications, including Facebook and Instagram.
+
+<box type="tip">
+    If you have the <a href="https://github.com/facebook/react-devtools">React DevTools</a> extension installed in your browser, you can inspect the webpage of either of these applications and observe that React is being used by them! 
+</box>
```
damithc: This tip is probably not needed? I don't think anybody will doubt your claim that Facebook/Instagram is using React. After all, the three are from the same parent company.

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,142 @@
...
+
+# React
+
+Author: [Aadyaa Maddi](https://github.com/amad-person)
+
+Reviewers: [Amrut Prabhu](https://github.com/amrut-prabhu), [Marvin Chin](https://github.com/marvinchin)
+
+### What is React?
+
+**React** is an open-source JavaScript <tooltip content="React only handles the view component of your application. It is not an MVC framework!">library</tooltip> for building composable user interfaces. It was developed by Facebook and released to the world in 2013, and is now used in a huge number of popular applications, including Facebook and Instagram.
```
damithc: This tooltip doesn't match the keyword. A reader who knows what a library is will not checkout the tooltip, thinking that the tooltip gives the definition of a library. Perhaps this?
```
**React** is an open-source JavaScript library for <tooltip content="React only handles the UI componeent of your application. It is not a full web application framework!">building the UI of web apps</tooltip>. It was developed by Facebook and released to the world in 2013, and is now used in a huge number of popular applications, including Facebook and Instagram.
```
I left out the _composable_ part as it felt premature to introduce that technical term. In fact, there is no mention of that term anywhere else in the article which is odd given that it was mentioned in the definition of the tool. Omitted MVC for the same reasons; there is no guarantee that the reader will know what MVC is.
huge -> large?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,142 @@
...
+<box type="tip">
+    If you have the <a href="https://github.com/facebook/react-devtools">React DevTools</a> extension installed in your browser, you can inspect the webpage of either of these applications and observe that React is being used by them! 
+</box>
+
+React abstracts away the <tooltip content="The Document Object Model, or the 'DOM', is an interface to web pages. It represents the page as nodes and objects, allowing programs to read and manipulate the page's content, structure, and styles.">DOM</tooltip> from you, giving you a simpler programming model and better performance.
+It can also [render on the server using Node](https://alligator.io/react/server-side-rendering/), and it can power native mobile applications using [React Native](https://facebook.github.io/react-native/), which is an extension of React.
+
+### Why use React?
+
+React allows developers to create interactive web applications where data in specific parts of a page can be changed without reloading the entire page.
```
damithc: See if you can omit the word `interactive` if it doesn't actually add to the meaning of the sentence.

damithc: You say `data in specific parts of a page can be changed without reloading the entire page`. Is there a need to specifically mention `data`? i.e., only data can be changed?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,142 @@
...
+    If you have the <a href="https://github.com/facebook/react-devtools">React DevTools</a> extension installed in your browser, you can inspect the webpage of either of these applications and observe that React is being used by them! 
+</box>
+
+React abstracts away the <tooltip content="The Document Object Model, or the 'DOM', is an interface to web pages. It represents the page as nodes and objects, allowing programs to read and manipulate the page's content, structure, and styles.">DOM</tooltip> from you, giving you a simpler programming model and better performance.
+It can also [render on the server using Node](https://alligator.io/react/server-side-rendering/), and it can power native mobile applications using [React Native](https://facebook.github.io/react-native/), which is an extension of React.
+
+### Why use React?
+
+React allows developers to create interactive web applications where data in specific parts of a page can be changed without reloading the entire page. 
+The main purpose of React is to be fast, scalable, and simple.
```
damithc: This kind of claims are made by the tool vendor themselves. It's safer to quote such a claim and cite the source rather than you making the claim i.e., stay impartial. e.g.,

React [claims]() that it is _fast_, _scalable_, and _simple_.

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,142 @@
...
+
+React abstracts away the <tooltip content="The Document Object Model, or the 'DOM', is an interface to web pages. It represents the page as nodes and objects, allowing programs to read and manipulate the page's content, structure, and styles.">DOM</tooltip> from you, giving you a simpler programming model and better performance.
+It can also [render on the server using Node](https://alligator.io/react/server-side-rendering/), and it can power native mobile applications using [React Native](https://facebook.github.io/react-native/), which is an extension of React.
+
+### Why use React?
+
+React allows developers to create interactive web applications where data in specific parts of a page can be changed without reloading the entire page. 
+The main purpose of React is to be fast, scalable, and simple. 
+Some of its benefits are described in detail below.
+
```
damithc, line 26: These three sentences appear as three lines in the source code but they form a paragraph in the rendered page. Try to use the same layout in both.

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,142 @@
...
+### What is React?
+
+**React** is an open-source JavaScript <tooltip content="React only handles the view component of your application. It is not an MVC framework!">library</tooltip> for building composable user interfaces. It was developed by Facebook and released to the world in 2013, and is now used in a huge number of popular applications, including Facebook and Instagram.
+
+<box type="tip">
+    If you have the <a href="https://github.com/facebook/react-devtools">React DevTools</a> extension installed in your browser, you can inspect the webpage of either of these applications and observe that React is being used by them! 
+</box>
+
+React abstracts away the <tooltip content="The Document Object Model, or the 'DOM', is an interface to web pages. It represents the page as nodes and objects, allowing programs to read and manipulate the page's content, structure, and styles.">DOM</tooltip> from you, giving you a simpler programming model and better performance.
```
damithc: It's not clear why abstracting the DOM gives better performance. Should the performance be mentioned here at all?
How does abstracting DOM makes the programming model simpler? Perhaps you should give a side-by-side comparison?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,142 @@
...
+It can also [render on the server using Node](https://alligator.io/react/server-side-rendering/), and it can power native mobile applications using [React Native](https://facebook.github.io/react-native/), which is an extension of React.
```
damithc: Why the need to mention 'render on the server'? Is it an important characteristic that defines what React is?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,142 @@
...
+### Why use React?
+
+React allows developers to create interactive web applications where data in specific parts of a page can be changed without reloading the entire page. 
+The main purpose of React is to be fast, scalable, and simple. 
+Some of its benefits are described in detail below.
+
+#### React is Declarative
+
+With React's <tooltip content="Declarative programming focuses on what the program should accomplish without specifying how the program should achieve the result.">declarative</tooltip> approach to building user interfaces, you can create interaction-rich applications. 
+You can build these interfaces without manipulating the DOM directly, and even have an event system that doesn't require you to interact with <tooltip content="Different browsers might have differences in how their DOM event implementations work. React provides its own event system so that events can work in the same way across different browsers.">actual DOM events</tooltip>.
```
damithc: Might need to re-sequence the explanations because it feels like claims are made, and technical terms are being introduced, without explanations. Perhaps you can start by explaining the traditional approach (DOM, imperative approach, etc.) and then show who React differs (declarative approach). Introduce technical terms when you are ready to explain them at the same time.

damithc (changes requested):
Good start. Need a few more rounds of revisions though. 
I added some comments to the top portion. Now that you can see the kind of questions I will raise, take another critical look at the rest of the content and revise as needed.
In general, you may want to distinguish between _defining characteristics_ and _benefits_. Using a declarative approach is a defining characteristic of React, not necessarily a benefit. Perhaps you can give several defining characteristics (i.e., what makes React different from others), and then, give several benefits (ideally, arising from those characteristics)?

contents/javascript/Javascript-framework-React.md
amad-person: I actually wanted to show that React is a popular front-end library, and that React DevTools is a good way for readers to see that for themselves. 

> If you have the <a href="https://github.com/facebook/react-devtools">React DevTools</a> extension installed in your browser, you can inspect the webpages of [many popular applications](https://github.com/facebook/react/wiki/Sites-Using-React) (e.g. Airbnb, Netflix, etc.) and observe that React is being used by them!

Would this tip be okay?

contents/javascript/Javascript-framework-React.md
amad-person: Noted, will update!

contents/javascript/Javascript-framework-React.md
amad-person: I actually wanted to introduce React DevTools in the tip box. I have rephrased the introductory sentence and modified the contents of the tip box to reflect this:

> **React** is an open-source JavaScript library for <tooltip content="React only handles the UI component of your application. It is not a full web application framework!">building the UI of web apps</tooltip>. It was developed by Facebook and released to the world in 2013, and is now used in a [large number of popular applications](https://github.com/facebook/react/wiki/Sites-Using-React), including Facebook and Instagram.

```
<box type="tip"> 
The <a href="https://github.com/facebook/react-devtools">React DevTools</a> browser extension can be used for inspecting the webpages of these applications to see how React is being used by them.
</box>
```

Would this be better?

contents/javascript/Javascript-framework-React.md
amad-person: Internally, the data changes will trigger the different parts of the web page to get updated. I think I can omit the word `data`.

contents/javascript/Javascript-framework-React.md
damithc: Isn't it too early in the article to expect the reader to install a dev tool in the browser? The user has ready only two sentences so far.

amad-person: Good point, I'll remove the link or shift it to `Where to Go From Here` if needed.

amad-person:
@amrut-prabhu @marvinchin @damithc requesting your reviews!
I've made some major changes to the `Why use React?` section, as suggested, and some other improvements.

_markbind/navigation/mainNav.md
```diff
@@ -55,6 +55,7 @@
   * [Android]({{baseUrl}}/contents/android/IntroToAndroid.html)
   * [NodeJS]({{baseUrl}}/contents/node/nodejs.html)
   * [VueJS]({{baseUrl}}/contents/javascript/Javascript-framework-VueJs.html)
+  * [React]({{baseUrl}}/contents/javascript/Javascript-framework-React.html)
```
amrut-prabhu, line 61: Order alphabetically?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,152 @@
...
+  pageNav: 3
+</frontmatter>
+
+{{ navbar | safe }}
+
+<div class="website-content">
+
+# React
+
+**Author: [Aadyaa Maddi](https://github.com/amad-person)** <br/>
```
amrut-prabhu: Change to `<br>` or use markdown to create an empty line?

MarkBind bug? 🤔

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,152 @@
...
+
+### What is React?
+
+**React** is an open-source JavaScript library for <tooltip content="React only handles the UI component of your application. It is not a full web application framework!">building the UI</tooltip> of web applications. It was developed by Facebook and released to the world in 2013. It is now used in a [large number of popular applications](https://github.com/facebook/react/wiki/Sites-Using-React), including Facebook and Instagram.
+
+### Why use React?
+
+React has three main benefits. Let's take a look at each of them in detail.
+
+#### Benefit: Simplicity
```
amrut-prabhu: Consider adding `:thumbsup:`?
Also, since you mentioned `three main benefits` above, should these be numbered?

Depends on how it ends up looking in the rendered page

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,152 @@
...
+<div class="website-content">
+
+# React
+
+**Author: [Aadyaa Maddi](https://github.com/amad-person)** <br/>
+Reviewers: [Amrut Prabhu](https://github.com/amrut-prabhu), [Marvin Chin](https://github.com/marvinchin)
+
+### What is React?
+
+**React** is an open-source JavaScript library for <tooltip content="React only handles the UI component of your application. It is not a full web application framework!">building the UI</tooltip> of web applications. It was developed by Facebook and released to the world in 2013. It is now used in a [large number of popular applications](https://github.com/facebook/react/wiki/Sites-Using-React), including Facebook and Instagram.
```
amrut-prabhu: Do you think it would be better to add the tooltip content into the paragraph itself? 🤔

amrut-prabhu: Broken link?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,152 @@
...
+
+### Why use React?
+
+React has three main benefits. Let's take a look at each of them in detail.
+
+#### Benefit: Simplicity
+
+The traditional <tooltip content="Imperative programming focuses on explicitly describing how a program operates.">imperative</tooltip> approach of building a web application requires you to describe how you want your UI to change when your application data changes. Let's take a look at an example to see how this works. 
+
+Suppose your web application displays a message and has a button to change the message when it is clicked.
```
amrut-prabhu: Suppose your web application displays a message and has a button. You want the message to change when this button is clicked.

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,152 @@
...
+
+One way of doing this is by using [jQuery](https://jquery.com/): 
+
+```javascript
+$("#button").click(function() {
+    $(#message).text('This is the imperative approach!')
```
amrut-prabhu: Suggestions for the Sandbox:

- How about changing the sandbox's message to `This is the declarative approach!`? 😅 

- > Click the button to change it, and React will handle the rest!

  And maybe change ^ to something like `Click the button!` instead? Since we want to urge the reader to click it.

- > To change the message above, you just need to update the application's state.

  And  `You can also change the message by simply changing the application's state.` 
Actually, is this necessary? It's relevant to `Benefit: Simplicity`, but no so much for the title `React is Declarative`, right?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,152 @@
...
+})
+``` 
+
+In the example above, we manually selected the <tooltip content="The Document Object Model, or the 'DOM', is an interface to web pages. It represents the page as nodes and objects, allowing programs to read and manipulate the page's content, structure, and styles.">DOM</tooltip> elements `button` and `message`, attached a listener to `button` for the DOM's `click` event, and wrote a function that describes how to change `message` on clicking `button`. 
+
+This approach is great for simple applications, but web applications usually have to interact with a lot more DOM elements and events. Different browsers have variations in their implementations of the [core DOM](https://quirksmode.org/dom/core/), [events system](https://www.quirksmode.org/dom/events/index.html) and even [styles](https://www.quirksmode.org/dom/w3c_css.html)! If you want your application to work across different browsers, you would need to manually take care of the variations with the imperative approach.
```
amrut-prabhu: but **most** web applications?

**Also,** different browsers?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,152 @@
...
+
+This approach is great for simple applications, but web applications usually have to interact with a lot more DOM elements and events. Different browsers have variations in their implementations of the [core DOM](https://quirksmode.org/dom/core/), [events system](https://www.quirksmode.org/dom/events/index.html) and even [styles](https://www.quirksmode.org/dom/w3c_css.html)! If you want your application to work across different browsers, you would need to manually take care of the variations with the imperative approach. 
+
+React simplifies this process by taking a <tooltip content="Declarative programming focuses on what the program should accomplish without specifying how the program should achieve the result.">**declarative**</tooltip> approach to building UIs. 
+
+To see how this works, let's build the same web application using React instead:
```
amrut-prabhu: The above jQuery example only shows 1 function. Would it be better to say `build the same functionality using`, and specify which line is relevant?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,152 @@
...
+
+<box type="info">
+    The sandbox below is editable. You can play around with the application data (or <code>state</code>) and see how the UI automatically gets updated.
```
amrut-prabhu: ~UI~ **preview** instead?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,152 @@
...
+
+Suppose you have a web application with the following markup in its UI:
+
+```html
+<h3>This is line 1.</h3>
+<h3>This is line 2.</h3>
+<h3>This is line 3.</h3>
+```
+
+The only difference in the three headings above is the number at the end of each line. Wouldn't it be nice if we could reuse a single line of markup for each heading and just pass in a different number each time?
```
amrut-prabhu: >  reuse a single line of markup

Something like `reuse the same markup` instead? 

I had to re-read this a couple of times. The above example also uses a **single line** of markup for each heading (albeit, it doesn't reuse)

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,152 @@
...
+</box>
+
+In the example above, we reused the markup for each heading using an `ExampleComponent`. A React **component** is just a JavaScript function that accepts arbitrary input (called `props`) and returns elements describing what should appear on the screen. 
+
+Components can be especially useful when your UI is complex. With components, you can not only *reuse* code, but also *divide* your UI into encapsulated units that manage their own data and logic. 
+
+For example, if you have a todo list application, it could have a `Form` component to handle adding new items and a `TodoList` component to display all the items. The `TodoList` component could further be composed of multiple `TodoListItem` components, one for each item in the todo list. 
+
+![Building a Todo List application using React](javascript-framework-react-imgs/ReactComponentsExample.png "Building a Todo List application using React")
```
amrut-prabhu: Did you make this yourself? If so, nice work :)
If not, or if it was modified, provide a link to the source please

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,152 @@
...
+
+You can read more about components and props in the React [documentation](https://reactjs.org/docs/components-and-props.html). 
+
+#### Benefit: Better Performance
+
+Web applications can have a lot of user interaction and data updates, which results in changes being made to the DOM. Adding and removing DOM nodes isn't slow. The performance-bottleneck that arises due to DOM manipulation is because the browser needs to [reflow and repaint](https://www.phpied.com/rendering-repaint-reflowrelayout-restyle/) the UI every time the DOM is changed.
+
+React minimizes this update time by using a virtual DOM. The virtual DOM is a JavaScript object that is kept in the memory of your application.
+
+![React's Virtual DOM](javascript-framework-react-imgs/ReactVirtualDOM.png "React's Virtual DOM")
```
amrut-prabhu: Same with this

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,152 @@
...
+
+<box type="info">
+    Notice the HTML-like syntax within the JavaScript method <code>render()</code>. This is <a href="https://reactjs.org/docs/introducing-jsx.html">JSX</a>, and it is a JavaScript syntax extension developed by the React team. React <a href="https://reactjs.org/docs/introducing-jsx.html#why-jsx">recommends</a> developers to use JSX, but it is optional to use it.
+</box>
+
+In the example above, we reused the markup for each heading using an `ExampleComponent`. A React **component** is just a JavaScript function that accepts arbitrary input (called `props`) and returns elements describing what should appear on the screen. 
+
+Components can be especially useful when your UI is complex. With components, you can not only *reuse* code, but also *divide* your UI into encapsulated units that manage their own data and logic. 
+
+For example, if you have a todo list application, it could have a `Form` component to handle adding new items and a `TodoList` component to display all the items. The `TodoList` component could further be composed of multiple `TodoListItem` components, one for each item in the todo list.
```
amrut-prabhu: ~it~ **the `App` component** could have a `Form` component?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,152 @@
...
+
+In the example above, we reused the markup for each heading using an `ExampleComponent`. A React **component** is just a JavaScript function that accepts arbitrary input (called `props`) and returns elements describing what should appear on the screen. 
+
+Components can be especially useful when your UI is complex. With components, you can not only *reuse* code, but also *divide* your UI into encapsulated units that manage their own data and logic. 
+
+For example, if you have a todo list application, it could have a `Form` component to handle adding new items and a `TodoList` component to display all the items. The `TodoList` component could further be composed of multiple `TodoListItem` components, one for each item in the todo list. 
+
+![Building a Todo List application using React](javascript-framework-react-imgs/ReactComponentsExample.png "Building a Todo List application using React")
+
+You can read more about components and props in the React [documentation](https://reactjs.org/docs/components-and-props.html).
```
amrut-prabhu: Should this be in the last section instead?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,152 @@
...
+
+For example, if you have a todo list application, it could have a `Form` component to handle adding new items and a `TodoList` component to display all the items. The `TodoList` component could further be composed of multiple `TodoListItem` components, one for each item in the todo list. 
+
+![Building a Todo List application using React](javascript-framework-react-imgs/ReactComponentsExample.png "Building a Todo List application using React")
+
+You can read more about components and props in the React [documentation](https://reactjs.org/docs/components-and-props.html). 
+
+#### Benefit: Better Performance
+
+Web applications can have a lot of user interaction and data updates, which results in changes being made to the DOM. Adding and removing DOM nodes isn't slow. The performance-bottleneck that arises due to DOM manipulation is because the browser needs to [reflow and repaint](https://www.phpied.com/rendering-repaint-reflowrelayout-restyle/) the UI every time the DOM is changed.
```
amrut-prabhu: `isn't slow` typo?

amrut-prabhu: `reflow and repaint` Better to avoid overly technical terms linking to external sources. Or you can add a tooltip?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,152 @@
...
+
+React minimizes this update time by using a virtual DOM. The virtual DOM is a JavaScript object that is kept in the memory of your application.
+
+![React's Virtual DOM](javascript-framework-react-imgs/ReactVirtualDOM.png "React's Virtual DOM")
+
+Any updates to the UI will first be made to the virtual DOM. Then, React will compare the virtual DOM with the actual DOM using a [diffing algorithm](https://reactjs.org/docs/reconciliation.html#the-diffing-algorithm).
```
amrut-prabhu: Would be nice to refer to the above diagram in this description

contents/javascript/Javascript-framework-React.md
amrut-prabhu: `Actual DOM Updated`: I think a bit more info is needed to bring across the point that you're trying to make. Either in the diagram itself, a title for the diagram (eg. `Figure 2. ...`), or the paragraph below (I think the latter options are easier :P)

contents/javascript/Javascript-framework-React.md
amrut-prabhu: be made to the virtual DOM **object**?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,152 @@
...
+
+![React's Virtual DOM](javascript-framework-react-imgs/ReactVirtualDOM.png "React's Virtual DOM")
+
+Any updates to the UI will first be made to the virtual DOM. Then, React will compare the virtual DOM with the actual DOM using a [diffing algorithm](https://reactjs.org/docs/reconciliation.html#the-diffing-algorithm). 
+Finally, React updates the actual DOM only in places it differs with the virtual DOM. It batches multiple changes together and updates the actual DOM in one go, minimizing update time. 
+
+#### Other Advantages of React
+
+Besides the three main benefits explained above, React has the following advantages:
+ - React only allows data to flow downwards (one-way data binding), which makes your application [easier to debug](https://flaviocopes.com/react-unidirectional-data-flow/).
```
amrut-prabhu: the link doesn't seem relevant to the hyperlink text 🤔

amrut-prabhu (changes requested):
Generally looks good. I enjoyed the new examples and images.

Some suggestions here, mostly nits

contents/javascript/Javascript-framework-React.md
amad-person: Sorry, `:thumbsup:` where exactly?

damithc:
@amad-person let me know when ready for my review.

contents/javascript/Javascript-framework-React.md
amad-person: The page has been removed recently, I'll change the link. Thanks!

contents/javascript/Javascript-framework-React.md
amrut-prabhu: Perhaps,before `Benefit`?

It's put before `Best practice` [here](https://deploy-preview-99--learningresources.netlify.com/contents/uix/accessibility.html).

Like I said, not a must. And may not be totally appropriate actually... 😕 Just thought that I'd point it out

contents/javascript/Javascript-framework-React.md
amad-person: It has the benefits of one-way data binding listed out wrt debugging after the introduction actually.

amrut-prabhu: Yep, it states the `benefits`, whereas the hyperlink text is `easier to debug`, which I expected to explain how and why it is easier in more detail (_maybe just me_). 
Maybe you can paraphrase and add `easier to debug, as you know what is coming from where` in your article?

amad-person: Ah, okay. I can paraphrase.

contents/javascript/Javascript-framework-React.md
amad-person: Modified `https://i.cloudup.com/DCcBG58qYv.png`, I'll put that in.

contents/javascript/Javascript-framework-React.md
amad-person: Not a typo, I think I could make it clear by changing the sentence to `adding and removing DOM nodes isn't slow, but the performance-bottleneck arises from...`

amad-person:
@damithc requesting your review.

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,155 @@
...
+<div class="website-content">
+
+# React
+
+**Author: [Aadyaa Maddi](https://github.com/amad-person)** <br>
+Reviewers: [Amrut Prabhu](https://github.com/amrut-prabhu), [Marvin Chin](https://github.com/marvinchin)
+
+### What is React?
+
+**React** is an open-source JavaScript library for building the UI of web applications. It only handles the UI component of your application, and it is not a full web application framework!
```
damithc: `It only handles the UI component of your application, and it is not a full web application framework! ` It's a bit odd for the reader to encounter this sentence here. On the one hand, it must be something very important as it is the second sentence of the article (and the `!` at the end). On other hand, she would wonder what to make of this fact. For example, is it there to imply 'If you are looking for a framework and not a library, stop reading this article now!'

Find some other way to convey this information. e.g., is it a disadvantage? In that case, it can be moved to the disadvantages section.

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,155 @@
...
+
+To see how this works, let's build the same web application as before, using React instead:
+
+<box type="info">
+    The sandbox below is editable. You can play around with the application data (or <code>state</code>) and see how the UI automatically gets updated. 
+</box>
+
+<iframe src="https://codesandbox.io/embed/p564px0j47?fontsize=12" style="width:100%; height:400px; border:0; border-radius: 4px; overflow:hidden;" sandbox="allow-modals allow-forms allow-popups allow-scripts allow-same-origin"></iframe>
+
+As you can see above, the declarative approach is useful because it abstracts the complexity of interacting with the actual DOM elements and <tooltip content="React provides its own event system so that events can work in the same way across different browsers.">events</tooltip>. With React, you just need to specify *what* you want to show in the UI when the application data changes.
```
damithc: All you are saying here in this section is _React is declarative_. It is a big leap from that to the section heading `Benefit 1: Simplicity`. Take a look at Ronak's article on Angular. It too had a similar confusion between features and benefits but we have managed to untangle the two by now.

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,155 @@
...
+# React
+
+**Author: [Aadyaa Maddi](https://github.com/amad-person)** <br>
+Reviewers: [Amrut Prabhu](https://github.com/amrut-prabhu), [Marvin Chin](https://github.com/marvinchin)
+
+### What is React?
+
+**React** is an open-source JavaScript library for building the UI of web applications. It only handles the UI component of your application, and it is not a full web application framework! 
+
+React was developed by Facebook and released to the world in 2013. It is now used in a [large number of popular applications](https://medium.com/@coderacademy/32-sites-built-with-reactjs-172e3a4bed81), including Facebook and Instagram.
```
damithc: Keep in mind that this is targeted to someone who is totally new to React. I suggest you walk the reader through a simple React example so that the reader has some concrete idea about what it is. After that, the user has a better context on which she can 'hang' the the advantages and disadvantages of React.

damithc (changes requested):
I added a couple of comments but aborted the review half-way. Take a look at Ronak's article on Angular. That article too had similar flow problems but they are mostly fixed now. I would like this to be structured similar to that, as this topic is similar. Unlike you guys, I have the advantage of actually not knowing React/Angular, and hence can feel the awkwardness of the current flow that you might not detect.

The WHAT section needs to be bigger so that the reader (supposedly, totally new to React) has a good sense of what React is before she considers advantages/disadvantages (i.e., the WHY) of it. 
In particular, the WHAT section should describe, using a suitable React example app, the key React features, such as declarative nature, being only a UI library, being non-opinionated, etc. Just describe them as what they are, without making judgments on whether they are good or bad. In the WHY section you can give your opinions (or cite others' opinions) on the advantages of React, referring back to the features explained in the WHAT section. The reader might not agree with your WHY section (i.e., advantages -- they are subjective) but she should not have room for any disagreements with the WHAT section (i.e., features).

amad-person:
@damithc ready for review.

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,189 @@
...
+
+### Why use React?
+
+Now that we know what React is, let us take a look at some of its benefits.
+
+#### Benefit 1: Abstraction
+
+The traditional <tooltip content="Imperative programming focuses on explicitly describing how a program operates.">*imperative*</tooltip> approach of building a web application requires you to describe how you want your UI to change when your application data changes. 
+
+Most web applications usually have to interact with a lot of <tooltip content="The Document Object Model, or the 'DOM', is an interface to web pages. It represents the page as nodes and objects, allowing programs to read and manipulate the page's content, structure, and styles.">DOM</tooltip> elements and events. Also, different browsers have variations in their implementations of the [core DOM](https://quirksmode.org/dom/core/), [events system](https://www.quirksmode.org/dom/events/index.html) and even [styles](https://www.quirksmode.org/dom/w3c_css.html)! If you want your application to work across different browsers, you would need to manually take care of these variations with the imperative approach.
```
damithc: For `core DOM` etc., you can also give a tooltip so that the reader has an option of learning what the term means without having to go to the external link.

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,189 @@
...
+
+Most web applications usually have to interact with a lot of <tooltip content="The Document Object Model, or the 'DOM', is an interface to web pages. It represents the page as nodes and objects, allowing programs to read and manipulate the page's content, structure, and styles.">DOM</tooltip> elements and events. Also, different browsers have variations in their implementations of the [core DOM](https://quirksmode.org/dom/core/), [events system](https://www.quirksmode.org/dom/events/index.html) and even [styles](https://www.quirksmode.org/dom/w3c_css.html)! If you want your application to work across different browsers, you would need to manually take care of these variations with the imperative approach. 
+
+React's declarative approach simplifies this process because it abstracts the complexity of interacting with the actual DOM elements and events. For example, it provides its own [event system](https://reactjs.org/docs/events.html) so that events can work in the same way across different browsers.
```
damithc: Is this related to the idea of _virtual DOM_? i.e., React programmers work with a virtual DOM that is same no matter which browser it is (similar to how Java programmers write against the JVM no matter what the OS is)

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,189 @@
...
+
+#### Benefit 2: Better Performance
+
+Web applications can have a lot of user interaction and data updates, which results in changes being made to the DOM. Adding and removing DOM nodes isn't slow, but the performance-bottleneck arises because the browser needs to *[reflow and repaint](https://www.phpied.com/rendering-repaint-reflowrelayout-restyle/)* the UI every time the DOM is changed.
+
+React minimizes this update time by using a *virtual DOM*. The virtual DOM is a JavaScript object that is kept in the memory of your application.
+
+![React's Virtual DOM](javascript-framework-react-imgs/ReactVirtualDOM.png "React's Virtual DOM")
+_Figure 1. How React's actual DOM gets updated._
```
damithc: The diagram is a bit hard to follow. Show actual and virtual DOMs slightly differently?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,189 @@
...
+
+A React application is made up of a combination of components. Components are independent from each other, and like functions, they map the same input to the same output. This makes it easy to write unit tests for your application.
+
+Additionally, React only allows data to flow downwards (one-way data binding) using `state` and `props`, which makes your application [easier to debug](https://reactjs.org/docs/design-principles.html#debugging) as you can be sure that the data updates the UI, and never the other way around. 
+
+#### Other Advantages of React
+
+Besides the three main benefits explained above, React has the following advantages:
+ - React is not an <tooltip content="An opiniated framework guides or locks you into their preferred way of doing things.">*opinionated*</tooltip> framework, which gives you the flexibility to choose your application stack.
+ - As React is an open-source library backed by Facebook, it is constantly being updated with new features and performance optimizations.
```
damithc: > As React is an open-source library backed by Facebook, it is constantly being updated with new features and performance optimizations.

It's not clear why the former results in the latter.

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,189 @@
...
+
+### Disadvantages of React
+
+Some of React's disadvantages are given below:
+ - The high pace of development means that you would need to regularly relearn how to do things.
```
damithc: development of what? React or your app?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,189 @@
...
+ - React is just a UI library. As React only allows one-way data binding, you can't use it in applications that follow the <tooltip content="MVC, or Model-View-Controller is a popular application structure for building web applications. It separates the UI and the data of your application, and changes can be made to your application using controllers.">*MVC*</tooltip> architecture. You would have to use [Flux](https://github.com/facebook/flux), an application architecture that favours one-way data binding instead.
```
damithc: favors (se-edu uses American English)

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,189 @@
...
+
+### How to get started with React?
+
+The official React [website](https://reactjs.org/) is a great place to get started. It includes:
+ - A step-by-step [tutorial](https://reactjs.org/tutorial/tutorial.html) for building a React application, if you prefer to learn by doing.
+ - A [guide](https://reactjs.org/docs/hello-world.html) to master the main concepts of React, if you prefer to learn by reading instead.
+ 
+If you want to add React to an existing project, you can take a look at React's official [guide](https://reactjs.org/docs/add-react-to-a-website.html) for doing so. Alternatively, if you are creating a new React application, you can use one of the [recommended toolchains](https://reactjs.org/docs/create-a-new-react-app.html) to get the best user and developer experience.
+
+[Create React App](https://github.com/facebook/create-react-app) is a comfortable environment for learning React, and it is the recommended way to create <tooltip content="A single-page application is an app that works inside a browser and does not require page reloading during use.">*single-page applications*</tooltip> with React.
```
damithc: `comfortable`: did you mean `convenient`?

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,189 @@
...
+
+The official website also has [advanced guides](https://reactjs.org/docs/getting-started.html#advanced-concepts) if you want to understand how React works behind the scenes.
+
+### Where to go from here?
```
damithc: Perhaps name this `Other React resources`? The reason is, the previous section explains _how to get started_ and for most our readers that is also _where to go next_.

contents/javascript/Javascript-framework-React.md
```diff
@@ -0,0 +1,189 @@
...
+
+React's declarative approach simplifies this process because it abstracts the complexity of interacting with the actual DOM elements and events. For example, it provides its own [event system](https://reactjs.org/docs/events.html) so that events can work in the same way across different browsers.
+
+#### Benefit 2: Better Performance
+
+Web applications can have a lot of user interaction and data updates, which results in changes being made to the DOM. Adding and removing DOM nodes isn't slow, but the performance-bottleneck arises because the browser needs to *[reflow and repaint](https://www.phpied.com/rendering-repaint-reflowrelayout-restyle/)* the UI every time the DOM is changed.
+
+React minimizes this update time by using a *virtual DOM*. The virtual DOM is a JavaScript object that is kept in the memory of your application.
+
+![React's Virtual DOM](javascript-framework-react-imgs/ReactVirtualDOM.png "React's Virtual DOM")
```
damithc: `imgs` -> `images`?

damithc (changes requested):
Looks quite good now. Almost LGTM. Just  few cosmetic nits.

contents/javascript/Javascript-framework-React.md
amad-person: React 😅 I'll update this

contents/javascript/Javascript-framework-React.md
amad-person: Yep, it is related. I think I should change the order of the benefits (i.e. introduce the virtual DOM section first) so I can link back later.

contents/javascript/Javascript-framework-React.md
amad-person: Noted, I'll update this

amad-person:
@damithc updated after the latest review! Please take a look

damithc:
Merged and deployed. Good work @amad-person 
Thanks to reviewers too.

PR #75, opened 2019-02-16
Add scala page

fzdy1914:


damithc:
A quick comment: Don't write this as a step-by-step tutorial. I'm sure there are enough such tutorials out there already without us having to write another. The article is meant to persuade, not train. Focus on WHAT and WHY, not HOW. Give links to such tutorials rather than replicate their content.

fzdy1914:
> A quick comment: Don't write this as a step-by-step tutorial. I'm sure there are enough such tutorials out there already without us having to write another. The article is meant to persuade, not train. Focus on WHAT and WHY, not HOW. Give links to such tutorials rather than replicate their content.

What I have written is not all of my content, I will add more content later.

damithc:
> What I have written is not all of my content, I will add more content later.

Sure, but some the current content needs to be removed e.g., set up instructions

fzdy1914:
@junming403 @Adoby7 Can you help review my page?

junming403:
Sure, will review by tonight

contents/scala/Scala.md
```diff
@@ -0,0 +1,241 @@
...
+  * [Type Enrichment](#type-enrichment)
+  * [Pattern Matching](#pattern-matching)
+
+# Overview
+
+**Scala** is a modern *multi-paradigm* programming language. It designed to integrate features of 
+*object-oriented* and *functional* languages in a *concise*, *elegant*, and *type-safe* way.
+
+## Characteristic of Scala
+
```
junming403: One big feature about Scala is its JVM and JavaScript runtimes let you build high-performance systems with easy access to huge ecosystems of libraries. Scala provides language interoperability with Java, so that libraries written in either language may be referenced directly in Scala or Java code. I think it is worth mention here.

contents/scala/Scala.md
```diff
@@ -0,0 +1,241 @@
...
+### Statically Typed
+Scala is **statically typed**, which means it enforces type checking ike verifying and enforcing the constraints of 
+types at compile-time. 
+
+Scala allows [**type inference**](https://docs.scala-lang.org/tour/type-inference.html), which means it detects the data 
+type of an expression automatically. User is not required to annotate redundant type information in Scala. 
+
+In combination, these features provide a clean but reliable programming basis for the user.
```
junming403: Can we have a simple but concrete example here?

contents/scala/Scala.md
```diff
@@ -0,0 +1,241 @@
...
+
+### Object Oriented
+Scala is a **pure OOP language**, which means every value in it is an object, including functions and primitives.
```
junming403: Mention Scala's relation with java will help user quickly understand how scala's OOP model works

contents/scala/Scala.md
```diff
@@ -0,0 +1,241 @@
...
+
+Types and behavior of objects are described by *classes* and *traits*.
```
junming403: Provide a link to traits as not everyone knows what it is? Or can mention it is similar to java interface?

contents/scala/Scala.md
```diff
@@ -0,0 +1,241 @@
...
+Scala is a **Functional Programming language** in the sense that every function in Scala is a value. 
+
+Scala provides a lightweight syntax for defining [*anonymous functions*](https://docs.scala-lang.org/tour/basics.html#functions).
+
+It also support many other features of FP including [*higher-order functions*](https://docs.scala-lang.org/tour/higher-order-functions.html), 
+[*nested functions*](https://docs.scala-lang.org/tour/nested-functions.html), [*currying*](https://docs.scala-lang.org/tour/multiple-parameter-lists.html), 
+[*pattern matching*](https://docs.scala-lang.org/tour/pattern-matching.html), *immutability* and *lazy evaluation*.
+
+# Getting Started
+There are mainly two ways to work in Scala.
```
junming403: Feels like this section should be at the end of the article, and other than set up, you probably can provide some links to beginner-friendly resources. 

For example, 
[tour of scala](https://docs.scala-lang.org/tour/tour-of-scala.html)

contents/scala/Scala.md
```diff
@@ -0,0 +1,241 @@
...
+2. Using the **command line**. Refer to the [instructions](https://docs.scala-lang.org/getting-started-sbt-track/getting-started-with-scala-and-sbt-on-the-command-line.html) for more details.
+
+# Why Scala
+There are mainly two benefits of Scala, **simple syntax** and **multi-paradigm**. 
+
+*Simple syntax* allows us to write more simple and elegant code. 
+
+*Multi-paradigm* allows us to use an OOP language while using the feature in functional programming. 
+
+Following are some examples to illustrate the benefit of Scala:
```
junming403: *benefits

contents/scala/Scala.md
```diff
@@ -0,0 +1,241 @@
...
+print in TraitC
+print in TraitB
+print in TraitA
+print in Sample
+```
+The explanation is that the call to `print()` firstly executes the code in `TraitC`, which is the last trait mixed in.
+Then through the `super()` calls, it threads back through the other mixed-in traits. And eventually to the code in 
+`Sample`. Even though none of the traits inherited from one another, it works properly. 
+
+This is similar to the decorator pattern but is more *concise* and less *error-prone*. In other languages, a similar
```
junming403: Provide a link to decorator pattern?

contents/scala/Scala.md
```diff
@@ -0,0 +1,241 @@
...
+    println("print in TraitB")
+    super.print()
+  }
+}
+
+trait TraitC extends ParentTrait {
+  abstract override def print() {
+    println("print in TraitC")
+    super.print()
+  }
```
junming403: length code might scare the reader, probably remove `TraitC` ?

contents/scala/Scala.md
```diff
@@ -0,0 +1,241 @@
...
+print in TraitA
+print in Sample
+```
+The explanation is that the call to `print()` firstly executes the code in `TraitC`, which is the last trait mixed in.
+Then through the `super()` calls, it threads back through the other mixed-in traits. And eventually to the code in 
+`Sample`. Even though none of the traits inherited from one another, it works properly. 
+
+This is similar to the decorator pattern but is more *concise* and less *error-prone*. In other languages, a similar 
+effect could be achieved with a long linear chain of inheritance. But the disadvantage is that for every possible 
+combination of the mix-ins, we need to declare an inheritance chain for it.
```
junming403: Good point!

contents/scala/Scala.md
```diff
@@ -0,0 +1,241 @@
...
+Following is an example of QuickSort algorithm using pattern matching:
+
+```
+def qsort(list: List[Int]): List[Int] = list match {
+  case Nil => Nil
+  case pivot :: tail =>
+    val (smaller, rest) = tail.partition(_ < pivot)
+    qsort(smaller) ::: pivot :: qsort(rest)
+}
+```
```
junming403, line 198: Readers new to this might find it hard to understand, start with some simpler examples first?

contents/scala/Scala.md
```diff
@@ -0,0 +1,241 @@
...
+But `pivot :: tail` matches a non-empty list, and it will destructure the list according to the pattern given. 
+In this case, the code will have a variable `pivot` holding the head of the list, and another variable `tail`
+ holding the tail of the list.
+ 
+**Pattern matching** also happens in local variable declarations. The return value of the call to `tail.partition` is a 
+tuple. In this case, the code will have a variable `smaller` holding the first element of the tuple, and another 
+variable `rest` holding the second element of the list. 
+**Pattern matching** is the easiest way of fetching the two parts of the tuple.
+
+</div>
```
junming403, line 226: Have a section for "where to go from here", provide some useful resources will be ideal

contents/scala/Scala.md
```diff
@@ -0,0 +1,241 @@
...
+
+```
+def qsort(list: List[Int]): List[Int] = list match {
+  case Nil => Nil
+  case pivot :: tail =>
+    val (smaller, rest) = tail.partition(_ < pivot)
+    qsort(smaller) ::: pivot :: qsort(rest)
+}
+```
+
```
junming403, line 215: Many simple but concrete examples will be better. 

For example, list matching with guards
```
        lst match {
            case head :: _ if head == 1 => println("head element is 1")
            case head :: tail => println(s"General:  head ${head} and tail ${tail}!")
            case Nil => println("Empty list")
        }
```

contents/scala/Scala.md
```diff
@@ -0,0 +1,241 @@
...
+The idea here is that we *partition* a list recursively, sort each part and combine the results together.(See [QuickSort](https://en.wikipedia.org/wiki/Quicksort) for the details of the algorithm)
+
+The `match` operator is used to do *pattern matching* on the object stored in list. Each case expression is tried to 
+see if it will match, and the first match will be executed. 
+
+In this case, `Nil` only matches the object `Nil` (Empty list).
+But `pivot :: tail` matches a non-empty list, and it will destructure the list according to the pattern given. 
+In this case, the code will have a variable `pivot` holding the head of the list, and another variable `tail`
+ holding the tail of the list.
+
```
junming403: There are many ways of matching in Scala, case class pattern matching, constructor pattern matching, tuple matching etc, you can either briefly mention them or have a [link](https://alvinalexander.com/scala/how-to-use-pattern-matching-scala-match-case-expressions) to show to the users how strong is Scala's pattern matching.

junming403 (changes requested):
Good work overall, I have given some suggestions below. However, you may need to find someone who doesn't know about Scala to help you review, because I am not so sure if it is too hard for readers who do not know Scala(or even functional programming) before.

contents/scala/Scala.md
```diff
@@ -0,0 +1,241 @@
...
+</frontmatter>
+
+{{ navbar | safe }}
+
+<div class="website-content">
+
+# Introduction to Scala
+
+Author: [Wang Chao](https://github.com/fzdy1914)
+
```
Adoby7: Suggest to change Author(s) for other people to add content

contents/scala/Scala.md
```diff
@@ -0,0 +1,241 @@
...
+# Overview
+
+**Scala** is a modern *multi-paradigm* programming language. It designed to integrate features of 
+*object-oriented* and *functional* languages in a *concise*, *elegant*, and *type-safe* way.
+
+## Characteristic of Scala
+
+### Statically Typed
+Scala is **statically typed**, which means it enforces type checking ike verifying and enforcing the constraints of
```
Adoby7: typo "like"

contents/scala/Scala.md
Adoby7: On the top of what Junming has said, it's better to provide a learning path (e.g. from beginner to advanced skills in Scala). Also, give the summary of each link.

contents/scala/Scala.md
```diff
@@ -0,0 +1,241 @@
...
+
+```
+val a = ComplexInt(1, 1)
+val b = ComplexInt(1, -1)
+val c = a * b
+val d = -c + a
+```
+
+In other languages like Java, the similar operations of the same class may need to performed like following:
+
```
Adoby7, line 89: I think it's better to put the Java at beginning, and say something like "You can only write new methods to handle computation, which may cause XXX problems". Scala supports operator overloading that can solve them. This could be more impressive

contents/scala/Scala.md
Adoby7: Similar here. The disadvantage part can be move at the beginning, and provide some short but clear example of disadvantage. If readers know what's the problem, they may understand the features in Scala better.

contents/scala/Scala.md
```diff
@@ -0,0 +1,241 @@
...
+[*nested functions*](https://docs.scala-lang.org/tour/nested-functions.html), [*currying*](https://docs.scala-lang.org/tour/multiple-parameter-lists.html), 
+[*pattern matching*](https://docs.scala-lang.org/tour/pattern-matching.html), *immutability* and *lazy evaluation*.
+
+# Getting Started
+There are mainly two ways to work in Scala.
+
+1. Using the **INTELLIJ IDE**. Refer to the [instructions](https://docs.scala-lang.org/getting-started-intellij-track/getting-started-with-scala-in-intellij.html) for more details.
+2. Using the **command line**. Refer to the [instructions](https://docs.scala-lang.org/getting-started-sbt-track/getting-started-with-scala-and-sbt-on-the-command-line.html) for more details.
+
+# Why Scala
```
Adoby7: This section looks like "how scala is used" which is quite technical. You may also add the motivation to make readers interesting in Scala. i.e. what's the WIIFY. E.g. learning Scala can enhance my programming skill. Learning Scala can help me find a job.

contents/scala/Scala.md
```diff
@@ -0,0 +1,241 @@
...
+  * [Mixed-In Trait](#mixed-in-trait)
+  * [Type Enrichment](#type-enrichment)
+  * [Pattern Matching](#pattern-matching)
+
+# Overview
+
+**Scala** is a modern *multi-paradigm* programming language. It designed to integrate features of 
+*object-oriented* and *functional* languages in a *concise*, *elegant*, and *type-safe* way.
+
+## Characteristic of Scala
```
Adoby7: Could this part be merged in "Why scala" section? It seems that both of them are Scala features.

Adoby7 (commented):
An overall comment: when you compare Scala with other languages. You may first state the problem of that language, and then says Scala has XX feature that can solve the problem. This will make a better impression, and readers will understand your example code of Scala better.

contents/scala/Scala.md
fzdy1914: For JavaScript, are you referring to Scala.js? As I am talking about pure Scala language, maybe only mention JVM here.

contents/scala/Scala.md
fzdy1914: I would like to categorize them as Characteristics instead of features. From my perspective, characteristics are more general and feature are more specific.

contents/scala/Scala.md
fzdy1914:
```
There are mainly two benefits of Scala, **simple syntax** and **multi-paradigm**. 

*Simple syntax* allows us to write more simple and elegant code. 

*Multi-paradigm* allows us to use an OOP language while using the feature in functional programming. 
```
I believe this part has illustrated the benefit of Scala?

fzdy1914:
> An overall comment: when you compare Scala with other languages. You may first state the problem of that language, and then says Scala has XX feature that can solve the problem. This will make a better impression, and readers will understand your example code of Scala better.

Maybe we can ask Prof's opinion on this? I am not sure which one is better. @damithc

damithc:
> Maybe we can ask Prof's opinion on this? I am not sure which one is better. @damithc

I have no strong opinion on this. I'll leave it to the author.

contents/scala/Scala.md
```diff
@@ -0,0 +1,279 @@
...
+  pageNav: 3
+</frontmatter>
+
+{{ navbar | safe }}
+
+<div class="website-content">
+
+# Introduction to Scala
+
+Author(s): [Wang Chao](https://github.com/fzdy1914)
```
Adoby7: From my experience, add reviewers here as well, and make these two lines bold

contents/scala/Scala.md
```diff
@@ -0,0 +1,279 @@
...
+#### Statically Typed
+Scala is **statically typed**, which means it enforces type checking like verifying and enforcing the constraints of 
+types at compile-time. 
+
+Scala allows [**type inference**](https://docs.scala-lang.org/tour/type-inference.html), which means it detects the data 
+type of an expression automatically. User is not required to annotate redundant type information in Scala. 
+
+Type Inference Example:
+```
+val x = "foo"
```
Adoby7: probably add `x = 2` and expect compile error to show `x` really has a type

contents/scala/Scala.md
```diff
@@ -0,0 +1,279 @@
...
+```
+val a = ComplexInt(1, 1)
+val b = ComplexInt(1, -1)
+val c = a * b
+val d = -c + a
+```
+
+In other languages like Java, the similar operations of the same class may need to performed like following:
+
+```
```
Adoby7: I'm not sure about Scala, but for Java you can use 
```java
// code
```
for code

contents/scala/Scala.md
```diff
@@ -0,0 +1,279 @@
...
+* [Scala Exercises](https://www.scala-exercises.org/) is a series of lessons and exercises created by 47 Degrees. 
+It's a great way to get a brief introduction to Scala while testing your knowledge along the way.
+
+* [Functional Programming Principles in Scala](https://www.coursera.org/course/progfun), free on Coursera. 
+This is a course about functional programming given by Martin Odersky himself. 
+You can access the course material and exercises by signing up for the course.
+
+* [allaboutscala](http://allaboutscala.com/) provides detailed tutorials for beginners.
+
+More details can be find here: [LEARNING RESOURCES](https://scala-lang.org/documentation/learn.html).
```
Adoby7: Probably name it "Scala Learning Resources"

Adoby7 (approved):
It looks better now, I give some more comments.

contents/scala/Scala.md
```diff
@@ -0,0 +1,279 @@
...
+* [Where to go from here](#where-to-go-from-here)
+
+## Overview
+
+**Scala** is a modern *multi-paradigm* programming language. It designed to integrate features of 
+*object-oriented* and *functional* languages in a *concise*, *elegant*, and *type-safe* way.
+
+Scala's JVM runtimes allows us to build high-performance systems with easy access to huge ecosystems of libraries. Scala 
+provides language interoperability with Java, so that libraries written in Java may be referenced directly in Scala.
+Also, by Scala.js, Scala code can be easily compiled to JavaScript.
```
junming403: It is not clear what is Scala.js, add a link for it?

contents/scala/Scala.md
```diff
@@ -0,0 +1,279 @@
...
+ holding the tail of the list.
+ 
+In local variable declarations, tuple matching is used. The return value of the call to `tail.partition` is a 
+tuple. In this case, the code will have a variable `smaller` holding the first element of the tuple, and another 
+variable `rest` holding the second element of the list. 
+
+## Getting Started
+There are mainly two ways to work in Scala.
+
+1. Using the **INTELLIJ IDE**. Refer to the [instructions](https://docs.scala-lang.org/getting-started-intellij-track/getting-started-with-scala-in-intellij.html) for more details.
```
junming403: I still feel it is not necessary to mention it here, probably the how to get started should focus on how to get started on using the language itself?

junming403 (approved):
Look good to me overall, just a few suggestions for you to consider.

fzdy1914:
@damithc Can you kindly give your review for this?

contents/scala/Scala.md
```diff
@@ -0,0 +1,284 @@
...
+    * [Statically Typed](#statically-typed)
+    * [Object Oriented](#object-oriented)
+    * [Functional](#functional)
+* [Why Scala](#why-scala)
+  * [Operator Overloading](#operator-overloading)
+  * [Mixed-In Trait](#mixed-in-trait)
+  * [Type Enrichment](#type-enrichment)
+  * [Pattern Matching](#pattern-matching)
+* [Getting Started](#getting-started)
+* [Where to go from here](#where-to-go-from-here)
```
damithc: No need for ToC as we have an auto-generated page navigation menu on the right.

contents/scala/Scala.md
```diff
@@ -0,0 +1,284 @@
...
+
+## Overview
+
+**Scala** is a modern *multi-paradigm* programming language. It designed to integrate features of
```
damithc: Can remove `modern`

contents/scala/Scala.md
```diff
@@ -0,0 +1,284 @@
...
+  * [Mixed-In Trait](#mixed-in-trait)
+  * [Type Enrichment](#type-enrichment)
+  * [Pattern Matching](#pattern-matching)
+* [Getting Started](#getting-started)
+* [Where to go from here](#where-to-go-from-here)
+
+## Overview
+
+**Scala** is a modern *multi-paradigm* programming language. It designed to integrate features of 
+*object-oriented* and *functional* languages in a *concise*, *elegant*, and *type-safe* way.
```
damithc: We usually don't add line breaks to enforce a line width in markdown files.

damithc: Can omit `concise` and `elegant` as these are debatable.

contents/scala/Scala.md
```diff
@@ -0,0 +1,284 @@
...
+
+Scala's JVM runtimes allows us to build high-performance systems with easy access to huge ecosystems of libraries. Scala 
+provides language interoperability with Java, so that libraries written in Java may be referenced directly in Scala.
+Also, by *[Scala.js](http://www.scala-js.org/)*, Scala code can be easily compiled to JavaScript.
+
+### Characteristic of Scala
+
+#### Statically Typed
```
damithc: Let's remove these headings. As this is a short article, just a few high-level headings are enough.

contents/scala/Scala.md
```diff
@@ -0,0 +1,284 @@
...
+    * [Functional](#functional)
+* [Why Scala](#why-scala)
+  * [Operator Overloading](#operator-overloading)
+  * [Mixed-In Trait](#mixed-in-trait)
+  * [Type Enrichment](#type-enrichment)
+  * [Pattern Matching](#pattern-matching)
+* [Getting Started](#getting-started)
+* [Where to go from here](#where-to-go-from-here)
+
+## Overview
```
damithc, line 20: I suggest you simply quote Scala website rather than try to rephrase the same thing using different words. This way, you can remain independent as the claims about the language (e.g., _elegant_, _concise_ etc.) are coming from an official source:

Example:

# Scala

## Overview

>Scala combines object-oriented and functional programming in one concise, high-level language. Scala's static types help avoid bugs in complex applications, and its JVM and JavaScript runtimes let you build high-performance systems with easy access to huge ecosystems of libraries. -- [Scala Website](https://www.scala-lang.org/)

Let's look at some characteristics of Scala mentioned in the above quote:

* **Scala supports _object-oriented_ programming**: Every value in Scala is an object, including functions and primitives. Classes are extended by *subclassing* and a flexible *mixin-based* mechanism to avoid the problems of multiple inheritance.
* **Scala supports _functional_ programming**: Every function in Scala is a value. Scala provides a lightweight syntax for defining [*anonymous functions*](https://docs.scala-lang.org/tour/basics.html#functions). It also support many other features of FP including [*higher-order functions*](https://docs.scala-lang.org/tour/higher-order-functions.html), [*nested functions*](https://docs.scala-lang.org/tour/nested-functions.html), [*currying*](https://docs.scala-lang.org/tour/multiple-parameter-lists.html), [*pattern matching*](https://docs.scala-lang.org/tour/pattern-matching.html), *immutability* and *lazy evaluation*.
* **Scala is _statically typed_**: It enforces type checking like verifying and enforcing the constraints of 
types at compile-time. 
* **Scala can work the Java and JavaScript runtimes**: ...

## Noteworthy Scala Features

### Operator overloading
...

## Getting Started
...

contents/scala/Scala.md
```diff
@@ -0,0 +1,284 @@
...
+
+## Getting Started
+There are mainly two ways to work in Scala.
+
+1. Using the **INTELLIJ IDE**. Refer to the [instructions](https://docs.scala-lang.org/getting-started-intellij-track/getting-started-with-scala-in-intellij.html) for more details.
+2. Using the **command line**. Refer to the [instructions](https://docs.scala-lang.org/getting-started-sbt-track/getting-started-with-scala-and-sbt-on-the-command-line.html) for more details.
+
+A more detailed tutorial can be found here: [Tour of Scala](https://docs.scala-lang.org/tour/tour-of-scala.html)
+
+## Where to go from here
```
damithc: Merge with the previous section.

contents/scala/Scala.md
```diff
@@ -0,0 +1,284 @@
...
+ holding the tail of the list.
+ 
+In local variable declarations, tuple matching is used. The return value of the call to `tail.partition` is a 
+tuple. In this case, the code will have a variable `smaller` holding the first element of the tuple, and another 
+variable `rest` holding the second element of the list. 
+
+## Getting Started
+There are mainly two ways to work in Scala.
+
+1. Using the **INTELLIJ IDE**. Refer to the [instructions](https://docs.scala-lang.org/getting-started-intellij-track/getting-started-with-scala-in-intellij.html) for more details.
```
damithc: Intellij is the only IDE for scala? doesn't sound right. In any case, give a learning path in this section i.e., in your opinion, what exact path should the reader follow to learn Scala?

damithc (changes requested):
The content is good 👍  Just need to tidy up the structure a bit.

contents/scala/Scala.md
fzdy1914: @damithc Do you mean that some add-on characteristics are not worth mention here, for example, type inference, Traits?

damithc: Yes, let's limit to those mentioned in the quote from the website. Traits, type inference etc. can be mentioned as _features_; they are not characteristic (i.e., not related to the overall nature of the language).

fzdy1914:
@damithc I have edited as request, please kindly give a review again.

damithc:
Merged and deployed.
Good work @fzdy1914 
Thanks to reviewers too.

Note that I did many language and formatting tweaks.
Also, minimized claims such as _elegant_ and _simple_ because such claims are debatable.
Removed the learning path as it was too detailed.

PR #82, opened 2019-03-01
Fix broken links on project readme

marvinchin:
The current links on the project README is broken. I've updated the links to point to the relevant pages on the site instead.


damithc:
Thanks for the fix @marvinchin

PR #84, opened 2019-03-01
Add page on password storage.

### LINKED ISSUE #65: Password storage
I would like to write a chapter on storing passwords

Textbook outline:

Password storage intro
What's wrong with plaintext?
What is hashing?
Why isn't hashing enough?
Adding salt
The salted hash
If an attacker has gained access to the entire application, what does it matter if my passwords are stored in plaintext or not?
Word of warning
Further reading

ChooJeremy:
Closes #65

Looking for anyone to review.


contents/security/passwordStorage.md
```diff
@@ -0,0 +1,98 @@
...
+
+### What's wrong with <trigger for="pop:plaintext">plaintext</trigger>?
+
+<popover id="pop:plaintext" title="_Plaintext_ refers to unencrypted information" placement="top">
+  <div slot="content">
+It is a cryptography term generally referring to text before encyption or after decrypting it. Another term for it is cleartext.
+  </div>
+</popover>
+
+While usernames are typically public information, thus encrypting them is not as important, passwords are typically secret information and require encryption. If passwords are stored in plaintext and your server's database is ever compromised, attackers can easily use the stored plaintext passwords to login as the user and potentially access other functionality. This grants the attacker more power than just reading values off the database and being unable to read them if they were encrypted.
```
marvinchin: Do you think explaining what encryption is might be helpful to beginners to this topic? 🙂 (perhaps you could consider linking it to an explanation elsewhere as well)

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,98 @@
...
+
+To prevent all of these measures, it is good to perform hashing on passwords. Unlike encryption, hashing is a <trigger for="pop:oneway">one-way function</trigger>. This means that an attacker cannot decrypt the encrypted password to retrieve the original password, thus ensuring that all of the above vulnerabilities isn't possible.
+
+<popover id="pop:oneway" title="A one way function is a function that is easy to compute in one way, but is close to impossible to compute the other way." placement="top">
+<div slot="content"></div>
+</popover>
+
+## What is hashing?
+
+Hashing is the act of transforming a set of data into another set of data that is difficult to reverse. As an example, suppose you wanted to multiply 2 prime numbers together, such as 17 and 11. This is easy for a computer to calculate, and it results in 187. However, suppose you were given a number that is the result of 2 prime numbers - such as 299. How do you find out which 2 prime numbers multipled together to create 299?
```
marvinchin: I wonder if such a detailed example is necessary to get the idea of hashing across. I feel like it introduces a significant amount of complexity to the article, while not providing much additional value to the reader.

amrut-prabhu:
@ChooJeremy I'll review it later today

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,98 @@
...
+
+One question that is commonly asked by developers is where to store the salt. The salt can be stored in plaintext, along with the user in the database. Since the goal of the salt is only to prevent precomputed rainbow tables from being used, it doesn't need to be encrypted in the database.
+
+### If an attacker has gained access to the entire application, what does it matter if my passwords are stored in plaintext or not?
+
+If an attacker has already gained access to the entire application, then he already has all the information hat he coculd possibly get. He would have access to all of your application's data, including data from your users or from any analytics software that you might be running. However, by adding salt and hashing your passwords, the attacker still doesn't know your customer's passwords and could take years to find out. Otherwise, since 59% of people use the same password across multiple sites<sup>1</sup>, the attacker could quickly try other websites such as banks to attempt to break into those accounts, which can yield great financial return.
```
marvinchin: Linking to the source for the statistic here might be good 🙂 

Also, do you think that the last line should be phrased to emphasise the detriment to the user, rather than the benefit to the attacker?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,98 @@
...
+
+Additionally, when a user signed up on your website and provided you a password, they implicitly trusted you to keep that information safe and secure for them, and in a sense you have a responsibility to keep their passwords secure as well. By doing proper password storage, if your servers ever get breached, you can assure your customers that their passwords is properly secured and maintain some of their trust in you.
```
marvinchin: I believe that it should be "you can assure your customers that their passwords *are* properly secured" instead of "is".

marvinchin (commented):
Just a couple of comments 🙂

contents/security/passwordStorage.md
ChooJeremy: Is there a way to link it nicely in markbind? The way I did it here was adding a superscript 1 and having the source on the very bottom. I suppose I could always write one up with HTML and CSS.

contents/security/passwordStorage.md
ChooJeremy: Good point! I'll let others review/comment on it before deciding on what to do with it. Without the example I felt like it was hard to bring across the point of what hashing meant.

contents/security/passwordStorage.md
amrut-prabhu: I guess the simplest way would be to just use a markdown hyperlink?

`since 59% of people `**`are reported`**` to use the same password across multiple sites according to [Source](link_to_source)`

I haven't tried it, but you can alternatively embed a hyperlink in the superscript 1 itself.

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,106 @@
...
+
+<div class="website-content">
+
+# Introduction to Password Storage
+
+Author: [Jeremy Choo](https://github.com/ChooJeremy)
+
+## Overview
+
+While developing your own application, it is very common to implement login systems to be able to identify users. Whenever there is a need to store user-specific information, developers typically implement user accounts to identify these users separately. You might have seen ways this are done in other websites - using registration forms and requring you to login with a username and password to identify you for the session.
```
amrut-prabhu: >seen ways this **are** done **in** other websites  

Grammatical error. Use `is` and `on` instead?

And I'd suggest that you rephrase it- are you trying to say that there are multiple different ways, or that it is done this way on other websites

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,106 @@
...
+
+As a developer, storing these usernames and passwords are essential to be able to refer to them later. While there are many ways to store these data - from text files to databases - this article will focus more on how to store passwords properly. For the purpose of this article, we will assume that you are storing the passwords in a database. Modern password mechanism typically involve **salting** and **hashing** the password. This article will describe what these mean and why they are important.
```
amrut-prabhu: I think `store this data` sounds better, in common usage 😅

amrut-prabhu: Modern password **storage** mechanism**s**?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,106 @@
...
+
+### What's wrong with <trigger for="pop:plaintext">plaintext</trigger>?
+
+<popover id="pop:plaintext" title="_Plaintext_ refers to unencrypted information" placement="top">
+  <div slot="content">
+It is a cryptography term generally referring to text before encyption or after decrypting it. Another term for it is cleartext.
+  </div>
+</popover>
+
+While usernames are typically public information, thus <trigger for="pop:encrypt">encrypting</trigger> them is not as important, passwords are typically secret information and require encryption. If passwords are stored in plaintext and your server's database is ever compromised, attackers can easily use the stored plaintext passwords to login as the user and potentially access other functionality. This grants the attacker more power than just reading values off the database and being unable to read them if they were encrypted.
```
amrut-prabhu: I think you can rephrase it:

> Since usernames are typically public information, encrypting them is not as important. However, passwords are typically secret information and hence, require encryption.

amrut-prabhu: database is ~~ever~~ compromised

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,106 @@
...
+
+<popover id="pop:encrypt" title="Encryption refers to the process of changing a message such that it becomes essentially random text." placement="top">
+  <div slot="content">
+ The opposite process is called decryption, which converts the text back into the original message.
+  </div>
+</popover>
+
+In addition,
+* Your employees can see the passwords of your users. If your employees are ever compromised, they might be willing to simply leak the password.
```
amrut-prabhu: can see the **plaintext** passwords

willing to leak the password**s**

Also, why did you decide to use a list here, when the previous related point is a prose? How about you use 2 paragraphs instead?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,106 @@
...
+* It is common for users to re-use the same password on multiple sites. If your passwords are stored as plaintext and accessed, attackers can easily re-use the same passwords on other websites (such as banking websites) and gain access to those.
+
+To prevent all of these measures, it is good to perform hashing on passwords. Unlike encryption, hashing is a <trigger for="pop:oneway">one-way function</trigger>. This means that an attacker cannot decrypt the encrypted password to retrieve the original password, thus ensuring that all of the above vulnerabilities isn't possible.
```
amrut-prabhu: I don't think `measures` is the right word here. Maybe, `prevent any of these situations from occurring, it` instead?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,106 @@
...
+
+Additionally, when a user signed up on your website and provided you a password, they implicitly trusted you to keep that information safe and secure for them, and in a sense you have a responsibility to keep their passwords secure as well. By doing proper password storage, if your servers ever get breached, you can assure your customers that their passwords are properly secured and maintain some of their trust in you.
+
+## Word of warning
+
+Reading this article doesn't make you an expert in writing your own password hashing code. It is far too easy to screw up and make a mistake. Instead, use one of the free libraries that provide a crypto function that has already been well-tested by the community. **Do not write your own crypto library.** For example, here are some libraries you can use instead depending on your language:
```
amrut-prabhu: ... instead`,` depending on ...

contents/security/passwordStorage.md
amrut-prabhu: ensuring that `none of the aforementioned vulnerabilities are possible`?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,106 @@
...
+## What is hashing?
+
+Hashing is the act of transforming a set of data into another set of data that is difficult to reverse. As an example, suppose you wanted to multiply 2 prime numbers together, such as 17 and 11. This is easy for a computer to calculate, and it results in 187. However, suppose you were given a number that is the result of 2 prime numbers - such as 299. How do you find out which 2 prime numbers multipled together to create 299?
+
+It definitely isn't 2, because the number isn't even.  
+It isn't 3, because the digits don't sum to a multiple of 3.  
+It isn't 5, because it doesn't end in a 5 or a 0.  
+Is it 7? Well, it's hard to tell, you would have to start doing division to test.  
+ 
+There are a few shortcuts that computers and mathematicans can use to make the operation faster, but overall it is a long process that isn't efficient at all, particularly for huge numbers. Thus, it is easy to perform the function one way, but hard to reverse.
```
amrut-prabhu: > easy to perform the function one way, but hard to reverse

Consider making this bold?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,106 @@
...
+Hashing is the act of transforming a set of data into another set of data that is difficult to reverse. As an example, suppose you wanted to multiply 2 prime numbers together, such as 17 and 11. This is easy for a computer to calculate, and it results in 187. However, suppose you were given a number that is the result of 2 prime numbers - such as 299. How do you find out which 2 prime numbers multipled together to create 299?
+
+It definitely isn't 2, because the number isn't even.  
+It isn't 3, because the digits don't sum to a multiple of 3.  
+It isn't 5, because it doesn't end in a 5 or a 0.  
+Is it 7? Well, it's hard to tell, you would have to start doing division to test.  
+ 
+There are a few shortcuts that computers and mathematicans can use to make the operation faster, but overall it is a long process that isn't efficient at all, particularly for huge numbers. Thus, it is easy to perform the function one way, but hard to reverse. 
+
+Computers use a similar forumla to manipulate the password to another random string, one that is easy to generate from the original, but would take an extremely long time to generate the original from the hash.
```
amrut-prabhu: `to convert passwords to random strings` instead?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,106 @@
...
+
+Some examples of hashing formulas are [MD5](https://en.wikipedia.org/wiki/MD5) and [SHA256](https://en.wikipedia.org/wiki/SHA-2).
```
amrut-prabhu: Since you stated `hashing formulas`, consider linking to the actual formula/algorithm instead of a wikipedia article. After all, this article is meant for beginners and we want to introduce them to good resources

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Why isn't hashing enough?
+
+Unfortunately, [rainbow tables](https://en.wikipedia.org/wiki/Rainbow_table) exist.
+
+A rainbow table is a precomputed table of hashes for some set of passwords. Basically, since the hashing function is easy to compute one-way, people build huge tables of hashes wherein the plaintext is already known, so that attempting to crack hashes becomes a simple problem of looking up the hash in the table and it's corresponding value, instead of attempting to reverse the hash. Because of this, it is very easy to crack simple hashes by simply doing a lookup.
+
+An example of a service that provides this is [Crackstation](https://crackstation.net/).
+
+Since attackcs like rainbow tables exist, passwords need another layer of security.
```
amrut-prabhu: `attackcs` typo

Also, would rainbow table be considered an "attack"?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,106 @@
...
+
+One question that is commonly asked by developers is where to store the salt. The salt can be stored in plaintext, along with the user in the database. Since the goal of the salt is only to prevent precomputed rainbow tables from being used, it doesn't need to be encrypted in the database.
+
+### If an attacker has gained access to the entire application, what does it matter if my passwords are stored in plaintext or not?
+
+If an attacker has already gained access to the entire application, then he already has all the information hat he coculd possibly get. He would have access to all of your application's data, including data from your users or from any analytics software that you might be running. However, by adding salt and hashing your passwords, the attacker still doesn't know your customer's passwords and could take years to find out. Otherwise, since 59% of people use the same password across multiple sites<sup>1</sup>, the attacker could quickly try other websites such as banks to attempt to break into those accounts, which can yield great financial return.
```
amrut-prabhu: > all the information hat he coculd possibly get

Typos

And, `possibly want` instead?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,106 @@
...
+
+Additionally, when a user signed up on your website and provided you a password, they implicitly trusted you to keep that information safe and secure for them, and in a sense you have a responsibility to keep their passwords secure as well. By doing proper password storage, if your servers ever get breached, you can assure your customers that their passwords are properly secured and maintain some of their trust in you.
```
amrut-prabhu: I think this sentence is too long and would be better if you start a new one from `, and in a sense ...`

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,106 @@
...
+
+An example of a service that provides this is [Crackstation](https://crackstation.net/).
+
+Since attackcs like rainbow tables exist, passwords need another layer of security.
+
+## Adding salt
+
+A salt refers to appending a string of text, unique to each user, to their password before hashing them. Since each user has a unique salt, this makes rainbow tables ineffective, as the majority of the precomputed hashes won't even contain the salt, so they wouldn't even matter anyway! In this way, the salt forces the attacker to recompute the rainbow table for each password in order to be able to effectively use it. This effectively converts the attack to brute force, as each hash must be recomputed for each possible password. Additionally, the computed rainbow table would only be useful for that specific user, as each user would have a different salt. It could take years before a password is cracked!
```
amrut-prabhu: Put a tooltip for `brute force attack`?

contents/security/passwordStorage.md
amrut-prabhu: Personally, I feel that this comes off a bit strong and aggressive (but, I guess that's the point) 😅 

You can put this into a `Getting Started` section instead

It may be better to just recommend these to the reader instead of telling them to not try to do it themselves. At the end of the day, these libraries still have to be made by someone

amrut-prabhu (changes requested):
Caught a few typos and added some recommendations

contents/security/passwordStorage.md
marvinchin: I do feel that it might be appropriate in this case though, since one of the most common advice (and for good reason) regarding security is "don't roll your own crypto". 😅 

Do you feel like putting this in a warning box might make this point stand out more to readers?

contents/security/passwordStorage.md
ChooJeremy: I think that for the purposes of this article, it should be ok to consider it an attack.

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,120 @@
...
+
+### What's wrong with <trigger for="pop:plaintext">plaintext</trigger>?
+
+<popover id="pop:plaintext" title="_Plaintext_ refers to unencrypted information" placement="top">
+  <div slot="content">
+It is a cryptography term generally referring to text before encyption or after decrypting it. Another term for it is cleartext.
+  </div>
+</popover>
+
+Since usernames are typically public information, <trigger for="pop:encrypt">encrypting</trigger> them is not as important. However, passwords are typically secret information and hence, require encryption.
```
Xenonym: Hashing != encryption, since hashing is theoretically irreversible. Since this article is explicitly about hashing and related cryptography concepts, we need to be especially careful with terminology. 

Would the word "hide" be less problematic here? If that is too vague, perhaps `obfuscated` is a more clunky but technically correct word.

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,120 @@
...
+Since usernames are typically public information, <trigger for="pop:encrypt">encrypting</trigger> them is not as important. However, passwords are typically secret information and hence, require encryption. 
+
+<popover id="pop:encrypt" title="Encryption refers to the process of changing a message such that it becomes essentially random text." placement="top">
+  <div slot="content">
+ The opposite process is called decryption, which converts the text back into the original message.
+  </div>
+</popover>
+
+If passwords are stored in plaintext:
+* Should your database be compromised, attackers can easily use the stored plaintext passwords to login as the user and potentially access other functionality. This grants the attacker more power than just reading values off the database and being unable to read them if they were encrypted.
```
Xenonym: `encrypted` -> `hidden`? (see previous comment)

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,120 @@
...
+* Your employees can see the plaintext passwords of your users. If your employees are ever compromised, they might be willing to simply leak the password.
```
Xenonym: `leak the password`**`s`**?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,120 @@
...
+* It is common for users to re-use the same password on multiple sites. If your passwords are stored as plaintext and accessed, attackers can easily re-use the same passwords on other websites (such as banking websites) and gain access to those.
```
Xenonym: `other websites `**`with sensitive data`**` (such as banking websites)`?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,120 @@
...
+
+To prevent these situations from occuring, it is good to perform hashing on passwords. Unlike encryption, hashing is a <trigger for="pop:oneway">one-way function</trigger>. This means that an attacker cannot decrypt the encrypted password to retrieve the original password, thus ensuring that none of the aforementioned vulnerabilities are possible.
```
Xenonym: With reference to the first comment, here you mention that hashing is _not_ encryption, despite mentioning passwords should be encrypted earlier.

contents/security/passwordStorage.md
Xenonym: Would it be easier to just show a hash of a large file (say, [the works of Shakespeare](http://www.gutenberg.org/files/100/100-0.txt)), and explain that hashing is irreversible because you can't reconstruct the file itself from the hash?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,120 @@
...
+It definitely isn't 2, because the number isn't even.  
+It isn't 3, because the digits don't sum to a multiple of 3.  
+It isn't 5, because it doesn't end in a 5 or a 0.  
+Is it 7? Well, it's hard to tell, you would have to start doing division to test.  
+ 
+There are a few shortcuts that computers and mathematicans can use to make the operation faster, but overall it is a long process that isn't efficient at all, particularly for huge numbers. Thus, **it is easy to perform the function one way, but hard to reverse**.
+
+Computers use a similar forumla to convert passwords to random strings, one that is easy to generate from the original, but would take an extremely long time to generate the original from the hash.
+
+Some examples of hashing formulas are [MD5](https://www.quora.com/How-does-the-MD5-algorithm-work) and [SHA1](https://deadhacker.com/2006/02/21/sha-1-illustrated/).
```
Xenonym: 1. `hashing functions` -> `hash functions`?
2. I think it's important that we use examples of _password_ hashing algorithms here (eg. `Argon2`/`scrypt`/`bcrypt`/`PBKDF2`) and explain why password hashing algorithms are different from _cryptographic_ hashing algorithms (password hashing algorithms are deliberately slow to prevent brute force compared to cryptographic algorithms which are designed for speed).

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,120 @@
...
+
+An example of a service that provides this is [Crackstation](https://crackstation.net/).
+
+Since attacks like rainbow tables exist, passwords need another layer of security.
+
+## Adding salt
+
+A salt refers to appending a string of text, unique to each user, to their password before hashing them. Since each user has a unique salt, this makes rainbow tables ineffective, as the majority of the precomputed hashes won't even contain the salt, so they wouldn't even matter anyway! In this way, the salt forces the attacker to recompute the rainbow table for each password in order to be able to effectively use it. This effectively converts the attack to <trigger for="pop:brute">brute force</trigger>, as each hash must be recomputed for each possible password. Additionally, the computed rainbow table would only be useful for that specific user, as each user would have a different salt. It could take years before a password is cracked!
```
Xenonym: 1. I think it's important to point out that the salt should be _random_, otherwise unknowing developers may use the username as the salt (the username is technically unique to each user after all). This would defeat the purpose of salting, since common usernames can be "rainbow-tabled".
2. We should point out that many password hashing algorithms have built-in salts (`Argon2`/`scrypt`/`bcrypt`) and a good password hashing library and algo will salt automatically.

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,120 @@
...
+## Getting started
+
+<p style="color: #721c24; background-color: #f8d7da; border-color: #f5c6cb; padding: .75rem 1.25rem; border: 1px solid transparent; border-radius: .25rem;">
+Reading this article doesn't make you an expert in writing your own password hashing code. It is far too easy to screw up and make a mistake. Instead, use one of the free libraries that provide a crypto function that has already been well-tested by the community. <b>Do not write your own crypto library.</b>
+</p>
+
+Here are some libraries you can use to implement password storage, depending on your language:
+* If you're using PHP, the [password_hash](https://secure.php.net/manual/en/function.password-hash.php) function will do the job for you. 
+* If you're using java, use [SecretKeyFactory](https://www.owasp.org/index.php/Hashing_Java) to apply hashes with SHA256.
+* If you're using Python, use [hashlib](https://docs.python.org/3/library/hashlib.html) to salt and hash your password. [Example usage](https://stackoverflow.com/questions/9594125/salt-and-hash-a-password-in-python)
```
Xenonym: [`passlib`](https://passlib.readthedocs.io/en/stable/) may be a better recommendation for this.

Xenonym (commented):
I enjoyed this article, was clear and well-explained! Some comments for consideration :)

contents/security/passwordStorage.md
ChooJeremy: Good idea!

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,121 @@
...
+
+# Introduction to Password Storage
+
+Author: [Jeremy Choo](https://github.com/ChooJeremy) <br>
+Reviewers: [Amrut Prabhu](https://github.com/amrut-prabhu), [Marvin Chin](https://github.com/marvinchin), [Tan Zhen Yong](https://github.com/Xenonym)
+
+## Overview
+
+While developing your own application, it is very common to implement login systems to be able to identify users. Whenever there is a need to store user-specific information, developers typically implement user accounts to identify these users separately. You might have seen this mechanism implemented on other websites - using registration forms and requring you to login with a username and password to identify you for the <trigger for="pop:session">session</trigger>.
+
```
junming403, line 26: Can we provide a link some resources that explains session in more detail in case the reader wonder?

junming403 (approved):
Good work in general, I liked the way you put sticky notes for term that you feel reader might not understand. A very useful article that I really enjoyed reading! Just a small suggestion for you to consider.

ChooJeremy:
Thanks for the reviews! @amrut-prabhu @marvinchin @Xenonym would you guys mind giving the article a once-over and giving a LGTM if you feel it's good to go?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,118 @@
...
+<div class="website-content">
+
+# Introduction to Password Storage
+
+Author: [Jeremy Choo](https://github.com/ChooJeremy) <br>
+Reviewers: [Amrut Prabhu](https://github.com/amrut-prabhu), [Marvin Chin](https://github.com/marvinchin), [Tan Zhen Yong](https://github.com/Xenonym), [Wang Junming](https://github.com/junming403)
+
+## Overview
+
+While developing your own application, it is very common to implement login systems to be able to identify users. Whenever there is a need to store user-specific information, developers typically implement user accounts to identify these users separately. You might have seen this mechanism implemented on other websites - using registration forms and requring you to login with a username and password to identify you for the <trigger for="pop:session"><a href="https://support.google.com/analytics/answer/2731565?hl=en">session</a></trigger>.
```
amrut-prabhu: Typo in `requring `

amrut-prabhu: `log in` instead of `login`

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,118 @@
...
+
+<popover id="pop:session" title="A _session_  is used to broadly describe a user's visit" placement="top">
+  <div slot="content">
+Generally, a session starts when a user visits your page, and ends when the user closes the browser window.
```
amrut-prabhu: I'm not sure about this; just pointing out for your consideration:
Does a session end when a user closes the `browser window`, or the `browser tab` itself?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,118 @@
...
+  </div>
+</popover>
+
+As a developer, storing these usernames and passwords are essential to be able to refer to them later. While there are many ways to store this data - from text files to databases - this article will focus more on how to store passwords properly. For the purpose of this article, we will assume that you are storing the passwords in a database. Modern password storage mechanisms typically involve **salting** and **hashing** the password. This article will describe what these mean and why they are important.
+
+### What's wrong with <trigger for="pop:plaintext">plaintext</trigger>?
```
amrut-prabhu: Just a nit:
Would it be better to leave the title clean, and include this popover in the next sentence (`... to store them in plaintext`) instead?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,118 @@
...
+
+<popover id="pop:plaintext" title="_Plaintext_ refers to unencrypted information" placement="top">
+  <div slot="content">
+It is a cryptography term generally referring to text before encyption or after decrypting it. Another term for it is cleartext.
+  </div>
+</popover>
+
+Since usernames are typically public information, it is OK to store them in plaintext. However, passwords are typically secret information and hence, cannot be stored in plaintext.
```
amrut-prabhu: ~cannot~ should not/must not

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,118 @@
...
+
+If passwords are stored in plaintext:
+* Should your database be compromised, attackers can easily use the stored plaintext passwords to login as the user and potentially access other functionality. This grants the attacker more power than just reading values off the database. If the passwords aren't even in plaintext, the attacker might not be able to read the passwords.
```
amrut-prabhu: ~login~ log in

amrut-prabhu: This seems unnecessary
> ~This grants the attacker more power than just reading values off the database.~ If the passwords aren't even in plaintext, the attacker might not be able to read the passwords.

Replace with this instead?
> If the passwords weren't in plaintext, the attacker might not have been able to even decipher what the actual passwords were.

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,118 @@
...
+* Your employees can see the plaintext passwords of your users. If your employees are ever compromised, they might be willing to simply leak the passwords.
+* It is common for users to re-use the same password on multiple sites. If your passwords are stored as plaintext and accessed, attackers can easily re-use the same passwords on other websites with sensitive data (such as banking websites) and gain access to those.
```
amrut-prabhu: > If your passwords are stored as plaintext

This is repetitive as it's the same as the list description above

Adding to that, your list items should be a logical extension to the list description. The first two bullets seem fine, but the beginning of this one does not fit.

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,118 @@
...
+<div slot="content">
+	A one way function is a function that is easy to compute in one way, but is close to impossible to compute the other way.
+</div>
+</popover>
+
+## What is hashing?
+
+Hashing is the act of transforming a set of data into another set of data that is impossible to reverse. As an example, suppose you wanted to hash [The Complete Works of Shakespeare](https://www.gutenberg.org/files/100/100-0.txt). If you performed a SHA1 hash on the entire contents of that file, including the licensing information at the very end, you'll end up hashing the entire 157257 lines of that file into `C9B38187EA0CC82B72142B92034DA859A3550B57`. Obviously, it is impossible to condense all 157257 lines of text into just those 40 characters. 
+
+This means that is is impossible to recreate the complete works of Shakespeare from just the hash result. Thus, we say that a hashing function is a one-way function, as it can only be performed in a one direction, and it would be impossible to perform it in the other direction.
```
amrut-prabhu: >  impossible to perform it in the other direction

`impossible to reverse the process` instead? Or something along the lines of it not having an inverse function

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,118 @@
...
+
+<popover id="pop:oneway" title="" placement="top">
+<div slot="content">
+	A one way function is a function that is easy to compute in one way, but is close to impossible to compute the other way.
+</div>
+</popover>
+
+## What is hashing?
+
+Hashing is the act of transforming a set of data into another set of data that is impossible to reverse. As an example, suppose you wanted to hash [The Complete Works of Shakespeare](https://www.gutenberg.org/files/100/100-0.txt). If you performed a SHA1 hash on the entire contents of that file, including the licensing information at the very end, you'll end up hashing the entire 157257 lines of that file into `C9B38187EA0CC82B72142B92034DA859A3550B57`. Obviously, it is impossible to condense all 157257 lines of text into just those 40 characters.
```
amrut-prabhu: > performed a SHA1 hash

`performed hashing using  SHA1` or `used the SHA1 hash function`

amrut-prabhu: > Obviously, it is impossible to condense all 157257 lines of text into just those 40 characters. 

I feel that this should be in the next paragraph

amrut-prabhu: The link redirects to http://www.gutenberg.org/ebooks/100?msg=welcome_stranger when it is clicked for the first time  :no_mouth:

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,118 @@
...
+
+This means that is is impossible to recreate the complete works of Shakespeare from just the hash result. Thus, we say that a hashing function is a one-way function, as it can only be performed in a one direction, and it would be impossible to perform it in the other direction.
+
+Some examples of cryptographic hashing algorithms are [MD5](https://www.quora.com/How-does-the-MD5-algorithm-work) and [SHA1](https://deadhacker.com/2006/02/21/sha-1-illustrated/). However, when you perform hashing for passwords, you should use password hashing algorithms (such as [Argon2](https://github.com/P-H-C/phc-winner-argon2), [SCrypt](https://passlib.readthedocs.io/en/stable/lib/passlib.hash.scrypt.html) and [bcrypt](https://security.stackexchange.com/questions/4781/do-any-security-experts-recommend-bcrypt-for-password-storage)). These password hashing algorithms are designed to be slow to prevent brute force attacks, unlike cryptographic hashing algorithms which are built for speed. Despite that, if you plan on learning more about hashing algorithms, I recommend starting with learning about MD5 and SHA1, as they are easier algorithms to start learning with.
```
amrut-prabhu: > starting with learning about MD5

starting with MD5

> as they are easier algorithms to start learning with

as they are easier algorithms to learn about

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Why isn't hashing enough?
+
+Unfortunately, [rainbow tables](https://en.wikipedia.org/wiki/Rainbow_table) exist.
```
amrut-prabhu: Doesn't match the title

`Because rainbow tables exist, unfortunately`?

contents/security/passwordStorage.md
Xenonym: `the attacker might not be able to read the passwords` sounds a bit weak. `should not be able`?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,118 @@
...
+
+A rainbow table is a precomputed table of hashes for some set of passwords. Basically, since the hashing function is easy to compute one-way, people build huge tables of hashes wherein the plaintext is already known, so that attempting to crack hashes becomes a simple problem of looking up the hash in the table and it's corresponding value, instead of attempting to reverse the hash. Because of this, it is very easy to crack simple hashes by simply doing a lookup.
```
amrut-prabhu: I think this is the first instance of `hashes`. Mention that it means `hash results` in a tooltip?

amrut-prabhu: >  and it's corresponding value

its*

contents/security/passwordStorage.md
Xenonym: Also: only this text file will hash to `C9B38187EA0CC82B72142B92034DA859A3550B57`. Another text file, or even changing one character in this file, will result in a dramatically different hash. Otherwise, the reader might think that other files (or passwords!) could also produce this value.

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Getting started
+
+<p style="color: #721c24; background-color: #f8d7da; border-color: #f5c6cb; padding: .75rem 1.25rem; border: 1px solid transparent; border-radius: .25rem;">
+Reading this article doesn't make you an expert in writing your own password hashing code. It is far too easy to screw up and make a mistake. Instead, use one of the free libraries that provide a crypto function that has already been well-tested by the community. <b>Do not write your own crypto library.</b>
+</p>
+
+Here are some libraries you can use to implement password storage, depending on your language:
+* If you're using PHP, the [password_hash](https://secure.php.net/manual/en/function.password-hash.php) function will do the job for you. 
+* If you're using java, use [SecretKeyFactory](https://www.owasp.org/index.php/Hashing_Java) to apply hashes with SHA256.
```
amrut-prabhu: Capitalise Java

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,118 @@
...
+* If you're using Python, use [PassLib](https://passlib.readthedocs.io/en/stable/narr/hash-tutorial.html).
+* If you're using npm, use [bcrypt](https://www.npmjs.com/package/bcrypt).
```
amrut-prabhu: Node.js (npm) ?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Other resources
+
+* [Secure Salted Password Hashing - How to do it properly](https://crackstation.net/hashing-security.htm) is an excellent resource that explains how to perform salted password hashing correctly, including links to other good libraries and what else can be done.
+* [Awesome Cryptography](https://github.com/sobolevn/awesome-cryptography) is a curated list of resources - articles, blogs, books, libraries and more.
```
amrut-prabhu: resources about what?

contents/security/passwordStorage.md
Xenonym: More generally, `using JavaScript`?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,118 @@
...
+
+<popover id="pop:brute" title="A brute force attack is an attack where all possible combinations are tested to see if they work." placement="top">
+  <div slot="content">
+ A brute force attack usually takes very long to carry out.
+  </div>
+</popover>
+
+Note that the salt should be randomly generated, as opposed to choosing a static value that is different for every user. For example, if an application used the username of the user as the salt, then attackers can pre-generate rainbow tables for common usernames, causing users with common users to be vulnerable to a rainbow table attack, even with salt applied to their passwords.
+
+Thus, the way to store passwords properly is to use a salted hash - take the user's password, append some bytes to it, and hash the result. That hash is the user's password. When the user attempts to login again, perform the same procedure again. If the hash matches, then you know that the user is who he says he is, even if you don't actually know the original password that the user provided.
```
amrut-prabhu: > That hash is the user's password

`The result is the user's hashed password`?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,118 @@
...
+
+Thus, the way to store passwords properly is to use a salted hash - take the user's password, append some bytes to it, and hash the result. That hash is the user's password. When the user attempts to login again, perform the same procedure again. If the hash matches, then you know that the user is who he says he is, even if you don't actually know the original password that the user provided. 
+
+One question that is commonly asked by developers is where to store the salt. The salt can be stored in plaintext, along with the user in the database. Since the goal of the salt is only to prevent precomputed rainbow tables from being used, it doesn't need to be encrypted in the database.
+
+Many good password hashing algorithms today have built-in salts, such as [Argon2](https://github.com/P-H-C/phc-winner-argon2), [SCrypt](https://passlib.readthedocs.io/en/stable/lib/passlib.hash.scrypt.html) and [bcrypt](https://security.stackexchange.com/questions/4781/do-any-security-experts-recommend-bcrypt-for-password-storage). A good password hashing library and algorithm will salt automatically.
```
amrut-prabhu: >  hashing library and algorithm will

hashing algorithm **or** library will

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,118 @@
...
+
+<popover id="pop:encrypt" title="Encryption refers to the process of changing a message such that it becomes essentially random text." placement="top">
+  <div slot="content">
+ The opposite process is called decryption, which converts the text back into the original message.
+  </div>
+</popover>
+
+<popover id="pop:oneway" title="" placement="top">
+<div slot="content">
+	A one way function is a function that is easy to compute in one way, but is close to impossible to compute the other way.
```
Xenonym: It's not clear what is the `one way` and what is the `other way`. Perhaps `easy to compute the result of, but whose results are difficult to reverse back to the original input`?

Xenonym (commented):
Just some more minor points!

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,118 @@
...
+
+### If an attacker has gained access to the entire application, what does it matter if my passwords are stored in plaintext or not?
+
+If an attacker has already gained access to the entire application, then he already has all the information that he could possibly want from your server. He would have access to all of your application's data, including data from your users or from any analytics software that you might be running. However, by adding salt and hashing your passwords, the attacker still doesn't know your customer's passwords and could take years to find out. Otherwise, since 59% of people use the same password across multiple sites <sup>[source](https://securityboulevard.com/2018/05/59-of-people-use-the-same-password-everywhere-poll-finds/)</sup>, the attacker could quickly try other websites such as banks to attempt to break into those accounts, which can yield great financial return.
```
amrut-prabhu: > try other websites such as banks to attempt to break into those accounts, which can yield great financial return.

try breaking into those accounts on other websites (such as banking websites), which can potentially yield great returns in terms of information and/or money.

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,118 @@
...
+
+Additionally, when a user signed up on your website and provided you a password, they implicitly trusted you to keep that information safe and secure for them. In a sense, you have a responsibility to keep their passwords secure as well. By doing proper password storage, if your servers ever get breached, you can assure your customers that their passwords are properly secured and maintain some of their trust in you.
```
amrut-prabhu: > In a sense, you have a responsibility to keep their passwords secure ~as well~.

In a sense, you **do** have a responsibility to keep their passwords secure.

amrut-prabhu: > properly secured and maintain 

properly secured **,** and **still** maintain

amrut-prabhu (commented):
Just a couple of suggestions (mostly nits) for your consideration.

Also, a question I have regarding creating our own hash functions and rainbow hashes is `Why can't I just use some random function (not a randomised function) that no one else knows about?`

contents/security/passwordStorage.md
ChooJeremy: Generally speaking it should be the browser window. If, for example, you login into IVLE (without checking the remember me box) and then close the tab and reopen a new tab for IVLE, it will keep you logged in.

contents/security/passwordStorage.md
amrut-prabhu: Oh, sorry. Should have been clearer. Change to `log in` in `requiring you to log in` (verb form).  
The first occurrence should remain `common to implement login systems` since it's a noun.

amrut-prabhu (approved):

amrut-prabhu:
Nice work on the article! I've managed to learn something new about security, and now know recommended resources that I can get started with

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,126 @@
...
+  </div>
+</popover>
+
+If passwords are stored in plaintext:
+* Should your database be compromised, attackers can easily use the stored plaintext passwords to log in as the user and potentially access other functionality. If the passwords weren't in plaintext, the attacker should not be able to decipher what the actual passwords are.
+* Your employees can see the plaintext passwords of your users. If your employees are ever compromised, they might be willing to simply leak the passwords.
+* Attackers can try to re-use the same passwords on other websites with sensitive data (such as banking websites) and gain access to those. Since it is common for users to re-use the same password on multiple sites, attackers can potentially be successful here and gain access to more sensitive data.
+
+To prevent these situations from occuring, it is good to perform hashing on passwords. Unlike <trigger for="pop:encrypt">encryption</trigger>, hashing is a <trigger for="pop:oneway">one-way function</trigger>. This means that an attacker cannot decrypt the encrypted password to retrieve the original password, thus ensuring that none of the aforementioned vulnerabilities are possible.
```
Xenonym: `decrypt the encrypted password` -> `reverse the hashed password`?

Xenonym (approved):
LGTM! 👍 Will be a good resource for learning about security best practices.

marvinchin (approved):
LGTM 🌈 🌤

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,126 @@
...
+  pageNav: 3
+</frontmatter>
+
+{{ navbar | safe }}
+
+<div class="website-content">
+
+# Introduction to Password Storage
+
+Author: [Jeremy Choo](https://github.com/ChooJeremy) <br>
```
damithc: make this line bold

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,126 @@
...
+Reviewers: [Amrut Prabhu](https://github.com/amrut-prabhu), [Marvin Chin](https://github.com/marvinchin), [Tan Zhen Yong](https://github.com/Xenonym), [Wang Junming](https://github.com/junming403)
+
+## Overview
+
+While developing your own application, it is very common to implement login systems to be able to identify users. Whenever there is a need to store user-specific information, developers typically implement user accounts to identify these users separately. You might have seen this mechanism implemented on other websites - using registration forms and requiring you to log in with a username and password to identify you for the <trigger for="pop:session"><a href="https://support.google.com/analytics/answer/2731565?hl=en">session</a></trigger>.
```
damithc: In the first sentence, why the need for _your own_?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,126 @@
...
+  <div slot="content">
+ The opposite process is called decryption, which converts the text back into the original message.
+  </div>
+</popover>
+
+<popover id="pop:oneway" title="" placement="top">
+<div slot="content">
+	A one way function is a function that is easy to compute the result of, but whose results are difficult to reverse back to the original input.
+</div>
+</popover>
```
damithc: Consider shortening the intro given so far as the reader would want to get to the good stuff quickly without beating about the bush. We can assume the ready has a basic understanding of user accounts, usernames, passwords, and the need to store them somewhere. For example, how about something like this?

>Many software applications use _username + password_  as user account credentials to authenticate users. Obviously, it is not a good idea for the software to store these credentials as plain text because if someone else were to gain access to them either lawfully (e.g., an employee who has access to the data) or unlawfully (e.g., someone hacking into the data storage), that person can use those credentials to impersonate the account owner. This article explains some techniques that can be used to store user credentials more securely:
>1. Encrypting
>2. Hashing
>3. Salting

Adding the encryption to the list might make the article more complete. What do you think?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,126 @@
...
+
+## What is hashing?
+
+Hashing is the act of transforming a set of data into another set of data that is impossible to reverse. As an example, suppose you wanted to hash [the entire play of Romeo and Juliet](http://shakespeare.mit.edu/romeo_juliet/full.html). If you performed hashing using SHA1 on the entire contents of that file, excluding the header, you'll end up hashing the entire 6048 lines of that file into `7A52CDA9F04051B0D8263AA61A42531DD27FCE92`. 
+
```
damithc, line 71: I'm not sure if bringing in the R&J play is necessary. Not that related to passwords

damithc, line 71: Unexplained term SHA

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,126 @@
...
+Obviously, it is impossible to condense all 6048 lines of text into just those 40 characters. This means that is is impossible to recreate the entire play of Romeo and Juliet from just the hash result. Thus, we say that a hashing function is a one-way function, as it can only be performed in a one direction, and it would be impossible to reverse the process. 
+
+Additionally, due to how hashing algorithms work, even a small change in the file will result in a completely different hash. For example, if I were to change the first line from `ACT I` to `ACT IX`, it would change the hash to `D180AE51CEEFA8EF6D01F7FD54528AC21659433E`, completely different from the previous hash. This means that even very similar passwords that vary only by capitalization or a letter would still result in different <trigger for="pop:hashes">hashes</trigger>, so attackers can't compare the hash to try and figure out which users have similar passwords.
+
+<popover id="pop:hashes" title="" placement="top">
+<div slot="content">
+  Hashes are the hash result of a hashing function.
+</div>
+</popover>
```
damithc, line 84: MarkBind tooltips are simpler to use than popovers.

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,126 @@
...
+
+Some examples of cryptographic hashing algorithms are [MD5](https://www.quora.com/How-does-the-MD5-algorithm-work) and [SHA1](https://deadhacker.com/2006/02/21/sha-1-illustrated/). However, when you perform hashing for passwords, you should use password hashing algorithms (such as [Argon2](https://github.com/P-H-C/phc-winner-argon2), [SCrypt](https://passlib.readthedocs.io/en/stable/lib/passlib.hash.scrypt.html) and [bcrypt](https://security.stackexchange.com/questions/4781/do-any-security-experts-recommend-bcrypt-for-password-storage)). These password hashing algorithms are designed to be slow to prevent brute force attacks, unlike cryptographic hashing algorithms which are built for speed. Despite that, if you plan on learning more about hashing algorithms, I recommend starting with MD5 and SHA1, as they are easier algorithms to learn about.
```
damithc: The difference between _cryptographic hashing_ and _password hashing_ is not explained.

damithc: Similarly, terms like _brute force attack_ slow/fast algorithm are not explained. In fact, even the involvement of an algorithm is not explained.

damithc: It's best to use 'we' rather than 'I' as the article already involves more than one person. In the future, there might be more than one author too.

damithc (changes requested):
Good effort @ChooJeremy, given that this is a highly technical topic. Needs more work to make it accessible to the target audience. At present, some concepts that are likely to be known to the reader are explained in details while some concepts that are unlikely to be known to the target reader are not explained in enough details.

Reviewed the first part and added some comments. Revise the rest based on comments that you can expect from me, based on the current comments.

ChooJeremy:
Added explanations, a new section on encryption and removed the R&J play. Also did some revising on the next section and made sure there was no unclear terms. Would like another review.

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,113 @@
...
+
+<popover id="pop:oneway" title="" placement="top">
+<div slot="content">
+	A one way function is a function that is easy to compute the result of, but whose results are difficult to reverse back to the original input.
+</div>
+</popover>
+
+## What is encryption?
+
+Encryption refers to the process of changing a message such that it becomes essentially random text. The opposite process is called decryption, which converts the text back into the original message. Encryption might seem like a good idea because it transfoms the data into essentially random bytes, which prevents all of the problems with storing the data directly in plaintext. However, because encryption is <tooltip content="If a function is reversible, and it converts from x to y, then it can also convert from y back to x"> reversible</tooltip>, it is always possible to regain the original password from the <trigger for="pop:ciphertext">ciphertext</trigger>.
```
damithc: You can use italics to emphasize a new term being introduced. e.g., 
> The opposite process is called _decryption_, which ...

damithc: two almost similar phrases used to mean the same thing, and also in close proximity: `essentially random text` `essentially random bytes`

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,113 @@
...
+
+<popover id="pop:ciphertext" title="_Ciphertext_ refers to encrypted information" placement="top">
+  <div slot="content">
+It is a cryptography term generally referring to data after encrypting it.
+  </div>
+</popover>
+
+While encryption might seem suitable for storing passwords, it faces some severe problems. Since the data is encrypted, the encryption key must also be stored somewhere. This means that if someone manages to hack into the application and read the encrypted passwords, it is also likely that they will be able to read the encryptio key. With the encryption key, they will be able to decrypt all the passwords and read them anyway. This makes encryption unsuitable for password storage.
```
damithc: Wait, what is an _encryption key_? Not explained before.

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,113 @@
...
+
```
damithc, line 48: This section seems to say everything twice. Can try to explain in one pass?
* encryption turns data into cyphertext by combining it with a key
* decryption can get the original data from cypertext if the key is available
* hence encryption is not suitable for secure storage of passwords

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,113 @@
...
+## What is hashing?
+
+Hashing is the act of transforming a set of data into another set of data that is impossible to reverse. Additionally, due to how <trigger for="pop:hashing-algo">hashing algorithms</trigger> work, even a small change in the data will result in a completely different hash. This means that even very similar passwords that vary only by capitalization or a letter would still result in different <tooltip content="Hashes are the result of a hashing function">hashes</tooltip>, so attackers can't compare the hash to try and figure out which users have similar passwords.
```
damithc: > even a small change in the data will result in a completely different hash

This being mentioned here specifically seems to imply that encryption doesn't have this property. Is that right?

damithc: Hashing is like encryption but not reversible? How come? Provide some more information.

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,113 @@
...
+
+<popover id="pop:hashing-algo" title="A _hashing algorithm_ is a specific type of operation that hashes the input" placement="top">
+  <div slot="content">
+Different types of hashing algorithms will result in different output.
+  </div>
+</popover>
+
+Some examples of cryptographic hashing algorithms are [MD5](https://www.quora.com/How-does-the-MD5-algorithm-work) and [SHA1](https://deadhacker.com/2006/02/21/sha-1-illustrated/). However, when you perform hashing for passwords, you should use password hashing algorithms (such as [Argon2](https://github.com/P-H-C/phc-winner-argon2), [SCrypt](https://passlib.readthedocs.io/en/stable/lib/passlib.hash.scrypt.html) and [bcrypt](https://security.stackexchange.com/questions/4781/do-any-security-experts-recommend-bcrypt-for-password-storage)). The main difference between these password hashing algorithms are designed to be slow to prevent <trigger for="pop:brute">brute force</trigger> attacks, unlike cryptographic hashing algorithms which are built for speed. Despite that, if you plan on learning more about hashing algorithms, we recommend starting with MD5 and SHA1, as they are easier algorithms to learn about.
```
damithc: Something wrong with this sentence:  The main difference between these password hashing algorithms are designed to be slow to prevent ...

contents/security/passwordStorage.md
damithc, line 48: Can give a helloworld kind of example of a encrypting something. i.e., the data, the algorithm, and the output

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,113 @@
...
+
```
damithc, line 67: Similar to above, can give an example here?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,113 @@
...
+### Why isn't hashing enough?
+
+Because [rainbow tables](https://en.wikipedia.org/wiki/Rainbow_table) exist, unfortunately.
+
+A rainbow table is a precomputed table of hashes for some set of passwords. Basically, since the hashing function is easy to compute one-way, people build huge tables of hashes wherein the plaintext is already known, so that attempting to crack hashes becomes a simple problem of looking up the hash in the table and its corresponding value, instead of attempting to reverse the hash. Because of this, it is very easy to crack simple hashes by simply doing a lookup.
```
damithc: `since the hashing function is easy to compute one-way` this wasn't mentioned before

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,113 @@
...
+
+An example of a service that provides this is [Crackstation](https://crackstation.net/).
+
+Since attacks like rainbow tables exist, passwords need another layer of security.
```
damithc: These two sentences say the same thing:
> Because [rainbow tables](https://en.wikipedia.org/wiki/Rainbow_table) exist, unfortunately.
> Since attacks like rainbow tables exist, passwords need another layer of security.

Get rid of the first?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,113 @@
...
+
+One question that is commonly asked by developers is where to store the salt. The salt can be stored in plaintext, along with the user in the database. Since the goal of the salt is only to prevent precomputed rainbow tables from being used, it doesn't need to be encrypted in the database.
+
+Many good password hashing algorithms today have built-in salts, such as [Argon2](https://github.com/P-H-C/phc-winner-argon2), [SCrypt](https://passlib.readthedocs.io/en/stable/lib/passlib.hash.scrypt.html) and [bcrypt](https://security.stackexchange.com/questions/4781/do-any-security-experts-recommend-bcrypt-for-password-storage). A good password hashing algorithm or library will salt automatically.
+
+### If an attacker has gained access to the entire application, what does it matter if passwords are stored in plaintext or not?
```
damithc: Shorten? e.g., `What if there is a server breach?`

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,113 @@
...
+
+Additionally, when a user signed up on your website and provided you a password, they implicitly trusted you to keep that information safe and secure for them. In a sense, you do have a responsibility to keep their passwords secure. By doing proper password storage, if your servers ever get breached, you can assure your customers that their passwords are properly secured, and still maintain some of their trust in you.
+
+## Getting started
+
+<p style="color: #721c24; background-color: #f8d7da; border-color: #f5c6cb; padding: .75rem 1.25rem; border: 1px solid transparent; border-radius: .25rem;">
+Reading this article doesn't make you an expert in writing your own password hashing code. It is far too easy to screw up and make a mistake. Instead, use one of the free libraries that provide a crypto function that has already been well-tested by the community. <b>Do not write your own crypto library.</b>
+</p>
```
damithc: Use a `<box type="danger">` instead?

~~Reading this article doesn't make you an expert in writing your own password hashing code.~~ Sounds superfluous.

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,113 @@
...
+
+Here are some libraries you can use to implement password storage, depending on the language:
+* If you're using PHP, the [password_hash](https://secure.php.net/manual/en/function.password-hash.php) function will do the job for you. 
+* If you're using Java, use [SecretKeyFactory](https://www.owasp.org/index.php/Hashing_Java) to apply hashes with SHA256.
+* If you're using Python, use [PassLib](https://passlib.readthedocs.io/en/stable/narr/hash-tutorial.html).
+* If you're using JavaScript, use [bcrypt](https://www.npmjs.com/package/bcrypt).
```
damithc: Suggestion:

>Here are some libraries you can use to implement password storage:
>* PHP:  [password_hash](https://secure.php.net/manual/en/function.password-hash.php) 
>* Java: [SecretKeyFactory](https://www.owasp.org/index.php/Hashing_Java)
>* Python: [PassLib](https://passlib.readthedocs.io/en/stable/narr/hash-tutorial.html)
>* JavaScript: [bcrypt](https://www.npmjs.com/package/bcrypt)

damithc (changes requested):
Getting close to LGTM. Need to make the content a bit more beginner-friendly. This is meant for readers with zero prior knowledge on the topic.

contents/security/passwordStorage.md
ChooJeremy: Sure, but this makes it so it doesn't fit the header of that section?
`### Why isn't hashing enough?`

damithc: The whole section answers the question posed by the heading. The first sentence of the section need not be the direct answer. It's more appropriate for the last sentence to be the answer. 
Example:

>**Why isn't hashing enough?** Since the hashing function is easy to compute one-way, one possible strategy to break hashes is to build huge tables of plaintext and their corresponding hashes (these tables are called _rainbow tables_), so that cracking hashes becomes a simple problem of looking up the hash in the table. Because attacks like rainbow tables exist, hashing is not enough to store passwords securely.

ChooJeremy:
Updated based on review. Not sure if I should _italicize_ all new terms or just those that aren't explained with  tooltip/popover. I moved the `small change -> completely different output` to the encryption section so that it doesn't imply that it applies only to hashing. Elaborated on how hashing can't be reversed, but it feels far harder to understand with that explanation. Ready for another review.

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,150 @@
...
+- Hashing
+- Salting
+
+<popover id="pop:plaintext" title="_Plaintext_ refers to unencrypted information" placement="top">
+  <div slot="content">
+It is a cryptography term generally referring to text before encyption or after decrypting it. Another term for it is cleartext.
+  </div>
+</popover>
+
+## What is encryption?
```
damithc: Make these headings `Encryption`, `Hashing` and `Salting`? `What is ` feels superfluous.

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,150 @@
...
+
+Encryption refers to the process of changing a message such that it becomes essentially random text. The opposite process is called _decryption_, which converts the text back into the original message. In order to do this, it takes in an <trigger for="pop:encrypt-key">encryption key</trigger>, which is then re-used in the decryption process.
```
damithc: >Encryption refers to the process of changing a message such that it becomes essentially random text.

I have concerns about this definition. What `message`? Is converting plain text to random text is encryption if there is no way to recover the original text?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,150 @@
...
+plaintext = AESCipher2.decrypt(ciphertext)
+
+print(plaintext))
+```
+
+<popover id="pop:aes" title="_AES_ stands for Advanced Encryption Standard" placement="top">
+  <div slot="content">
+It is one of many possible encryption algorithms.
+  </div>
+</popover>
```
damithc, line 64: When I asked for an example, i didn't mean python code. Rather, explain the concept using a concrete example. You can find some inspiration here https://www.howtogeek.com/howto/33949/htg-explains-what-is-encryption-and-how-does-it-work/

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,150 @@
...
+  <div slot="content">
+It is a cryptography term generally referring to text before encyption or after decrypting it. Another term for it is cleartext.
+  </div>
+</popover>
+
+## What is encryption?
+
+Encryption refers to the process of changing a message such that it becomes essentially random text. The opposite process is called _decryption_, which converts the text back into the original message. In order to do this, it takes in an <trigger for="pop:encrypt-key">encryption key</trigger>, which is then re-used in the decryption process. 
+
+Additionally, even a small change in the message will result in a completely different <trigger for="pop:ciphertext">ciphertext</trigger>. This means that even very similar passwords that vary only by capitalization or a letter would still result in different ciphertext, so attackers can't compare the hash to try and figure out which users have similar passwords.
```
damithc: Are you sure about this? This sounds like a desirable quality rather than a required quality.

damithc: Reader: 🤔 What is a `hash`?

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,150 @@
...
+print(plaintext))
+```
+
+<popover id="pop:aes" title="_AES_ stands for Advanced Encryption Standard" placement="top">
+  <div slot="content">
+It is one of many possible encryption algorithms.
+  </div>
+</popover>
+
+Encryption might seem like a good idea because it transfoms the data into essentially random text, which prevents all of the problems with storing the data directly in plaintext. However, because encryption is <tooltip content="If a function is reversible, and it converts from x to y, then it can also convert from y back to x"> reversible</tooltip>, it is always possible to regain the original password from the <trigger for="pop:ciphertext">ciphertext</trigger>. Since the data is encrypted, the <trigger for="pop:encrypt-key">encryption key</trigger> must also be stored somewhere. This means that if someone manages to hack into the application and read the encrypted passwords, it is also likely that they will be able to read the encryption key. With the encryption key, they will be able to decrypt all the passwords and read them anyway. This makes encryption unsuitable for password storage.
```
damithc: Previously, we talked about `messages`. Now we are taking about `data`.

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,150 @@
...
+
+<popover id="pop:encrypt-key" title="All encryption algorithms require an _encryption key_" placement="top">
+  <div slot="content">
+The key is usually randomly generated text that is used to encrypt the original data. To then decrypt the data, the same key (and algorithm) must be used.
+  </div>
+</popover>
```
damithc, line 84: It is enough to use the tooltip at the first instance of using a term.

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,150 @@
...
+
+## What is hashing?
+
+Hashing is a <trigger for="pop:oneway">one-way</trigger> function that transforms a set of data into another set of data. Unlike encryption, hashing works by re-iterating on the input multiple times, each time using values that are dependent on the previous state, then destroying the previous state. This makes it impossible to reverse as a computer would have to test all possible states on each iteration, which compounds into several billion possible states.
```
damithc: I'm doubtful if this is definition of hashing in general. If feels like you are taking a specific use of hashing and giving it as the general definition.

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,150 @@
...
+</popover>
+
+## What is hashing?
+
+Hashing is a <trigger for="pop:oneway">one-way</trigger> function that transforms a set of data into another set of data. Unlike encryption, hashing works by re-iterating on the input multiple times, each time using values that are dependent on the previous state, then destroying the previous state. This makes it impossible to reverse as a computer would have to test all possible states on each iteration, which compounds into several billion possible states. 
+<popover id="pop:oneway" title="" placement="top">
+  <div slot="content">
+    A one way function is a function that is easy to compute the result of, but whose results are difficult to reverse back to the original input.
+  </div>
+</popover>
```
damithc: This tooltip also appear as normal text

contents/security/passwordStorage.md
```diff
@@ -0,0 +1,150 @@
...
+
+For example, the following python code hashes the message "HashingIsFun!" with SHA1:
+```python
+import hashlib
+
+sha1Hash = hashlib.sha1()
+sha1Hash.update("HashingIsFun!".encode('utf-8'))
+hashOutput = sha1Hash.hexdigest()
+print(hashOutput)
+```
```
damithc: Same comment as before: an example to illustrate the concept, not code

damithc (changes requested):
Added some comments. There are some small gaps in the explanation that still makes it not fully accessible to the target reader. Try to forget everything you know about the topic and read through the eyes of the target reader.

damithc:
@ChooJeremy Any progress on this?

ChooJeremy:
Some. It is hard to address the reader's questions while still keeping it simple and non-specific, without going into too much detail into a single algorithm.

damithc:
Merged and deployed.
Good work @ChooJeremy 
Thanks to reviewers too

PR #87, opened 2019-03-02
Intro to SQL

### LINKED ISSUE #68: Introduction to SQL
Outline:
- What is SQL
- Why use SQL
- An example showing basic commands
- Getting started
- Other resources for learning and integrating SQL

amrut-prabhu:
Fixes #68

Preview [here](https://deploy-preview-87--learningresources.netlify.com/contents/data/sql/sql)


amrut-prabhu:
@RonakLakhotia @ChooJeremy Can you review this?

damithc:
One quick comment: A large part of the article is explaining SQL which is independent from MySQL.

ChooJeremy:
I'm not super familiar with the differences between MySQL, mssql, PostgreSQL etc. so I might not be the best person to review this.

However, I'm wondering who is the intended audience of the article? Is it someone who's building their own application and who's finding it difficult to just store objects in text? Someone who's interested in learning about databases? I agree with prof that your article appears to be more on teaching SQL than MySQL, and I feel like it's hard to see who is article is useful for.

amrut-prabhu:
@damithc Would it be better for me to specify that it is SQL (and hence, MySQL)? And I can mention that are slight syntax differences across SQL instances, but the syntax used in the article is MySQL syntax compliant

Or should I talk more about the MySQL Workbench software capabilities?

damithc:
> @damithc Would it be better for me to specify that it is SQL (and hence, MySQL)? And I can mention that are slight syntax differences across SQL instances, but the syntax used in the article is syntax compliant
> 
> Or should I talk more about the MySQL Workbench software capabilities?

It's fine if the connection between MySQL and SQL is made clear. Alternatively, you can extract a separate article on SQL.

amrut-prabhu:
@ChooJeremy The intended audience is someone who is new to databases. After reading the DBMS article, they can refer to this MySQL one to see a concrete example of how to use database systems.

Also, the example in MySQL/SQL serves to introduce them to the concept of clauses. It doesn't seem useful to talk about MySQL without SQL. So perhaps, the SQL portion can be extracted out to a separate chapter in the future, when there are chapters for PostgreSQL, MSSQL etc.?

crphang:
I think this article is largely about SQL. The part about MySQL is too general to justify the title currently. It is largely the same as most other implementations of RDBMS. I think it would really shine if we can contrast the internal workings of MySQL as compared to PostgreSQL and etc and perhaps when each will be more optimal.

However, I think at this stage, it might be helpful to focus more on SQL. Which really goes beyond just being a query language for RDBMS these days (research more on this).

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,243 @@
...
+---
+
+## Why learn SQL?
+
+Being open source and one of the first DBMSs, MySQL increased the popularity of SQL in its early days. It ended up being a core component of the <tooltip content="Linux OS, Apache Server, MySQL RDBMS, PHP  language">LAMP</tooltip> web application stack (and many others).
+Here are some of the main reasons behind the widespread adoption of SQL and why it still remains popular today.
+
+### Easy to learn
+
+SQL is very easy to learn, even for beginners who do not have any prior experience with databases. Since it has been around for a few decades, there are many good books and online resources to learn from.
```
RonakLakhotia: Maybe don't use terms like `very easy` to trivialize it?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,243 @@
...
+## Disadvantages
+
+SQL is not without its problems. In general, the biggest problem is with regards to the features of SQL.
+
+Although SQL databases use established <tooltip content="American National Standard Institutes">ANSI</tooltip> & <tooltip content="International Organization for Standardization">ISO</tooltip> standards, some distributions add proprietary extensions to the standard SQL to ensure customer lock-in.
+Thus, the available feature set varies according to the SQL distribution that you are using. This can make it confusing and frustrating to use SQL when switching across distributions.
+
+Apart from that, most **SQL** problems are not uniformly applicable to all of its distributions.
+
+For example, **MySQL** suffers from concurrency issues. Though it performs well with read operations, it *can* be problematic when there are many concurrent read-write operations. A symptom of this issue would be a sudden slowdown of a well-optimized query. However, **PostgreSQL** deals with concurrency well by being fully <tooltip content="Atomicity, Consistency, Isolation, Durability properties are satisfied">ACID</tooltip> compliant and implementing transactions isolation.
```
RonakLakhotia: Can elaborate a bit more on how PostgreSQL is helpful here?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,243 @@
...
+One of the benefits (and reasons) for SQL's popularity is that there are free distributions available (like MySQL, PostgreSQL and SQLite) as well as paid ones (like Microsoft SQL Server and Oracle) that come with more functionality.
+
+One such distribution is MySQL. Since it is an open source software, you do not have to pay to use MySQL. It even comes with official ([MySQL Workbench](https://dev.mysql.com/doc/workbench/en/)) as well third party easy-to-use <tooltip content="Graphical User Interface">GUIs</tooltip>, which are less daunting to new users as compared to a <tooltip content="Command Line Interface">CLI</tooltip>.
+
+### Compatibility
+
+SQL distributions work on many operating systems and more importantly, can be integrated with many languages. This includes languages like PHP, PERL, C++, Ruby, Java as well as *relatively "newer"* ones like Python, JavaScript (Node.js) and Go.
+
+In addition, online playgrounds like [DB Fiddle](https://www.db-fiddle.com/) and [JDoodle](https://www.jdoodle.com/execute-sql-online) make it easy to use and learn the SQL language quickly without the hassle of setting up an environment or application.
+
```
RonakLakhotia: You should credit your sources when referring to external articles

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,243 @@
...
+Here are some recommended resources that you can use to start learning SQL:
+
+- Khan Academy offers a very good [SQL course](https://www.khanacademy.org/computing/computer-programming/sql) for free. It has a mix of videos, text and exercises that you can choose to do at your own pace. It is well organized and even provides summary notes.
+
+- W3Schools offers an interactive written tutorial for [SQL](https://www.w3schools.com/sql/default.asp).
+
+- You can also use a [SQL Quick Reference Guide](https://www.w3schools.com/sql/sql_ref_mysql.asp) to look up information quickly once you're familiar with the language.
+
+- Understand what a database and [<tooltip content="DataBase Management System">DBMS</tooltip>]({{baseUrl}}/contents/data/databases/databases.html) are. More specifically, understand the basic concepts of [RDBMS](https://www.tutorialspoint.com/sql/sql-rdbms-concepts.htm). This is not essential, but will give you a better high-level understanding before diving in to programming.
```
RonakLakhotia: Add this point in the beginning? As this seems to be a preliminary step

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,243 @@
...
+# Introduction to SQL
+
+Author(s): [Amrut Prabhu](https://github.com/amrut-prabhu)
+
+## What is SQL?
+
+Databases are integral to any commercial software application, and most student side projects as well, whether it is a web app, desktop app or otherwise.
+Though you can choose to just use a text file to store your data, this will not be sustainable as the size of your application (and the data stored) grows. This is why there are several dedicated database software packages available today.
+
+**SQL** is a special-purpose programming language designed for managing (storing, querying and manipulating) data in a relational database. Several <tooltip content="Relational DataBase Management System">RDBMS</tooltip> distributions like MySQL, Oracle, SQL Server and PostgreSQL use the SQL language. However, the syntaxes used in these distributions vary slightly. These differences may be in terms of case-sensitivity, available built-in functions, custom functions, date and time formats, and so on.
```
RonakLakhotia: A tooltip to give a quick understanding of `relational database`?

RonakLakhotia (commented):
The structure looks good in general, you could just a do a quick reading yourself to check for missing full stops and commas.

contents/data/sql/SQL.md
RonakLakhotia: What's the implication with `special purpose`?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,243 @@
...
+
+<box type="tip">
+  Many online resources refer to Microsoft SQL Server as SQL. So, keep that in mind before you start making conclusions about the characteristics and features of SQL, the language.
+</box>
+
+---
+
+## Why learn SQL?
+
+Being open source and one of the first DBMSs, MySQL increased the popularity of SQL in its early days. It ended up being a core component of the <tooltip content="Linux OS, Apache Server, MySQL RDBMS, PHP  language">LAMP</tooltip> web application stack (and many others).
```
RonakLakhotia: Make the relation between MySQL and SQL more clear?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,243 @@
...
+### Easy to learn
+
+SQL is very easy to learn, even for beginners who do not have any prior experience with databases. Since it has been around for a few decades, there are many good books and online resources to learn from.
+In addition, SQL and its distributions have a huge support community (such as [Stack Overflow](https://stackoverflow.com/questions/tagged/sql) and the [official MySQL forum](https://forums.mysql.com/)) which can prove useful when you run into problems while using SQL.
+
+### Free of cost
+
+One of the benefits (and reasons) for SQL's popularity is that there are free distributions available (like MySQL, PostgreSQL and SQLite) as well as paid ones (like Microsoft SQL Server and Oracle) that come with more functionality.
+
+One such distribution is MySQL. Since it is an open source software, you do not have to pay to use MySQL. It even comes with official ([MySQL Workbench](https://dev.mysql.com/doc/workbench/en/)) as well third party easy-to-use <tooltip content="Graphical User Interface">GUIs</tooltip>, which are less daunting to new users as compared to a <tooltip content="Command Line Interface">CLI</tooltip>.
```
RonakLakhotia: Repeated point of MySql being open sourced?

contents/data/sql/SQL.md
RonakLakhotia: Apart from concurrency, are there any other disadvantages of using Sql as a developer

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,243 @@
...
+Table 4. Final result after `ORDER BY`
+
+
+This is the same as our expected output from Table 2!
+
+<box type="info">
+  You can experiment with this example on <a href="https://www.db-fiddle.com/f/yxjjgbkKmsa46cKjeEg1X/2">DB Fiddle</a> by entering queries into the <code>Query SQL</code> pane and then clicking the <code>Run</code> icon.
+</box>
+
+The example shown here is relatively simple. Typical SQL queries have the capability to be much more complex as there are a lot of clauses, functions and operators that are not covered here.
```
RonakLakhotia: How does it affect developers using sql as the queries become more complex?

RonakLakhotia (changes requested):

contents/data/sql/SQL.md
amrut-prabhu: I think this is enough. AC**I**D gives the reader an idea of where to look, in case they want more details.

amrut-prabhu: I feel that this is covered in the preceding paragraph

amrut-prabhu:
@rrtheonlyone Since you have experience with SQL, would you mind reviewing this as well?

damithc:
Some quick comments:
You never mentioned what SQL stands for. 😀 
Perhaps the HOW part should be integrated into the WHAT part (i.e., intro). That way, the reader has a better idea about WHAT SQL is before considering the WHY. Otherwise the WHY points could sound too abstract. For example, you should be able to say in the WHY section `As you saw in the example above, SQL is not complex and therefore it is easy to learn`.

Phrase sibling headings the same way. e.g.,
* Easy to Learn
* Free { `of cost` is not necessary?}
* Highly Compatible {`Compatibility` doesn't match the other two headings}

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,242 @@
...
+
+<div class="website-content">
+
+# Introduction to SQL
+
+Author(s): [Amrut Prabhu](https://github.com/amrut-prabhu)
+
+## What is SQL?
+
+Databases are integral to any commercial software application, and most student side projects as well, whether it is a web app, desktop app or otherwise.
```
rrtheonlyone: You have put the header as what is SQL. Keep in mind a new user may not even know what is SQL when reading this.

The next thing they read is something about databases? Are we supposed to make the logical connection that SQL has something to do with databases implicitly? 

I feel its cleaner if you start with defining SQL and touching on it a bit before going into the importance of databases!

rrtheonlyone: Why say 'most student side projects'?

 As a reader, this immediately makes me think the article isn't describing something that important.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,242 @@
...
+Though you can choose to just use a text file to store your data, this will not be sustainable as the size of your application (and the data stored) grows. This is why there are several dedicated database software packages available today.
```
rrtheonlyone: Minor nit:
Why not say something like => As the size of your application grows, we need to look into dedicated database software packages to store your data. It will not be enough to simply rely on in-memory solutions or a simple text file to store data..

It is easier to read if you start a paragraph with your point and elaborate further as we read. (the impact is clearer). This is your call though!

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,242 @@
...
+
+**Structured Query Language** (**SQL**) is a programming language that is specifically designed for managing (storing, querying and manipulating) data in a <tooltip content="relational databases store data in tables">[relational database]({{baseUrl}}/contents/data/databases/databases.htmll#database-models)</tooltip>. As mentioned in [Codecademy](https://www.codecademy.com/articles/what-is-rdbms-sql), most <tooltip content="Relational DataBase Management System">RDBMSs</tooltip> like MySQL, Oracle, SQL Server and PostgreSQL use the SQL language. However, the syntaxes used in these distributions vary slightly. These differences may be in terms of case-sensitivity, available built-in functions, custom functions, date and time formats, and so on.
```
rrtheonlyone: As mentioned in Codecademy? Are you saying this to add credibility to your definition?

I dont see how this adds any value.

rrtheonlyone: "These differences may be in terms of case-sensitivity, available built-in functions, custom functions, date and time formats, and so on. " 

Maybe rephrase this sentence? Sounds odd when I say it out.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,242 @@
...
+
+**Structured Query Language** (**SQL**) is a programming language that is specifically designed for managing (storing, querying and manipulating) data in a <tooltip content="relational databases store data in tables">[relational database]({{baseUrl}}/contents/data/databases/databases.htmll#database-models)</tooltip>. As mentioned in [Codecademy](https://www.codecademy.com/articles/what-is-rdbms-sql), most <tooltip content="Relational DataBase Management System">RDBMSs</tooltip> like MySQL, Oracle, SQL Server and PostgreSQL use the SQL language. However, the syntaxes used in these distributions vary slightly. These differences may be in terms of case-sensitivity, available built-in functions, custom functions, date and time formats, and so on.  
+
+SQL distributions are widely used in many small and big companies. According to the official [MySQL website](https://www.mysql.com/why-mysql/), it is used in companies like Facebook, Google, Adobe, Paytm and Zappos (though they may not be using MySQL exclusively).
```
rrtheonlyone: I am confused between MySQL and SQL when I read this.

Are the companies using MySQL specifically? It felt like you were talking about general SQL distributions in your first line. 

"many small and big companies" => why not just say many companies (or use another adjective for greater impact)

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,242 @@
...
+ORDER BY column_name(s);
+```
+
+<box type="tip">
+  The uppercase words in this query are clauses or statements, and are SQL keywords.
+</box>
+
+#### Breaking down the clauses
+
+Let's understand what the above query means by using an example to make it more concrete.
```
rrtheonlyone: The above 'query' does not mean anything actually. I think you want to explain how to use the syntax further here right? (the structure)

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,242 @@
...
+
+Suppose have the data table `Students` shown here. We want to get a list of CS courses that have more than 1 student. The output should display the Course name and number of students, and should be sorted by increasing number of students.
```
rrtheonlyone: Suppose we have the data table...

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,242 @@
...
+| CS202  | 3   |
+
+Table 2. Expected output
+
+The corresponding query will be:
+
+```
+SELECT Course, COUNT(*) num
+FROM Students
+WHERE Course LIKE 'CS%'
```
rrtheonlyone: I personally dont like that you had to use pattern matching in your introductory example! Why not create another column with the course faculty (so its easier to write a query)

However, this is okay as well (I fear its hard to understand if your seeing it as a beginner)

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,242 @@
...
+
+- Khan Academy offers a very good [SQL course](https://www.khanacademy.org/computing/computer-programming/sql) for free. It has a mix of videos, text, and exercises that you can choose to do at your own pace. It is well organized and even provides summary notes.
+
+- W3Schools offers an interactive written tutorial for [SQL](https://www.w3schools.com/sql/default.asp).
+
+- You can also use a [SQL Quick Reference Guide](https://www.w3schools.com/sql/sql_ref_mysql.asp) to look up information quickly once you're familiar with the language.
+
+---
+
+## Where to go from here?
```
rrtheonlyone: Do you think choosing from the sql distribution first is the way to go?

I personally prefer to do it the other way round. Start with a web application and use a lightweight sql instance to store your data (e.g. sqlite?) Furthermore, many frameworks already provide database support by default (e.g. Django). 

This is okay as well!

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,242 @@
...
+  You can experiment with this example on <a href="https://www.db-fiddle.com/f/yxjjgbkKmsa46cKjeEg1X/2">DB Fiddle</a> by entering queries into the <code>Query SQL</code> pane and then clicking the <code>Run</code> icon.
+</box>
+
+The example shown here is relatively simple. Typical SQL queries have the capability to be much more complex as there are a lot of clauses, functions, and operators that are not covered here. As the complexity of the query grows, there is a possible decrease in the performance of the query. This is why [query planning](https://www.khanacademy.org/computing/computer-programming/sql/relational-queries-in-sql/a/more-efficient-sql-with-query-planning-and-optimization) and [optimization](https://www.sisense.com/blog/8-ways-fine-tune-sql-queries-production-databases/) is important.
+
+---
+
+## Why learn SQL?
+
+Being open source and one of the first DBMSs, MySQL (a specific implementation of SQL) increased the popularity of **SQL** in its early days. It ended up being a core component of the <tooltip content="Linux OS, Apache Server, MySQL RDBMS, PHP  language">LAMP</tooltip> web application stack (and many others).
```
rrtheonlyone: What does this sentence have to do with 'Why learn SQL?'

rrtheonlyone (commented):
Really nice article! I think this is a good introduction to SQL and how to get started! The step by step explanation is good (I liked that you provided a fiddle as well)

Some suggestions:
- In the tutorial, why not create a slightly bigger student data table (with more fields)? Readers will then be able to appreciate the importance of SQL and how to query it

- One important idea behind SQL is the DB schema. I don't think you have mentioned that here. Why not take a top-down approach (show a picture of various fields and how SQL neatly organises them into different tables).

- Some of the content here feels largely copied pasted/inferred from various sources? When I read it from start to end, there is no clear flow. Each section has no connection with the previous and it sometimes starts rather abruptly. Perhaps you could just touch up the start/end of sections a little bit?

amrut-prabhu:
@rrtheonlyone Thanks for your review!

> why not create a slightly bigger student data table (with more fields)

I wanted the example to showcase how SQL queries can be used, rather than showing that complex data needs a DBMS. I feel a simplistic approach would be more useful to our target reader, rather than overwhelming them with complexity.

> One important idea behind SQL is the DB schema. I don't think you have mentioned that here. Why not take a top-down approach (show a picture of various fields and how SQL neatly organises them into different tables).

Once again, this feels like it has more to do with relational databases itself, rather than SQL. Maybe it can be added as an extension to the DBMS chapter, or a new RDBMS chapter can be created instead?

The intention is to introduce the reader to the benefits of SQL and related resources, which is why I didn't mention anything about DDL, DCL, DML, TCL (since it also varies across SQL distributions). I felt that this would also make the chapter much longer than it needs to be. Do you think it would be good to introduce these as well (or not, since they aren't SQL specific)?

contents/data/sql/SQL.md
amrut-prabhu: > Are we supposed to make the logical connection that SQL has something to do with databases implicitly?

Umm to be fair, it _is_ under the `Databases` section of the book 😅

contents/data/sql/SQL.md
amrut-prabhu: > Do you think choosing from the sql distribution first is the way to go?

I feel that in order to **learn** a  completely new type of language (not just use it), it's best to start from the basics. And, that means straight up writing queries to a database. 

> and use a lightweight sql instance to store your data (e.g. sqlite?)

Doesn't this involve implicitly choosing a distribution? 😛 

> Start with a web application and use a lightweight sql instance to store your data

Yes, but that requires you to develop a web application as well, right? And it will require you to have knowledge of the language as well (Python, in the case of Django). It won't be independent SQL, which is why I suggested starting out on MySQL/SQLite online instead.

contents/data/sql/SQL.md
amrut-prabhu: It provides some context for `increased the popularity of SQL in its early days`.

I guess I can remove it too

contents/data/sql/SQL.md
rrtheonlyone: Haha yes fair point :)

contents/data/sql/SQL.md
rrtheonlyone: Yes fair enough! If you intend to learn pure SQL, this is the path.

I never actually started learning SQL on its own (and most people dont right?) -> its just something you pick up along the way when building bigger applications :)

rrtheonlyone (approved):
I am good with this 👍 

This article serves as a good introduction to SQL (without mixing in concepts of RDMS). Readers can use this as a stepping stone to pick up some basics before moving on to actually use SQL in their projects :)

RonakLakhotia (approved):
@amrut-prabhu good effort in writing this book chapter. Having no prior experience in SQL, I can get a good overview of the concepts after reading the chapter.

amrut-prabhu:
@rrtheonlyone I made some minor changes since you last reviewed. Would you mind taking a quick look at  it again?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,241 @@
...
+# Introduction to SQL
+
+**Author(s): [Amrut Prabhu](https://github.com/amrut-prabhu)**
+
+Reviewers: [Ronak Lakhotia](https://github.com/RonakLakhotia), [Rahul Rajesh](https://github.com/rrtheonlyone)
+
+## What is SQL?
+
+**Structured Query Language** (**SQL**) is a programming language that is specifically designed for managing (storing, retrieving and manipulating) data in a <tooltip content="relational databases store data in tables">[relational database]({{baseUrl}}/contents/data/databases/databases.html#database-models)</tooltip>. Most <tooltip content="Relational DataBase Management System">RDBMS</tooltip> distributions like MySQL, Oracle, SQL Server and PostgreSQL use the SQL language.
+However, the syntaxes used in these distributions vary slightly. These differences may be in terms of whether the syntax is case-sensitive, what  functions are available out-of-the-box, the format for specifying date and time, and so on.
```
damithc: You can consider putting the last sentence as a tooltip of `vary slightly` in the previous sentence.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,241 @@
...
+
+## What is SQL?
+
+**Structured Query Language** (**SQL**) is a programming language that is specifically designed for managing (storing, retrieving and manipulating) data in a <tooltip content="relational databases store data in tables">[relational database]({{baseUrl}}/contents/data/databases/databases.html#database-models)</tooltip>. Most <tooltip content="Relational DataBase Management System">RDBMS</tooltip> distributions like MySQL, Oracle, SQL Server and PostgreSQL use the SQL language.
+However, the syntaxes used in these distributions vary slightly. These differences may be in terms of whether the syntax is case-sensitive, what  functions are available out-of-the-box, the format for specifying date and time, and so on.  
+
+Databases are integral to any commercial software application, whether it is a web app, desktop app or otherwise.
+As the size of your application grows, you need to look into dedicated database software packages (like SQL distributions) to store your data. It will not be enough to simply rely on in-memory solutions or a simple text file to store the data.
```
damithc: Earlier you used _RDBMS Distributions_ and here you use _SQL distributions_. Do they refer to the same thing?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,241 @@
...
+
+SQL distributions (such as MySQL) are widely used in many successful companies. For example, according to the official [MySQL website](https://www.mysql.com/why-mysql/), it is used in companies like Facebook, Google, Adobe, Paytm and Zappos (though they may not be using MySQL exclusively).
+
```
damithc: Would this order of explanation be more natural to the reader?
* Many software needs to store data.
* Many of them use RDBMS for that purpose.
* SQL is a widely use language to manage data in an RDBMS, although with minor variations.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,241 @@
...
+<box type="tip">
+  Many online resources refer to Microsoft SQL Server as SQL. So, keep that in mind before you start making conclusions about the characteristics and features of SQL, the language.
+</box>
+
+
+### How does SQL work?
+
+Now that you have a general idea of what SQL is, let's take a look at how it can be used. SQL uses **queries** to retrieve data that was previously stored in the database. This is what a basic SQL query (or "command") for retrieving information from a data table looks like:
```
damithc: I appreciate the role of the first sentence as a connector. Although it feels unnecessary. See if the section can survive without it.

damithc: A shorter version:

~~Now that you have a general idea of what SQL is, let's take a look at how it can be used.~~ SQL uses **queries** to retrieve data ~~that was previously stored~~ from a database. ~~This is what a basic SQL query (or "command") for retrieving information from a data table looks like~~ Here is an example SQL query:

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,241 @@
...
+
+```
+SELECT column_name(s)
+FROM table_name
+WHERE condition
+GROUP BY column_name(s)
+HAVING condition
+ORDER BY column_name(s);
+```
```
damithc: Make this an actual query? There is no need to give the generic form as you are not trying to teach SQL.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,241 @@
...
+
+<box type="tip">
+  The uppercase words in this query are clauses or statements, and are SQL keywords.
+</box>
+
+#### Breaking down the clauses
```
damithc: Probably no need for this heading

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,241 @@
...
+
+
+This is the same as our expected output from Table 2!
+
+<box type="info">
+  You can experiment with this example on <a href="https://www.db-fiddle.com/f/kHqV2edUGxCc1dU6vE6CmS/0">DB Fiddle</a> by entering SQL queries and then running them.
+</box>
+
+The example shown here is pretty typical. However, SQL queries for retrieving information have the capability to be much more complex as there can be a lot of other clauses, functions, and operators that are not covered here. As the complexity of the query grows, there is a possible decrease in the performance of the query. This is why [query planning](https://www.khanacademy.org/computing/computer-programming/sql/relational-queries-in-sql/a/more-efficient-sql-with-query-planning-and-optimization) and [query optimization](https://www.sisense.com/blog/8-ways-fine-tune-sql-queries-production-databases/) is important.
+
```
damithc, line 90: It feels like the explanation is a bit too long, as if we are trying to teach SQL. Perhaps, this would do?
* Give `Students` table
* Give an example query with a brief explanation.
* Give the output for the above query

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,241 @@
...
+---
```
damithc: Also briefly mention other types of SQL statements, such as insert, delete, etc. to give a taste of things possible?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,241 @@
...
+
+### Free
+
+One of the benefits (and reasons) for SQL's popularity is that there are free distributions available (like MySQL, PostgreSQL and SQLite) as well as paid ones (like Microsoft SQL Server and Oracle) that come with more functionality.
+
+One such distribution is MySQL. It is free of cost and comes with [official](https://dev.mysql.com/doc/workbench/en/) as well third party easy-to-use <tooltip content="Graphical User Interface">GUIs</tooltip>, which are less daunting to new users as compared to a <tooltip content="Command Line Interface">CLI</tooltip>.
+
+### Highly Compatible
+
+SQL distributions work on many operating systems and more importantly, can be integrated with many languages. This includes languages like PHP, PERL, C++, Ruby, Java as well as *relatively "newer"* ones like Python, JavaScript (Node.js) and Go.
```
damithc: Python is older than Java :-)

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,241 @@
...
+
+In addition, online playgrounds like [DB Fiddle](https://www.db-fiddle.com/) and [JDoodle](https://www.jdoodle.com/execute-sql-online) make it easy to use and learn the SQL language quickly without the hassle of setting up an environment or application.
+
+## Disadvantages
+
+SQL is not without its problems. In general, the biggest problem is with regards to the features of SQL.
+
+Although SQL databases use established <tooltip content="American National Standard Institutes">ANSI</tooltip> & <tooltip content="International Organization for Standardization">ISO</tooltip> standards, some distributions (PostgreSQL, for example) add proprietary extensions to the standard SQL to ensure customer lock-in.
```
damithc: Well, that's a point of view. The vendors might say the extensions are to give users extra value, not to ensure lock-in :-)

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,241 @@
...
+In this way, the available feature set can vary according to the SQL distribution that you are using. This can make it confusing and frustrating to use SQL when switching across distributions.
+
+Apart from that, most **SQL** problems are not uniformly applicable across all of its distributions. For example, MySQL suffers from concurrency issues. Though it performs well with read operations, it *can* be problematic when there are many concurrent read-write operations. A symptom of this issue would be a sudden slowdown of a query. However, PostgreSQL deals with concurrency well by being fully <tooltip content="Atomicity, Consistency, Isolation, Durability properties are satisfied">ACID compliant</tooltip> and implementing <tooltip content="each query transaction is isolated from other  transactions running simultaneously in the system">transactions isolation</tooltip>.
```
damithc: Not fair to throw in ACID as a mere tooltip. It's best to avoid adding more technical terms if possible.

contents/data/sql/SQL.md
damithc: Are you sure _SQL Distributions_ is the correct term? It didn't feel right because SQL is a language but unlike other languages, it doesn't come as a standalone installation. Rather; it comes with a database software. Besides, the phrase has a very low count in Google search results.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,241 @@
...
+  - W3Schools have interactive written tutorials such as [MySQL Node.js](https://www.w3schools.com/nodejs/nodejs_mysql.asp) and [MySQL Python](https://www.w3schools.com/python/python_mysql_getstarted.asp), for learning how to integrate MySQL with Node.js and Python respectively.
+
+---
+
+## References
+
+- [SQL capabilities](https://www.databasejournal.com/sqletc/article.php/3915331/Top-10-SQL-Hierarchical-Data-Processing-Capabilities.htm)
+- [Advantages and disadvantages of SQL](https://www.quora.com/What-are-the-advantages-and-disadvantages-of-SQL)
+- [MySQL and PostgreSQL differences](http://mwiki.gichd.org/IM/Difference_MySQL_PostGreSQL)
+- [Execution order of SQL query](https://www.designcise.com/web/tutorial/what-is-the-order-of-execution-of-an-sql-query)
```
damithc: Are these already linked somewhere in the main body? In that case no need to list them again. If they are additional resources, you should give your own take of the each.

damithc (changes requested):
Good start @amrut-prabhu Some comments added.

contents/data/sql/SQL.md
amrut-prabhu: Good point, I'll change it to `SQL software` instead

amrut-prabhu: > Earlier you used RDBMS Distributions and here you use SQL distributions. Do they refer to the same thing?

Yes, SQL is specifically for relational databases (RDBMSs). Hence SQL ~distribution~ implementation => RDBMS ?

In that sentence, I wanted to point out that SQL is the dominant language across RDBMSs.

amrut-prabhu:
@damithc Updated based on the review comments

contents/data/sql/SQL.md
damithc: _RDBMS software_ seem more appropriate, as the software is sold as a database, and the SQL support is just a necessary feature?

amrut-prabhu: I considered that too. But the thing is that though most RDBMS use SQL, not all of them do so. Some exceptions are [Rel](https://reldb.org/c/) and [Dataphor](http://www.dataphor.org/About%20Dataphor.ashx). 
In this case, do you think that we can still use RDBMS to be synonymous with SQL? Because it practically is, barring some exceptions

damithc: Perhaps you can use _SQL Software_ (or _SQL Implementations_ or _SQL Distributions_) but explain the exact meaning the first time you use it.

amrut-prabhu: Updated to this:

Unlike other languages, SQL doesn't come as a standalone installation. Rather, _SQL Software_ (or _SQL Implementations_) like MySQL, Oracle, SQL Server and PostgreSQL come with a RDBMS that uses the SQL language.

amrut-prabhu:
@damithc Can you do another round of review for [this](https://deploy-preview-87--learningresources.netlify.com/contents/data/sql/sql)?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,145 @@
...
+
+# Introduction to SQL
+
+**Author(s): [Amrut Prabhu](https://github.com/amrut-prabhu)**
+
+Reviewers: [Ronak Lakhotia](https://github.com/RonakLakhotia), [Rahul Rajesh](https://github.com/rrtheonlyone)
+
+## What is SQL?
+
+Databases are integral to any commercial software application, whether it is a web app, desktop app or otherwise.
```
damithc: This statement is not accurate. 1. Being commercial has no relevant 2. There can be apps that doesn't use a database at all. You don't want to open with such a questionable statement.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,145 @@
...
+As the size of your application grows, you need to look into dedicated database software packages (like SQL software) to store your data. It will not be enough to simply rely on in-memory solutions or a simple text file to store the data. Many companies like Facebook, Google, Adobe, Paytm and Zappos use <tooltip content="Relational DataBase Management System">RDBMS</tooltip> software to fulfil this need.
```
damithc: `As the size of your application grows, you need to look into dedicated database software packages`: this doesn't sound right either. It seems to assume people will start with an in-memory database or a text file and migrate to a database later.  I don't think that happens too often.

damithc: The name dropping is not needed. You don't need to cite companies to prove that RDBMS is widely used.

damithc: You can replace the entire paragraph with a statement such as `Relational Database Management Systems (RDBMS) is one popular type of tools software applications use to manage their data.`

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,145 @@
...
+
+**Structured Query Language** (**SQL**) is a programming language that is specifically designed for managing (storing, retrieving and manipulating) data in a <tooltip content="relational databases store data in tables">[relational database]({{baseUrl}}/contents/data/databases/databases.html#database-models)</tooltip>. Unlike other languages, SQL doesn't come as a standalone installation. Rather, _SQL Software_ (or _SQL Implementations_) like MySQL, Oracle, SQL Server and PostgreSQL come with a RDBMS that uses the SQL language.
+However, the syntaxes and functions used across these software systems <tooltip content="in terms of whether the syntax is case-sensitive, the format for specifying date and time, what functions are available out-of-the-box">vary</tooltip>.
+
```
damithc: First sentence: can be shortened as ... designed for interacting with an RDBMS.
Second sentence: I think the reality is the other way around. An RDBMS comes with an SQL implementation because without it the RDBMS is useless.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,145 @@
...
+Reviewers: [Ronak Lakhotia](https://github.com/RonakLakhotia), [Rahul Rajesh](https://github.com/rrtheonlyone)
+
+## What is SQL?
+
+Databases are integral to any commercial software application, whether it is a web app, desktop app or otherwise.
+As the size of your application grows, you need to look into dedicated database software packages (like SQL software) to store your data. It will not be enough to simply rely on in-memory solutions or a simple text file to store the data. Many companies like Facebook, Google, Adobe, Paytm and Zappos use <tooltip content="Relational DataBase Management System">RDBMS</tooltip> software to fulfil this need.
+
+**Structured Query Language** (**SQL**) is a programming language that is specifically designed for managing (storing, retrieving and manipulating) data in a <tooltip content="relational databases store data in tables">[relational database]({{baseUrl}}/contents/data/databases/databases.html#database-models)</tooltip>. Unlike other languages, SQL doesn't come as a standalone installation. Rather, _SQL Software_ (or _SQL Implementations_) like MySQL, Oracle, SQL Server and PostgreSQL come with a RDBMS that uses the SQL language.
```
damithc: Along the same vein, _MySQL, Oracle, SQL Server and PostgreSQL_ are not SQL implementations that comes with an RDBMS. They are RDBMS that come with an SQL implementation. Anyway, that's my understanding; but I'm not an expert.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,145 @@
...
+However, the syntaxes and functions used across these software systems <tooltip content="in terms of whether the syntax is case-sensitive, the format for specifying date and time, what functions are available out-of-the-box">vary</tooltip>.
+
+<box type="tip">
+  Many online resources refer to Microsoft SQL Server as SQL. So, keep that in mind before you start making conclusions about the characteristics and features of SQL, the language.
+</box>
```
damithc: Are you sure about this? Microsoft SQL Server is sometimes called _Sequel Server_ but I've never seen it referred to as SQL.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,145 @@
...
+SELECT Course, COUNT(*) num
+FROM Students
+WHERE Faculty = 'CS'
+GROUP BY Course
+HAVING COUNT(*) > 1
+ORDER BY num;
+```
+
+<box type="tip">
+  The uppercase words in this query are either **clauses** or **statements**, and are SQL keywords.
```
damithc: Add a blank line before this line. MarkBind needs blank line at the start of markdown syntax. Otherwise it renders like this:
image

damithc: Do you even need this tip? The info given doesn't seem to be used anywhere else in the article.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,145 @@
...
+| ------ | --- |
+| CS202  | 3   |
+| CS101  | 2   |
+
+<box type="info">
+  You can experiment with this example on <a href="https://www.db-fiddle.com/f/kHqV2edUGxCc1dU6vE6CmS/0">DB Fiddle</a> by entering SQL queries and then running them.
+</box>
+
+You can see how this simple query can prove to be extremely useful for getting this information when the table has tens of entries (or many more). Apart from retrieving information, SQL can also be used for creating, deleting and manipulating data with queries like  `INSERT`, `DELETE` and `UPDATE` for entries, in addition to `CREATE`, `DROP` and `ALTER` for tables as a whole.
+
```
damithc, line 71: This section is generally good. Perhaps give two more SQL statements of different types, with a one sentence explanation of each?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,145 @@
...
+Now that you have seen how SQL is used, let's take a look at some of the main reasons behind the widespread adoption of SQL and why you should learn it.
+
+### Easy to learn
+
+As you saw in the example in the previous section, SQL is not really that complex. It is easy to learn, even for beginners who do not have any prior experience with databases. Since it has been around for a few decades, there are many good books and online resources to learn from.
+In addition, SQL (and SQL software) has a huge support community (such as [Stack Overflow](https://stackoverflow.com/questions/tagged/sql), and the [official MySQL forum](https://forums.mysql.com/)) which can prove useful when you run into problems while using it.
+
+### Free
+
+One of the benefits (and reasons) for SQL's popularity is that there is free software available (%like MySQL, PostgreSQL and SQLite%) as well as paid ones (%%like Microsoft SQL Server and Oracle%%) that come with more functionality.
```
damithc: You need `%%` to get dim text.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,145 @@
...
+
+One such example is MySQL. It is free of cost and comes with the official [MySQL Workbench](https://dev.mysql.com/doc/workbench/en/) as well as other third party easy-to-use <tooltip content="Graphical User Interface">GUIs</tooltip>, which are less daunting to new users as compared to a <tooltip content="Command Line Interface">CLI</tooltip>.
```
damithc: free of cost == free ?

damithc: This reminds me. Perhaps you can mention about GUI support and give a screenshot in the WHAT section? It can even show how you would run the example query and the output.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Highly Compatible
+
+SQL software works on many operating systems (%%like Windows, Mac OS, Ubuntu, Red Hat among others%%) and more importantly, can be integrated with many languages. This includes programming languages like C++, Ruby, Java, Python, JavaScript (Node.js), and Go.
+
+In addition, online playgrounds like [DB Fiddle](https://www.db-fiddle.com/) and [JDoodle](https://www.jdoodle.com/execute-sql-online) make it easy to use and learn the SQL language quickly without the hassle of setting up an environment or application.
```
damithc: Again, can show a screenshot  of a example to make things more visual. But not a must though.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Disadvantages
+
+SQL is not without its problems. In general, the biggest problem is with regards to the features of SQL.
+
+Although SQL databases use established <tooltip content="American National Standard Institutes">ANSI</tooltip> & <tooltip content="International Organization for Standardization">ISO</tooltip> standards, some SQL software systems (%%PostgreSQL, for example%%) add proprietary extensions to standard SQL.
+Due to this, the available feature set can vary according to what you're using. This can make SQL confusing and frustrating to use when switching SQL software.
```
damithc: when switching RDBMS?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,145 @@
...
+
+Apart from that, most **SQL** problems are not uniformly applicable across all SQL software. For example, MySQL suffers from concurrency issues. Though it performs well with read operations, it *can* be problematic when there are many concurrent read-write operations. A symptom of this issue would be a sudden slowdown of a query. However, PostgreSQL deals with concurrency well by implementing <tooltip content="each query transaction is isolated from other transactions running simultaneously in the system">transactions isolation</tooltip>.
+
```
damithc, line 115: Should we mention about non-relational databases too? Also, should we mention No-SQL?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,145 @@
...
+---
+
+## How to get started with SQL?
+
+The first thing to do when getting started with SQL is to choose a SQL software. You can either just use an online SQL playground or install a dedicated application. Our suggestion is to use either MySQL or SQLite online.
+
+SQL has been around for a long time and hence, you can find really good books and resources to learn it well. However, these resources can be overwhelming. So, you can decide which parts you want to learn, as you will likely not need to know everything to get started with using a SQL database.
```
damithc: The first sentence is repeated earlier in the article. Try to avoid repetition given that this is a short article.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,145 @@
...
+
+Here are some recommended steps for learning SQL:
+
+- Understand what a database and [<tooltip content="DataBase Management System">DBMS</tooltip>]({{baseUrl}}/contents/data/databases/databases.html) are. More specifically, understand the basic concepts of [RDBMS](https://www.tutorialspoint.com/sql/sql-rdbms-concepts.htm). This is not _essential_, but will give you a better high-level understanding before diving into programming.
```
damithc: To be more useful, you should point to where the reader can go to accomplish this step.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,145 @@
...
+
+- Khan Academy offers a very good free [SQL course](https://www.khanacademy.org/computing/computer-programming/sql). It uses a mix of videos, text, and exercises that you can choose to do at your own pace. It is well organized and even provides summary notes.
+
+- W3Schools offers an interactive written tutorial for [SQL](https://www.w3schools.com/sql/default.asp) in which you can run SQL queries to see the examples in action.
+
+- Once you're familiar with the language, you can make use of a [SQL Quick Reference Guide](https://www.w3schools.com/sql/sql_ref_mysql.asp) to look up information quickly.
+
+- You can also refer to [40 important SQL queries](https://bytescout.com/blog/20-important-sql-queries.html) to get a quick overview (or refresher) of SQL.
```
damithc: I don't think this resource is relevant to the intended target audience.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,145 @@
...
+
+---
+
+## Where to go from here?
```
damithc: It's hard to differentiate between this section and the section above. Merge and shorten?

damithc (changes requested):
Getting better now. Added some suggestions. Mostly nits. We'll probably need a couple of more rounds though.

contents/data/sql/SQL.md
amrut-prabhu: Hi Prof, you're right. I tried looking for some counter examples, but they're hard to come by.

However, [here](https://www.dnnsoftware.com/docs/developers/setup/set-up-sql.html) is something that I came across while doing my research. It is a little misleading, isn't it?

I've changed this tip box to be more generic rather than just pointing out SQL Server, since I feel the point is still valid

amrut-prabhu: Hi Prof, you're right. I tried looking for some counter examples, but they seem hard to come by.

However, [here](https://www.dnnsoftware.com/docs/developers/setup/set-up-sql.html) is something that I came across while doing my research. It is a little misleading, but it's a rare exception.
The previous paragraph does mention that there's no way to "install SQL" as such. So, I've removed the tip box.

contents/data/sql/SQL.md
amrut-prabhu: I feel that specifying that it is SQL syntax is useful, no?
I can remove the clauses/statements part

contents/data/sql/SQL.md
amrut-prabhu: Did you mean the HOW section (the one with the example)? 🤔

damithc: Sure, wherever it fits.

contents/data/sql/SQL.md
damithc: Try to minimize additional terminology introduced. Also, tip boxes attract a lot of attention due to their visual formatting. Don't use them for unimportant info. Consider less conspicuous means such as tooltips or dim text.

contents/data/sql/SQL.md
damithc: That is indeed misleading, but most likely an isolated incident. We shouldn't pollute our article due to such rare cases.

amrut-prabhu:
@damithc I've made the requested changes. Can you take another look?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,154 @@
...
+
+A Relational Database Management System (RDBMS) is a popular database solution used by software applications to manage their data.
+
+**Structured Query Language** (**SQL**) is a programming language that is specifically designed for interacting with an RDBMS. Unlike other languages, SQL doesn't come as a standalone installation.
+Rather, RDBMSs %%like MySQL, Oracle, SQL Server and PostgreSQL%% come with an _implementation_ of SQL. Note that the syntaxes may <tooltip content="in terms of whether the syntax is case-sensitive, the format for specifying date and time, what functions are available out-of-the-box">vary</tooltip> across these RDBMSs.
+
+These RDBMSs come with a GUI application, but can be accessed from the command line as well. Figure 1 shows MySQL Workbench GUI software and MySQL in the command line.
+
+<img src="MySQL_Workbench.png" alt="MySQL Workbench GUI Software" width="57%">
+<img src="MySQL_cli.png" alt="Running MySQL from the command line" width="40%">
```
damithc: The screenshot is too small; can make bigger. Also, the reader may appreciate the contents of the screenshot even more after they have gone through the query example. Perhaps you can mention (after the SQL query explanation) that there are multiple ways to run queries (GUI, CLI, API,) and show the screenshots there?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,154 @@
...
+| 1  | Alex  | CS202  | CS      |
+| 2  | Bob   | MA303  | MA      |
+| 3  | Cathy | CS202  | CS      |
+| 4  | Daren | CS202  | CS      |
+| 5  | Ellie | CS101  | CS      |
+| 6  | Fred  | MA303  | MA      |
+| 7  | Gary  | CS101  | CS      |
+| 8  | Henry | CS404  | CS      |
+
+We can use this SQL query to retrieve information from this table:
```
damithc, line 46: Good job using SQL syntax highlighting. I forgot about that totally.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,154 @@
...
+| ------ | --- |
+| CS202  | 3   |
+| CS101  | 2   |
+
+<br/>
+You can see how this simple query can prove to be extremely useful for getting this information when the table has tens of entries (or many more). Apart from retrieving information, SQL can also be used for creating, deleting and manipulating data with queries like `INSERT`, `DELETE` and `UPDATE` for entries, in addition to `CREATE`, `DROP` and `ALTER` for tables as a whole.
+
+<br/>
+For example, you can set up the structure of table shown in Table 1 by executing the `CREATE TABLE` query shown here.
+It creates a new data table `Students`, with 4 fields (`ID`, `Name`, `Course` and `Faculty`) and specifies their data types.
```
damithc, line 70: No need for `<br/>`? Anyway, you need a blank line before markdown content. Without it, this paragraph doesn't render as expected.
image

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,154 @@
...
+
+Apart from that, most **SQL** problems are not uniformly applicable across all RDBMSs that use SQL. For example, MySQL suffers from concurrency issues. Though it performs well with read operations, it *can* be problematic when there are many concurrent read-write operations. A symptom of this issue would be a sudden slowdown of a query. However, PostgreSQL deals with concurrency well by implementing <tooltip content="each query transaction is isolated from other transactions running simultaneously in the system">transactions isolation</tooltip>.
+
+Depending on your requirements, you can also choose to use non-relational databases like <tooltip content="Stands for- not only SQL">NoSQL</tooltip> databases instead of RDBMSs (and SQL). They are known for being highly scalable and handling large volumes of data well.
+
+---
+
+## How to get started with SQL?
+
+The first thing to do when getting started with SQL is to choose a RDBMS. You can either just use an online SQL playground or install a dedicated application. Our suggestion is to use either MySQL or SQLite online.
```
damithc: `MySQL or SQLite online` --> link to respective pages?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,154 @@
...
+
+The extensive number of books and resources available for learning SQL can seem overwhelming. To overcome this issue, you can decide on which parts you want to learn, as you will likely not need to know everything to get started with using a SQL database.
+
+Here are some recommended steps for learning SQL:
+
+- **Understand RDBMS**: Before jumping into the SQL language, refer to this book's [<tooltip content="DataBase Management System">**DBMS**</tooltip>]({{baseUrl}}/contents/data/databases/databases.html) chapter to understand the basic database concepts. More specifically, understand the basic concepts of RDBMS by going through the [**SQL- RDBMS Concepts**](https://www.tutorialspoint.com/sql/sql-rdbms-concepts.htm) page. This is not _essential_, but will give you a better high-level understanding before diving into programming.
```
damithc: As these are steps, maybe a numbered list is more appropriate?

damithc (changes requested):
Looks to be in near-LGTM. Good work. 👍  Just a few nits to fix.

contents/data/sql/SQL.md
amrut-prabhu: Ooh... I was wondering why that wasn't rendering properly. Thanks!

amrut-prabhu:
@damithc Can you take another look?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,164 @@
...
+
+```sql
+DELETE
+FROM Students
+WHERE Faculty='MA';
+```
+
+<box type="info">
+  You can experiment with this example on <a href="https://www.db-fiddle.com/f/kHqV2edUGxCc1dU6vE6CmS/1">DB Fiddle</a> (shown in Figure 1 below) by entering SQL queries and then running them.
+</box>
```
damithc, line 86: It is more meaningful if `this example on` is hyperlinked as well?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,164 @@
...
+
+<center>
+<img src="DBFiddle.png" alt="Running queries on DB Fiddle" width="90%">  
+
+_Figure 1. Using DB Fiddle to run queries in MySQL_
+</center>
+
+Apart from using **online editors** like DB Fiddle to execute SQL queries, you can also install and use a RDBMS. It usually comes with a **software application**, but can also be accessed from the **command line**. Figure 2 shows the MySQL Workbench application, and Figure 3 shows how to use the MySQL RDBMS in the command line. Finally, **APIs** in other programming languages, or those like [SQL API](https://carto.com/developers/sql-api/) (shown in Figure 4) can also be used to execute SQL queries.
```
damithc: `It usually comes with a **software application**`: did you mean _comes with a GUI_?

damithc: `Apart from using **online editors** like DB Fiddle to execute SQL queries, you can also install and use a RDBMS.` This makes it sounds like they are alternatives. They are alternatives only in the context of learning SQL, not real use.

damithc: Why not put each explanation together with the corresponding image?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,164 @@
...
+---
+
+## Why learn SQL?
+
+Now that you have seen how SQL is used, let's take a look at some of the main reasons behind the widespread adoption of SQL and why you should learn it.
+
+### Easy to learn
+
+As you saw in the example in the previous section, SQL is not really that complex. It is easy to learn, even for beginners who do not have any prior experience with databases. Since it has been around for a few decades, there are many good books and online resources to learn from.
+In addition, SQL and RDBMSs have a huge support community (such as [SQL in Stack Overflow](https://stackoverflow.com/questions/tagged/sql), and the [official MySQL forum](https://forums.mysql.com/)) which can prove useful when you run into problems while using it.
```
damithc: `while using it` -> `while using SQL`? (it's not clear what `it` refers to)

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Free
+
+One of the benefits (and reasons) for SQL's popularity is that there are free RDBMS software available %%(like MySQL, PostgreSQL and SQLite)%% as well as paid ones %%(like Microsoft SQL Server and Oracle)%% that come with more functionality.
```
damithc: This use of `as well as` reduces clarity of the statement. Break into multiple sentences? The point here should be 'there are many free options that are powerful enough for most situations', not 'commercial options having more functionality'.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,164 @@
...
+
+One such example is MySQL. It is free and comes with the official [MySQL Workbench](https://dev.mysql.com/doc/workbench/en/) (shown in Figure 2 above) as well as other third party easy-to-use <tooltip content="Graphical User Interface">GUIs</tooltip>, which are less daunting to new users as compared to a <tooltip content="Command Line Interface">CLI</tooltip>.
```
damithc: >which are less daunting to new users as compared to a <tooltip content="Command Line Interface">CLI</tooltip>

This point doesn't belong under this heading.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Highly Compatible
+
+The RDBMSs that use SQL work on many operating systems %%(like Windows, Mac OS, Ubuntu, Red Hat among others)%% and more importantly, can be integrated with many languages. This includes programming languages like C++, Ruby, Java, Python, JavaScript (Node.js), Go, and more.
+
+In addition, online playgrounds like [DB Fiddle](https://www.db-fiddle.com/) (shown in Figure 1 above) and [SqliteOnline](https://sqliteonline.com/) make it easy to use and learn the SQL language quickly without the hassle of setting up an environment or application.
```
damithc: This point belongs under 'easy to learn'?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Free
+
+One of the benefits (and reasons) for SQL's popularity is that there are free RDBMS software available %%(like MySQL, PostgreSQL and SQLite)%% as well as paid ones %%(like Microsoft SQL Server and Oracle)%% that come with more functionality.
+
+One such example is MySQL. It is free and comes with the official [MySQL Workbench](https://dev.mysql.com/doc/workbench/en/) (shown in Figure 2 above) as well as other third party easy-to-use <tooltip content="Graphical User Interface">GUIs</tooltip>, which are less daunting to new users as compared to a <tooltip content="Command Line Interface">CLI</tooltip>.
+
+### Highly Compatible
+
+The RDBMSs that use SQL work on many operating systems %%(like Windows, Mac OS, Ubuntu, Red Hat among others)%% and more importantly, can be integrated with many languages. This includes programming languages like C++, Ruby, Java, Python, JavaScript (Node.js), Go, and more.
```
damithc: This point seems a bit convoluted. I think a better reason for learning SQL is that it is widely used and not limited to any OS or programming language. Anyway, `Widely used` should be the first reason for learning SQL as it is a stronger motivation than all others?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,164 @@
...
+
+In addition, online playgrounds like [DB Fiddle](https://www.db-fiddle.com/) (shown in Figure 1 above) and [SqliteOnline](https://sqliteonline.com/) make it easy to use and learn the SQL language quickly without the hassle of setting up an environment or application.
+
+## Disadvantages
```
damithc: Wait. Are we taking about disadvantages of RDBM or SQL?

Talking of which, go back to WHY section and revise it to make it about SQL, not RDBMS. I can think of only one reason to include there: because many apps use RDBMS and hence you need to know SQL to interact with the database. The fact that it is easy to learn and there are free options to get started can be mentioned in passing in the _How to get started section_. What do you think?

damithc (changes requested):
Added a few more things to consider. The main thing is not to stray into the RDBMS topic and stay within the actual topic (i.e., SQL).

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,151 @@
...
+Then, it removes courses that do not have more than 1 student, i.e., `CS404` is removed from consideration.
+Finally, it returns a list of courses with a count of the number of students, ordered in increasing order.
+So, the output of the query is:
+
+| Course | num |
+| ------ | --- |
+| CS101  | 2   |
+| CS202  | 3   |
+
+You can see how this simple query can prove to be extremely useful for getting this information when the table has tens of entries (or many more). Apart from retrieving information, SQL can also be used for creating, deleting and manipulating data with queries like `INSERT`, `DELETE` and `UPDATE` for entries, in addition to `CREATE`, `DROP` and `ALTER` for tables as a whole.
```
damithc: `tens of entries (or many more)` -> `a large number of rows`?

`with queries like`: are these called 'queries'? feels more like directives or statements.

Also consider mentioning that a query can span multiple tables.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,151 @@
...
+
+<center>
+<img src="DBFiddle.png" alt="Running queries on DB Fiddle" width="90%">  
+
+_Figure 1. Using DB Fiddle to run queries in MySQL_
+</center>
+
+**Online editors** like DB Fiddle can be used while learning SQL. However, to use SQL in application development, you need to consider other alternatives for running SQL queries:
+
+- **RDBMS software**: RDBMS installations usually come with a <tooltip content="Graphical User Interface">GUI</tooltip> (see Figure 2), but can also be accessed from the command line (see Figure 3).
```
damithc, line 96: Feels like the images below can be made a bit larger so that the text in their matches the text size of the main body of the article.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,151 @@
...
+
+    <center>
+    <img src="MySQL_Workbench.png" alt="MySQL Workbench GUI Software" width="85%">  
+
+    _Figure 2. MySQL (an RDBMS) can be used in the MySQL Workbench GUI application_ <sup>[source](https://www.mysql.com/products/workbench/)</sup>
```
damithc: >_Figure 2. MySQL (an RDBMS) can be used in the MySQL Workbench GUI application_ ([source](https://www.mysql.com/products/workbench/))

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,151 @@
...
+
+Reviewers: [Ronak Lakhotia](https://github.com/RonakLakhotia), [Rahul Rajesh](https://github.com/rrtheonlyone)
+
+## What is SQL?
+
+A Relational Database Management System (RDBMS) is a popular database solution used by software applications to manage their data.
+
+**Structured Query Language** (**SQL**) is a programming language that is specifically designed for interacting with an RDBMS. Unlike other languages, SQL doesn't come as a standalone installation.
+Rather, RDBMSs %%like MySQL, Oracle, SQL Server and PostgreSQL%% come with an _implementation_ of SQL. Note that the syntaxes may <tooltip content="in terms of whether the syntax is case-sensitive, the format for specifying date and time, what functions are available out-of-the-box">vary</tooltip> across these RDBMSs.
+
```
damithc, line 26: >Note that the syntaxes may vary across these RDBMSs.

Perhaps no need to mention this here as it is mentioned again later?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,151 @@
...
+These factors can make SQL confusing and frustrating to use when switching across RDBMSs, since you would have to change your SQL code.
+
+---
+
+## How to get started with SQL?
+
+As you saw in the example in the previous section, SQL is not really that complex. It is easy to learn, even for beginners who do not have any prior experience with databases.
+
+To get started with SQL, you can either just use an online SQL playground or install a RDBMS. There are many free RDBMSs available like [PostgreSQL](https://www.postgresql.org/) and [MySQL](https://dev.mysql.com/downloads/).
+However, for the sole purpose of learning SQL, our suggestion is to use online editors like [DB Fiddle](https://www.db-fiddle.com/) %%(supports MySQL and SQLite)%% or [SqliteOnline](https://sqliteonline.com/) %%(supports SQLite)%% since they allow you to start learning without the hassle of setting up anything.
```
damithc: >However, for the sole purpose of learning SQL, our suggestion is to use online editors

I wouldn't go as far as to _recommend_ it. Something like this? `At the early stages of learning SQL, you can even use online editors ...`

damithc (changes requested):
Looks good. Some cosmetic nits.

amrut-prabhu:
@damithc I've made the requested changes

damithc:
Looks good @amrut-prabhu 
@RonakLakhotia @rrtheonlyone can do one last check as there were significant changes since your last review? Look for micro-level nits only. I don't think we want to do any further major structural changes at this point.

damithc (approved):

damithc:
@amrut-prabhu remind me to merge after the other two reviewers have approved the latest version.

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,151 @@
...
+Then, it removes courses that do not have more than 1 student, i.e., `CS404` is removed from consideration.
+Finally, it returns a list of courses with a count of the number of students, ordered in increasing order.
+So, the output of the query is:
+
+| Course | num |
+| ------ | --- |
+| CS101  | 2   |
+| CS202  | 3   |
+
+You can see how this simple query can prove to be extremely useful for getting this information when the table has a large number of rows. Queries can also be used to retrieve data that span across multiple tables. Apart from retrieving information, SQL can used for creating, deleting and manipulating data with commands like `INSERT`, `DELETE` and `UPDATE` for entries, in addition to `CREATE`, `DROP` and `ALTER` for tables as a whole.
```
RonakLakhotia: sql can be used?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,151 @@
...
+    <img src="SQL_API.png" alt="Using SQL API to execute queries" width="95%">  
+
+    _Figure 4. An example of how to use CARTO's SQL API, which uses a PostgreSQL database_ ([source](https://carto.com/developers/sql-api/))  
+    </center>
+
+---
+
+## Why learn SQL?
+
+The main reason why you should learn SQL relates to the widespread adoption of SQL. As SQL is not limited to any particular operating system or programming language, it is **widely used** in the RDBMSs used by a lot of web, mobile and other software applications. Hence, in order to utilize almost any relational database or RDBMS available today, you need to know the SQL language.
```
RonakLakhotia: >it is **widely used** in the RDBMSs used by a lot of web, mobile and other software applications

very minor, but you can consider rephrasing this?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,151 @@
...
+
+    - In order to keep track of and remember the syntaxes that you're learning, you can refer to the [**W3Schools SQL Quick Reference Guide**](https://www.w3schools.com/sql/sql_ref_mysql.asp). This is particularly useful when you want to look up specific information quickly.
+
+1. **Use SQL in practice**: Look into different RDBMSs and decide which one you want to work with. You can start by comparing the 3 mentioned in this [**Oracle vs MySQL vs SQL Server**](https://blog.udemy.com/oracle-vs-mysql-vs-sql-server/) post and work from there. Then, set up the corresponding RDBMS and integrate it with an application. This will give you good experience and exposure to how databases are used in practice.
+
+    In the case of MySQL, a good resource for getting started with this is [**MySQLTutorial.org**](http://www.mysqltutorial.org/). The `Interfaces` section of the website contains multiple tutorials explaining how to integrate and work with MySQL from [Node.js](http://www.mysqltutorial.org/mysql-nodejs/), [Java](http://www.mysqltutorial.org/mysql-jdbc-tutorial/), [Python](http://www.mysqltutorial.org/python-mysql/) and more.
```
RonakLakhotia, line 152: >In the case of MySQL, a good resource for getting started with this is 

to 
a good resource for getting started with MySQL is MySQLTutorial.org?

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,151 @@
...
+    - The [**Guru99 SQL Tutorial for Beginners**](https://www.guru99.com/sql.html) is a good course for beginners. It is well organized and provides comprehensive information, which will ensure that you know the basics well.
+
+    - The [**W3Schools SQL Tutorial**](https://www.w3schools.com/sql/default.asp) is a useful, interactive written tutorial in which you can run and modify SQL queries to see the examples in action.
+
+    - In order to keep track of and remember the syntaxes that you're learning, you can refer to the [**W3Schools SQL Quick Reference Guide**](https://www.w3schools.com/sql/sql_ref_mysql.asp). This is particularly useful when you want to look up specific information quickly.
+
+1. **Use SQL in practice**: Look into different RDBMSs and decide which one you want to work with. You can start by comparing the 3 mentioned in this [**Oracle vs MySQL vs SQL Server**](https://blog.udemy.com/oracle-vs-mysql-vs-sql-server/) post and work from there. Then, set up the corresponding RDBMS and integrate it with an application. This will give you good experience and exposure to how databases are used in practice.
```
RonakLakhotia, line 150: >give you a good experience?

RonakLakhotia (commented):
Just some very minor points. Feel free to ignore them :)

contents/data/sql/SQL.md
amrut-prabhu: I think the current one is better. We're just giving an example, rather than saying that you should choose MySQL.

contents/data/sql/SQL.md
amrut-prabhu: I feel that it's fine as is. `A good experience` sounds like you will have a good time using it. Just `good experience` (the current one) means you will get practical knowledge on how to use SQL in apps.

RonakLakhotia:
LGTM!

amrut-prabhu:
@rrtheonlyone Still awaiting your review :)

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,152 @@
...
+    <img src="SQL_API.png" alt="Using SQL API to execute queries" width="95%">  
+
+    _Figure 4. An example of how to use CARTO's SQL API, which uses a PostgreSQL database_ ([source](https://carto.com/developers/sql-api/))  
+    </center>
+
+---
+
+## Why learn SQL?
+
+The main reason why you should learn SQL relates to the widespread adoption of SQL. As SQL is not limited to any particular operating system or programming language, RDBMSs (which use SQL)  are **widely used** by a lot of software applications %%(web, mobile etc.)%%. Hence, in order to utilize almost any relational database or RDBMS available today, you need to know the SQL language.
```
rrtheonlyone: Feel like you could rephrase the first line. It sounds off...

"The main reason why you should learn SQL relates to the widespread adoption of SQL"

contents/data/sql/SQL.md
```diff
@@ -0,0 +1,152 @@
...
+Although SQL databases use established <tooltip content="American National Standard Institutes">ANSI</tooltip> & <tooltip content="International Organization for Standardization">ISO</tooltip> standards, some RDBMSs %%(like PostgreSQL, for example)%% add proprietary extensions to standard SQL.
+Due to this, the available feature set can vary according to what you're using.
+Furthermore, there may be <tooltip content="case-sensitivity, date and time format, functions available out-of-the-box, etc.">minor changes in the syntaxes</tooltip> across RDBMSs.
+These factors can make SQL confusing and frustrating to use when switching across RDBMSs, since you would have to change your SQL queries.
+
+---
+
+## How to get started with SQL?
+
+As you saw in the example in the previous section, SQL is not really that complex. It is easy to learn, even for beginners who do not have any prior experience with databases.
```
rrtheonlyone: SQL is really not that complex?

"really not" instead of "not really"

rrtheonlyone (approved):
I am good with the article.

It serves as a good introduction to SQL especially for beginners. I have some minor nits (just suggestions)

(Also while I was reading this, I noticed you have used the acronym 'RDBMS' quite excessively :P)

amrut-prabhu:
> @amrut-prabhu remind me to merge after the other two reviewers have approved the latest version.

@damithc Ready to merge

damithc:
Merged and deployed. Good work @amrut-prabhu 🎉 
Thanks to reviewers too.

PR #91, opened 2019-03-04
Add page on Linux shell

junming403:
Fixes #91


junming403:
Looking for reviews

damithc:
A quick comment: use `we` instead of `I` as more authors can join later.

damithc:
Another thing: cite the source when using assets from elsewhere.

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+
+<div class="website-content">
+
+# Introduction to Linux bash shell 
+
+Author: [Wang Junming](https://github.com/junming403)
+
+## Overview
+
+This chapter assumes that the reader is familiar with general concepts of operating systems. If you are not familiar with basic operating system concepts, I recommend you pick up the basic concepts first. The classic textbook of operating system **Mordern Operating Systems by Tanenbaum** is highly recommended.
```
rrtheonlyone: Do we really need that much familiarity with Operating Systems to understand how to use the basic bash shell?

The textbook is used for introductory courses such as Introduction to Operating Systems. Seems a bit much here to read that before coming here?

Why not you just link a simple article or something?

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+DESCRIPTION
+     The grep utility searches any given input files, selecting lines that
+     match one or more patterns.  By default, a pattern matches an input line
+     if the regular expression (RE) in the pattern matches the input line
+     without its trailing newline.  An empty expression matches every line.
+     Each input line that matches at least one of the patterns is written to
+     the standard output ......
+
+```
+## How to Get Started with linux command-line based shell?
```
rrtheonlyone: This part here just feels like your saying, 'jump right in and start using it' !

I liked your git example here! Maybe you can elaborate a bit more on specific use cases when I should turn to the command line?

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+<img src="images/shell-kernel.png" alt="shell-kernel" style="width:400px;"/>
+
+Computer can be divided into 3 abstraction layers, hardware, kernel and applications. Users cannot control the hardware directly, instructions have to be given through kernel as kernel is the one that controls the hardware. 
+Shell is the interface through which we can pass instructions to kernel, and then it could be executed on the hardware. In this chapter, we will look into **Linux command-line based shell(bash)**.
+
+> Linux is the best-known and most-used open source operating system. As an operating system, Linux is software that sits underneath all of the other software on a computer, receiving requests from those programs and relaying these requests to the computer’s hardware.
+
+[**bash**](https://www.gnu.org/software/bash/manual/html_node/) stands for **Bourne Again SHell**, an enhanced version of the original Unix Shell program. It has become the standard shell of various linux distributions. 
+
+## Why use command-line based shell?
```
rrtheonlyone: This section seems very technical! You could also add some points to convince total beginners to switch over to bash? People who have zero experience with command line may not be able to follow along with the code snippets you have put up.

rrtheonlyone (changes requested):
Overall, I feel like your article does a good job explaining the bash shell process.

However, my opinion is definitely biased as I am very familiar with bash (some of the concepts you mentioned may be overwhelming for a complete beginner?) 

Also, there are plenty of grammatical errors in your article :P You may want to read it through one more time and get rid of those!

damithc:
> Also, there are plenty of grammatical errors in your article :P You may want to read it through one more time and get rid of those!

You can also ask someone else in the class (who has a stronger command of the language) to help point out these.

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+#!/bin/bash
+./calculate < DataBundle > result 2> errors
+if [ -s errors]
+then
+     ./verify < result
+else
+     echo "error occured during execution."
+fi
+```
+Another useful command is `|`, which chained linux commands together, as such the output of previous commands is passed as input to the next command.
```
ongspxm: one nice thing to mention in here could be `tee`, and since you used the different streams here, might wanna put some link to it with more explanation?

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+
+Suppose you want to check the status of process `p` running on your system. Thus you typed in `ps aux`, and all the processes are listed, but there are too many processes that you can't find `p`. As such, you can pass the result of `ps aux` to `grep 'p'` to capture process `p`'s status. Thus the command will be look like `ps aux | grep 'p'`.
+
+Input/output streams can easily be redirected, allowing information to be sent or received from files or other applications(commands and user programs). This can mean test data can be easily supplied or output captured.
```
ongspxm: can mention list of common applications used, `awk`, `sed`, etc. so that they know what to start looking for?

damithc:
Some quick comments:
* Resize/tweak images so that their text size matches the normal text in the article
* Cite the source of images reused from elsewhere
* Lengthy code examples might scare off the readers

contents/linux/shell.md
ChooJeremy: Agree. This article isn't even for teaching how to use the bash shell - it's why one should use the bash shell.  I think that if the person is still at that stage, knowledge at this level should not be required.

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+
+Author: [Wang Junming](https://github.com/junming403)
+
+## Overview
+
+This chapter assumes that the reader is familiar with general concepts of operating systems. If you are not familiar with basic operating system concepts, I recommend you pick up the basic concepts first. The classic textbook of operating system **Mordern Operating Systems by Tanenbaum** is highly recommended. 
+
+<img src="images/shell-kernel.png" alt="shell-kernel" style="width:400px;"/>
+
+Computer can be divided into 3 abstraction layers, hardware, kernel and applications. Users cannot control the hardware directly, instructions have to be given through kernel as kernel is the one that controls the hardware.
```
ChooJeremy: Is this explanation necessary for understanding the article? It feels too complex for the target audience, using terms that may be unfamiliar to them - abstraction, kernel etc.

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+
+> Linux is the best-known and most-used open source operating system. As an operating system, Linux is software that sits underneath all of the other software on a computer, receiving requests from those programs and relaying these requests to the computer’s hardware.
+
+[**bash**](https://www.gnu.org/software/bash/manual/html_node/) stands for **Bourne Again SHell**, an enhanced version of the original Unix Shell program. It has become the standard shell of various linux distributions. 
+
+## Why use command-line based shell?
+
+A common question many people would ask is why would we learn command-line based shell as it seems to be very hard to use. There are tons of GUI tools out there, they seems to look more user-friendly and powerful. From my perspective, there are four major reasons for starting to use command line.
+
+- **Remote Management**: Use over remote command-line shell makes it much easier to perform tasks remotely, such as when a developer is connecting from home. A remote command-line shell usually requires less bandwidth.
```
ChooJeremy: If I'm not familiar with remote shells, I could argue that the solutions that I'm using work too - say, teamviewer to perform remote control. Arguably easier with a GUI too. The only point here that beats it is less bandwidth. Thus, it feels pretty weak - perhaps mention it's better supported with many other ssh programs, require less cpu usage for the connection, etc?

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+Shell is the interface through which we can pass instructions to kernel, and then it could be executed on the hardware. In this chapter, we will look into **Linux command-line based shell(bash)**.
+
+> Linux is the best-known and most-used open source operating system. As an operating system, Linux is software that sits underneath all of the other software on a computer, receiving requests from those programs and relaying these requests to the computer’s hardware.
+
+[**bash**](https://www.gnu.org/software/bash/manual/html_node/) stands for **Bourne Again SHell**, an enhanced version of the original Unix Shell program. It has become the standard shell of various linux distributions. 
+
+## Why use command-line based shell?
+
+A common question many people would ask is why would we learn command-line based shell as it seems to be very hard to use. There are tons of GUI tools out there, they seems to look more user-friendly and powerful. From my perspective, there are four major reasons for starting to use command line.
```
ChooJeremy: Expand it out, it's the first time you're using the phrase.
```suggestion
A common question many people would ask is why would we learn command-line based shell as it seems to be very hard to use. There are tons of Graphical User Interface (GUI) tools out there, they seems to look more user-friendly and powerful. From my perspective, there are four major reasons for starting to use command line.
```

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+
+- **Remote Management**: Use over remote command-line shell makes it much easier to perform tasks remotely, such as when a developer is connecting from home. A remote command-line shell usually requires less bandwidth.
+
+- **Stability**: No matter what linux distribution you are running, the command line shell is always available and usage is almost the same. However, GUI tools might be different in different distributions, make it very hard to learn.
+
+- **Freedom**: You will have complete control in command-line shell. GUIs usually tends to simplifying things, giving user fewer options to make things easier. In command line there is no such simplifications, you will have complete control of what you are doing.
```
ChooJeremy: Rephrase + grammer
```suggestion
- **Freedom**: You will have complete control in the command-line shell. GUIs tends to simplify things, giving the user fewer options. However, the command line always provides the full suite of options, giving you complete control.```

File: contents/linux/shell.md [hunk 10]
```diff
@@ -0,0 +1,180 @@
...
+
+- **Simplicity**: Although you might find command-line hard to use at first, but after using it enough you'll eventually come to the point where the command-line is far easier to use than trying to find options in the drop-down menus and alike of a GUI.
```
ChooJeremy: Not sure if better.  Just a thought. The original feels like run on sentence though.
```suggestion
- **Simplicity**: While you might find command-line difficult to grasp, eventual usage will familiarize yourself such that the command-line is far easier to use than trying to find options in the drop-down menus and alike of a GUI. 
```

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+
+- **Stability**: No matter what linux distribution you are running, the command line shell is always available and usage is almost the same. However, GUI tools might be different in different distributions, make it very hard to learn.
+
+- **Freedom**: You will have complete control in command-line shell. GUIs usually tends to simplifying things, giving user fewer options to make things easier. In command line there is no such simplifications, you will have complete control of what you are doing.
+
+- **Simplicity**: Although you might find command-line hard to use at first, but after using it enough you'll eventually come to the point where the command-line is far easier to use than trying to find options in the drop-down menus and alike of a GUI. 
+
+### Benefit: Program with bash script
+
+One important feature in bash is **bash script**. It enables us to do `task automation`. Suppose you are a system admin, there is a task you need to perform daily without any changes. Instead of typing in the set of commands for the task everyday, you can store these commands in a file and run the script.
```
ChooJeremy: Perhaps
```suggestion
One important feature in bash is the **bash script**. It enables us to do `task automation`. For example, suppose that you are a system admin and there is a task you need to perform daily. Instead of typing in the commands for the task daily, you can store these commands in the **bash script** and run it.
```

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+
+Here's one example of bash script, suppose you have to check if some given file exist periodically. The following script will ask for your input and output the results. `filename` is a variable used to store user input, `filetype` and `perm` stores corresponding attributes of the file for output.
```
ChooJeremy:
```suggestion
For example, suppose you have to check if some file exists periodically. The following script will ask for your input and output the results. `filename` is a variable used to store user input, `filetype` and `perm` stores corresponding attributes of the file for output.
```

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+test -r $filename && perm="readable"
+test -w $filename && perm="$perm writable"
+test -x $filename && perm="$perm executable"
+
+# output result.
+echo "The filename: $filename is a $filetype"
+echo "And permissions are : $perm"
+```
+
+The above bash script is still just a collection of bash commands. Although it saves us a lot of works if it is a daily routine task, it still can be done by typing those commands line by line. What makes bash script really exciting is the semantics of `conditionals`, `loops` and `functions`. Which makes bash script [**turning complete**](https://en.wikipedia.org/wiki/Turing_completeness).
```
ChooJeremy: Grammar. Maybe bold bash script?
```suggestion
The above bash script is still just a collection of bash commands. Although it saves us a lot of work if it is a routine task, it can still be done by typing those commands line by line. What makes bash script really exciting is the semantics of `conditionals`, `loops` and `functions`. Which makes bash script [**turning complete**](https://en.wikipedia.org/wiki/Turing_completeness).
```

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+
+The following is an example of bash script that make use of conditionals and loops. Suppose you want to let user input a directory name, check if it exists, and then output permissions for all files inside that directory.
```
ChooJeremy: Grammar
```suggestion
The following is another example of bash script that utilizes conditionals and loops. Suppose you want to allow the user to input a directory name, check if it exists and output permissions for all files inside that directory.
```

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+	perm=""
+	test -r "$dir/$filename" && perm="readable"
+	test -w "$dir/$filename" && perm="$perm writable"
+	test -x "$dir/$filename" && perm="$perm executable"
+	echo "The file $dir/$filename 's permission is $perm" 
+done
+```
+**Functions** in bash script just like functions in normal programming languages, more semantics of bash script can be found [here](https://www.gnu.org/software/bash/manual/html_node/). 
+
+bash script allows us to automate the frequently performed operations, it is easy to use and can be executed on any Unix-like operating systems without any modifications.
```
ChooJeremy: Perhaps something like "In conclusion," or "In summary,"

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Benefit: Data Stream Redirection
+
+In linux, everything is abstracted as file, including **input**, **output**, and **error** streams. **Redirection** is a feature in Linux such that when executing a command/user program, you can change the standard input/output devices by changing its [file descriptor](https://en.wikipedia.org/wiki/File_descriptor). Each command and user program in linux is an application with input, output and errors. We can redirect the output of one application to the input of another, as such, we combined the two applications together as if we build a pipe between them.
```
ChooJeremy: Grammar + suggested change near the end.
```suggestion
In linux, everything is abstracted as a file, including **input**, **output**, and **error** streams. **Redirection** is a feature in Linux such that when executing a command/user program, you can change the standard input/output devices by changing its [file descriptor](https://en.wikipedia.org/wiki/File_descriptor). Each command and user program in linux is an application with input, output and error streams. A common use case is to redirect the output of one application to the input of another, combining the two applications together as if there was pipe between them.
```

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+
+![file-descriptor](images/file-descriptor-illustration.jpg)
+
+For example, suppose you just finished your coding assignment, the resulting executable `calculate` takes a user input, do the calculation, and gives the output. To run it, you simply types:
```
ChooJeremy: Interesting, here you use "for example" but above it was "The following is an example of..." 
:P 
Would replace do with "performs"

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+
+```bash
+./calculate
+```
+
+But you are not satisfied with manual testing, instead, you want to test with larger data set file `DataBundle`. And verify the output with another executable `verify`, which takes in the result as input and verify its correctness and output **PASS** or **FAIL**. Here's how we can do it.
```
ChooJeremy:
```suggestion
But you are not satisfied with manual testing, instead, you want to test with a larger data set file `DataBundle` and verify the output with another executable `verify`. `verify` takes in the result as input and verifies its correctness. It outputs **PASS** or **FAIL** depending on if it passes or fails. Here's how we can do it.
```

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+
+```bash
+./calculate < DataBundle > result
+./verify < result
+```
+Notice in this case, `calculate` uses the data in `DataBundle` as input, and output the results to the file `result`. Then, `result` is taken as input to `verify`, and the final verification result is printed on screen.
+
+Furthermore, if the code we write is buggy, it throws exception and error messages are directed to error output, then our result file will be empty! To handle it properly, we can also redirect error output to a file `errors`, and verify it is empty(such that no error occurs during execution) before we execute `verify`.
```
ChooJeremy:
```suggestion
However, if the code we write is buggy, any thrown exception and error messages are directed to error output, resulting in an empty result file! To handle this scenario, we can also redirect error output to a file `errors`, and verify it is empty (meaning that no error occurred during execution) before we execute `verify`.
```

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+if [ -s errors]
+then
+     ./verify < result
+else
+     echo "error occured during execution."
+fi
+```
+Another useful command is `|`, which chained linux commands together, as such the output of previous commands is passed as input to the next command.
+
+Suppose you want to check the status of process `p` running on your system. Thus you typed in `ps aux`, and all the processes are listed, but there are too many processes that you can't find `p`. As such, you can pass the result of `ps aux` to `grep 'p'` to capture process `p`'s status. Thus the command will be look like `ps aux | grep 'p'`.
```
ChooJeremy: You should mention briefly what ps and grep does.

contents/linux/shell.md
ChooJeremy:
```suggestion
Input/output streams can be easily redirected, allowing information to be sent or received from files or other applications (commands and user programs). This means that test data can be easily supplied or output captured to a file.
```

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+     The grep utility searches any given input files, selecting lines that
+     match one or more patterns.  By default, a pattern matches an input line
+     if the regular expression (RE) in the pattern matches the input line
+     without its trailing newline.  An empty expression matches every line.
+     Each input line that matches at least one of the patterns is written to
+     the standard output ......
+
+```
+## How to Get Started with linux command-line based shell?
+In my opinion, there is no need to learning linux command-line based shell in particular, because there are tons of linux commands out there, it is impossible to remember and master them all at one time without actual practices. Instead, when it comes to the time that you have to use it, try to use the terminal instead of GUI based tool. Although it might be hard at first, it will benefit you more later. For example, `git`, I highly recommend get started to use command-line based git instead of visualisation tools such as `Source Tree`.
```
ChooJeremy: Simplified some text
```suggestion
In my opinion, there is no need to learning linux command-line based shell in particular, because there are tons of linux commands out there, it is impossible to remember and master them all at once without actual practice. Instead, when it comes time to use it, try to use the terminal instead of GUI based tool. Although it might be hard at first, it will benefit you more later. For example, `git`, I highly recommend get started to use command-line based git instead of visualisation tools such as `Source Tree`.
```

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+
+
+# Where to Go from Here?
```
ChooJeremy: Perhaps a link to start to a tutorial to start learning shell commands from the very basics? Like ls, cd, mkdir, mv, cp etc.

ChooJeremy (changes requested):
I originally thought that this article would be on learning how to use the linux shell, but it's really just why you should start learning. I might be the only one misinterpreting the contents based on the tile and first few lines, but if others are misinterpreting it that way too you might want to make some changes there.

contents/linux/shell.md
ChooJeremy: More thoughts 
```suggestion
A common question many people ask is why should we learn command-line based shell as it seems to be very hard to use. There are tons of Graphical User Interface (GUI) tools out there, that appear to be more user-friendly and powerful. From my perspective, there are four major reasons to use the command line:
```

_markbind/navigation/mainNav.md
```diff
@@ -72,6 +72,8 @@
     * [Cryptography]({{baseUrl}}/contents/security/cryptography.html)
     * [HTTPS]({{baseUrl}}/contents/security/Https.html)
     * [SQL Injection]({{baseUrl}}/contents/security/sqlInjection.html)
+  * Linux 
+    * [Shell]({{baseUrl}}/contents/linux/shell.html)
```
Xenonym: Could I propose that this article be classified under Shells/Bash instead? The current classification suggests that shells only exist on Linux.

Reasons:
- Bash is [not the only Linux shell](https://www.slant.co/topics/513/~best-unix-shells). Colloquially, `zsh`/`dash` does see use.
- Bash does not only exist on Linux. It is also the default shell in Mac OS X and can be used on Windows through [MSYS2](https://www.msys2.org/) (used in Git Bash) or [Windows Subsytem for Linux](https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux).
- Windows has shells too, in the form of [Command Prompt](https://en.wikipedia.org/wiki/Cmd.exe) and [PowerShell](https://en.wikipedia.org/wiki/PowerShell). You can also do shell scripting on Command Prompt and PowerShell (which is as powerful as bash at least).

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+Author: [Wang Junming](https://github.com/junming403)
+
+## Overview
+
+This chapter assumes that the reader is familiar with general concepts of operating systems. If you are not familiar with basic operating system concepts, I recommend you pick up the basic concepts first. The classic textbook of operating system **Mordern Operating Systems by Tanenbaum** is highly recommended. 
+
+<img src="images/shell-kernel.png" alt="shell-kernel" style="width:400px;"/>
+
+Computer can be divided into 3 abstraction layers, hardware, kernel and applications. Users cannot control the hardware directly, instructions have to be given through kernel as kernel is the one that controls the hardware. 
+Shell is the interface through which we can pass instructions to kernel, and then it could be executed on the hardware. In this chapter, we will look into **Linux command-line based shell(bash)**.
```
Xenonym: `bash` is not the only shell on Linux; see first comment.

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+
+> Linux is the best-known and most-used open source operating system. As an operating system, Linux is software that sits underneath all of the other software on a computer, receiving requests from those programs and relaying these requests to the computer’s hardware.
```
Xenonym: `best-known and most-used`: a reference to back this up? A causal reader would think this is Windows.

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+  head: head.md
+  siteNav: mainNav.md
+  pageNav: 3
+</frontmatter>
+
+{{ navbar | safe }}
+
+<div class="website-content">
+
+# Introduction to Linux bash shell
```
Xenonym: Could I propose that this article be called `Introduction to Bash shell`? Similar reasons as the first comment.

contents/linux/shell.md
Xenonym: I would say that a major reason to use a shell for remote management is _lag_: doing things via the shell is so much faster.

contents/linux/shell.md
Xenonym: * Instead of the abstract example of `a task you need to perform daily`, it would be clearer to give a concrete example, say, daily rotation of log files.

* Why is `task automation` in `code` styling? Emphasis should probably be done with _italics_ or **bolding**.

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+```bash
+#!/bin/bash
+ 
+echo -e "Please input a filename, I will check the filename's type and permission.\n\n"
+
+# waiting for user input.
+read -p "Input a filename : "  filename
+
+# check if file exists.
+test -z $filename && echo "You MUST input a filename." && exit 0
```
Xenonym: The Google Shell Style Guide [recommends `if [[...]]` for tests](https://google.github.io/styleguide/shell.xml?showone=Test,_%5B_and_%5B%5B#Test,_%5B_and_%5B%5B).

contents/linux/shell.md
Xenonym: Why do we care that Bash scripts are Turing complete? I feel that this is a theoretical point that may not be of practical relevance; the above example is good enough to demonstrate its power.

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+./verify < result
+```
+Notice in this case, `calculate` uses the data in `DataBundle` as input, and output the results to the file `result`. Then, `result` is taken as input to `verify`, and the final verification result is printed on screen.
+
+Furthermore, if the code we write is buggy, it throws exception and error messages are directed to error output, then our result file will be empty! To handle it properly, we can also redirect error output to a file `errors`, and verify it is empty(such that no error occurs during execution) before we execute `verify`.
+
+```bash
+#!/bin/bash
+./calculate < DataBundle > result 2> errors
+if [ -s errors]
```
Xenonym: `if [ -s errors ]`?

contents/linux/shell.md
Xenonym: I suggest using a concrete example; perhaps checking the status of long-running process (eg. zipping a large file with `gzip`).

contents/linux/shell.md
```diff
@@ -0,0 +1,180 @@
...
+
+
+# Where to Go from Here?
+
+- [Introduction to Linux](http://www.linfo.org/newbies.html) is a good introduction to Linux and contains many useful resource links.
+- [Bash Reference Manual](https://www.gnu.org/software/bash/manual/html_node/) is a good reference manual of linux bash shell. 
+- [Online man Page](https://www.linux.org/docs/) is a collection of  `man pages`  online.
+- [linux.org](https://www.linux.org/) is the Linux organization home page, you can find all relevant information about linux here. 
+- One big advantage of linux is its open source feature, you can read [Linux Source Code](https://github.com/torvalds/linux) by your own interest.
```
Xenonym: Someone who is just being introduced to Bash scripting may not get much out of Linux kernel code 🤔

Xenonym (commented):
The article does a good job explaining the advantages of using Bash! Some comments related to examples and categorization below.

contents/linux/shell.md
Xenonym: Some useful resources:

- The [Google Shell Style Guide](https://google.github.io/styleguide/shell.xml) is a good baseline for good shell scripting style.
- Gentoo has a good [`bash` shell reference](https://devmanual.gentoo.org/tools-reference/bash/).
- Apple's [Shell Scripting Primer](https://developer.apple.com/library/archive/documentation/OpenSource/Conceptual/ShellScripting/Introduction/Introduction.html) is a in-depth introduction to shell scripting for beginners.

contents/linux/shell.md
junming403: While in my understanding, one can only make good use of bash shell only if he understand how  OS kernel works. Indeed I agree that without learning basic concepts of OS, one can still use OS, but it's always good to understand how tools works before using the tool? I will mention this Book is a not prerequisite to read this article instead.

contents/linux/shell.md
junming403: Actually I do assume some understanding of OS in order to understand this article, because OS is an essentially a must-learn for everyone of CS background.  I have listed points below for why should get started to using shell.
1. `Remote Management` 
2. `Stability`
3. `Freedom`
4. `Simplicity`

contents/linux/shell.md
junming403: > I originally thought that this article would be on learning how to use the linux shell, but it's really just why you should start learning.`

That's because the requirement of book chapter is to focus more on what and why.

contents/linux/shell.md
junming403: I think it's essential, because I want my audience to understand how shell works before start to use it, as such one can really master it.

contents/linux/shell.md
junming403: > I would say that a major reason to use a shell for remote management is _lag_: doing things via the shell is so much faster.

Exactly, like what I have mentioned, "A remote command-line shell usually requires less bandwidth.", I will emphasise this more.

contents/linux/shell.md
junming403: Thanks, that looks better.

junming403:
Updated according to review, thanks for everyone's help. Special thanks to @ChooJeremy for helping me correcting all grammar errors.

damithc:
@ChooJeremy @rrtheonlyone waiting for your approval first.

contents/shells/bash.md
```diff
@@ -14,66 +14,60 @@
...
 
 ## Why use command-line based shell?
 
-A common question many people would ask is why would we learn command-line based shell as it seems to be very hard to use. There are tons of GUI tools out there, they seems to look more user-friendly and powerful. From my perspective, there are four major reasons for starting to use command line.
+A common question many people ask is why should we learn command-line based shell as it seems to be very hard to use. There are tons of Graphical User Interface (GUI) tools out there, that appear to be more user-friendly and powerful. From my perspective, there are four major reasons to use the command line:
 
-- **Remote Management**: Use over remote command-line shell makes it much easier to perform tasks remotely, such as when a developer is connecting from home. A remote command-line shell usually requires less bandwidth.
+- **Remote Management**: Use over remote command-line shell makes it much easier to perform tasks remotely, such as when a developer is connecting from home. A remote command-line shell usually requires less bandwidth, results in less lag such that we can do things faster.
```
ChooJeremy: `Use over remote command-line` seems weird. Consider rewording?

Also: `bandwidth, results in less lag such that` -> `bandwidth, which results in less lag so that`

contents/shells/bash.md
```diff
@@ -165,7 +157,9 @@ DESCRIPTION
...
+In our opinion, there is no need to learning linux command-line based shell in particular, because there are tons of linux commands out there, it is impossible to remember and master them all at one time without actual practices. Instead, when it comes to the time that you have to use it, try to use the terminal instead of GUI based tool. Although it might be hard at first, it will benefit you more later. For example, `git`, We highly recommend get started to use command-line based git instead of visualisation tools such as `Source Tree`.
+
+However, if you really wish to learn the bash systematically, below we have listed some resources you might find useful.
```
ChooJeremy: `wish to learn the bash` -> `wish to learn bash`

`below we have listed some resources`... not sure about this. `below are some resources`?

ChooJeremy (approved):
LGTM, just some last fixes.

junming403:
Updated according to review, ready for review now.

contents/shells/bash.md
```diff
@@ -0,0 +1,176 @@
...
+# Introduction to Linux bash shell 
+
+Author: [Wang Junming](https://github.com/junming403)
+
+Reviewers: [Rahul Rajesh](https://github.com/rrtheonlyone), [Ong Shu Peng](https://github.com/ongspxm), [Jeremy Choo](https://www.github.com/ChooJeremy), [Tan Zhen Yong
+](https://github.com/Xenonym)
+
+## Overview
+
+This chapter assumes that the reader is familiar with general concepts of operating systems. It is recommended to pick up basic operating system concepts first. The classic textbook of operating system **Mordern Operating Systems by Tanenbaum** is highly recommended.
```
rrtheonlyone: I feel this introduction should be modified (however, this is a personal opinion).

If I was reading this with limited knowledge of how the Operating System works, I would not be keen on reading anymore of this article after this introduction. I think the introduction should keep readers engaged (and not turn them away). Furthermore, chances are if a reader is familiar with general concepts of operating systems, he/she is also familiar with the bash shell.

However, your article is technical at some points. So I understand why this introduction is as such.

So it is up to if you want to change it! (I am fine either way)

contents/shells/bash.md
```diff
@@ -0,0 +1,176 @@
...
+Reviewers: [Rahul Rajesh](https://github.com/rrtheonlyone), [Ong Shu Peng](https://github.com/ongspxm), [Jeremy Choo](https://www.github.com/ChooJeremy), [Tan Zhen Yong
+](https://github.com/Xenonym)
+
+## Overview
+
+This chapter assumes that the reader is familiar with general concepts of operating systems. It is recommended to pick up basic operating system concepts first. The classic textbook of operating system **Mordern Operating Systems by Tanenbaum** is highly recommended. 
+
+<img src="https://cdn-images-1.medium.com/max/1200/1*zEv6mAa2wzHnz4a5uWW4gw.png" alt="shell-kernel" style="width:300px;"/>
+
+Computer can be divided into 3 abstraction layers, hardware, kernel and applications. Users cannot control the hardware directly, instructions have to be given through kernel as kernel is the one that controls the hardware.
```
rrtheonlyone: A computer can be divided into 3 abstraction layers, hardware, kernel and applications. Users cannot control the hardware directly; instructions have to be given through kernel as the kernel is the one that controls the hardware.

contents/shells/bash.md
```diff
@@ -0,0 +1,176 @@
...
+](https://github.com/Xenonym)
+
+## Overview
+
+This chapter assumes that the reader is familiar with general concepts of operating systems. It is recommended to pick up basic operating system concepts first. The classic textbook of operating system **Mordern Operating Systems by Tanenbaum** is highly recommended. 
+
+<img src="https://cdn-images-1.medium.com/max/1200/1*zEv6mAa2wzHnz4a5uWW4gw.png" alt="shell-kernel" style="width:300px;"/>
+
+Computer can be divided into 3 abstraction layers, hardware, kernel and applications. Users cannot control the hardware directly, instructions have to be given through kernel as kernel is the one that controls the hardware. 
+Shell is the interface through which we can pass instructions to kernel, and then it could be executed on the hardware. In this chapter, we will look into **Linux bash shell**.
```
rrtheonlyone: The shell is the interface through which we pass instructions to the kernel. These instructions will then be executed through the hardware.

contents/shells/bash.md
```diff
@@ -0,0 +1,176 @@
...
+
+> Linux is the best-known and most-used open source operating system. As an operating system, Linux is software that sits underneath all of the other software on a computer, receiving requests from those programs and relaying these requests to the computer’s hardware. -- <cite>[What is Linux](https://opensource.com/resources/linux)</cite>
+
+[**bash**](https://www.gnu.org/software/bash/manual/html_node/) stands for **Bourne Again SHell**, an enhanced version of the original Unix Shell program. It has become the standard shell of various linux distributions. 
+
+## Why use command-line based shell?
+
+A common question many people ask is why should we learn command-line based shell as it seems to be very hard to use. There are tons of Graphical User Interface (GUI) tools out there, that appear to be more user-friendly and powerful. From my perspective, there are four major reasons to use the command line:
```
rrtheonlyone: A common question many people ask is: "Why should we learn the command-line based shell when there are tons of Graphical User Interface (GUI) tools out there that appear to be more user-friendly and powerful?". Well, here are some key advantages the command-line shell can give you:

rrtheonlyone: Again up to you to change :) Just a suggestion

rrtheonlyone (approved):
The article is looking good 👍 I liked the level of detail that is provided on the various benefits of the bash shell.

I am okay to push it to the next stage of review.

However, there are still a number of grammer/phrasing issues throughout the article. It is difficult to find a good flow if you read it in one sitting :O I am normally quite particular about these fine details so it may be just me!

contents/shells/bash.md
damithc: I have a similar view about this aspect. Readers are likely to balk at being told that they need to know a classic OS textbook before reading this article. It's best if we can remove the requirement, and possibly remove/rewrite any part that necessitates this requirement in the first place.

contents/shells/bash.md
```diff
@@ -0,0 +1,176 @@
...
+Author: [Wang Junming](https://github.com/junming403)
+
+Reviewers: [Rahul Rajesh](https://github.com/rrtheonlyone), [Ong Shu Peng](https://github.com/ongspxm), [Jeremy Choo](https://www.github.com/ChooJeremy), [Tan Zhen Yong
+](https://github.com/Xenonym)
+
+## Overview
+
+This chapter assumes that the reader is familiar with general concepts of operating systems. It is recommended to pick up basic operating system concepts first. The classic textbook of operating system **Mordern Operating Systems by Tanenbaum** is highly recommended. 
+
+<img src="https://cdn-images-1.medium.com/max/1200/1*zEv6mAa2wzHnz4a5uWW4gw.png" alt="shell-kernel" style="width:300px;"/>
```
damithc: Image reused without citing the source.

contents/shells/bash.md
damithc: Agree with the above. Furthermore, does this apply to every computer or a certain category of computers. Try not to over-generalize.

contents/shells/bash.md
damithc: +1

contents/shells/bash.md
```diff
@@ -0,0 +1,176 @@
...
+
+> Linux is the best-known and most-used open source operating system. As an operating system, Linux is software that sits underneath all of the other software on a computer, receiving requests from those programs and relaying these requests to the computer’s hardware. -- <cite>[What is Linux](https://opensource.com/resources/linux)</cite>
+
+[**bash**](https://www.gnu.org/software/bash/manual/html_node/) stands for **Bourne Again SHell**, an enhanced version of the original Unix Shell program. It has become the standard shell of various linux distributions. 
+
+## Why use command-line based shell?
+
+A common question many people ask is why should we learn command-line based shell as it seems to be very hard to use. There are tons of Graphical User Interface (GUI) tools out there, that appear to be more user-friendly and powerful. From my perspective, there are four major reasons to use the command line:
+
+- **Remote Management**: Remote command-line shell makes it much easier to perform tasks in cloud server, such as when a developer is connecting from home. A remote command-line shell usually requires less bandwidth, which results in less lag so that we can do things faster.
```
damithc: A remote command-line ... in a cloud server, ...

damithc: As this is supposed to be a list of reasons, phrase them as reasons. e.g., `Faster to use for remote operations`, `More stable`, `More flexibility for power users`, `More powerful in the hands of an expert user`.

contents/shells/bash.md
```diff
@@ -0,0 +1,176 @@
...
+
+- **Stability**: No matter what linux distribution you are running, the command line shell is always available and usage is almost the same. However, GUI tools might be different in different distributions, make it very hard to learn.
+
+- **Freedom**: You will have complete control in the command-line shell. GUIs tends to simplify things, giving the user fewer options. However, the command line always provides the full suite of options, giving you complete control.
+
+- **Simplicity**: While you might find command-line difficult to grasp, eventual usage will familiarize yourself such that the command-line is far easier to use than trying to find options in the drop-down menus and alike of a GUI. 
+
+
+### Benefit: Program with bash script
```
damithc: The current heading doesn't sound like a benefit. `Benefit: the ability to automate tasks`?

damithc (changes requested):
Added some comments for the first half. Let me know when ready for the next iteration.

junming403:
Hi Prof, I have updated and ready for next round of review

contents/shells/bash.md
```diff
@@ -0,0 +1,179 @@
...
+DESCRIPTION
+     The grep utility searches any given input files, selecting lines that
+     match one or more patterns.  By default, a pattern matches an input line
+     if the regular expression (RE) in the pattern matches the input line
+     without its trailing newline.  An empty expression matches every line.
+     Each input line that matches at least one of the patterns is written to
+     the standard output ......
+
+```
+## How to Get Started with linux command-line based shell?
```
damithc: Shortened to `How to Get Started?`

contents/shells/bash.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Overview
+
+In general, a computer can be divided into 3 abstraction layers, hardware, kernel and applications. Users cannot control the hardware directly; instructions have to be given through kernel as the kernel is the one that controls the hardware.
+
+<img src="https://cdn-images-1.medium.com/max/1200/1*zEv6mAa2wzHnz4a5uWW4gw.png" alt="shell-kernel" style="width:300px;"/>
+
+<sub>[What (really) happens when you type ls -l in the shell](https://medium.com/meatandmachines/what-really-happens-when-you-type-ls-l-in-the-shell-a8914950fd73)</sub>
+
+The **shell** is the interface through which we pass instructions to the kernel. These instructions will then be executed through the hardware. In this chapter, we will look into **Linux bash shell**.
```
damithc: `In this chapter` -> `In this article`

contents/shells/bash.md
```diff
@@ -0,0 +1,179 @@
...
+
+> Linux is the best-known and most-used open source operating system. As an operating system, Linux is software that sits underneath all of the other software on a computer, receiving requests from those programs and relaying these requests to the computer’s hardware. -- <cite>[What is Linux](https://opensource.com/resources/linux)</cite>
```
damithc: Probably not needed? Surely the reader should know what Linux is? Our target audience is SE students.

contents/shells/bash.md
damithc: Umm... wouldn't it be simpler to say the shell is a way for users to issue commands to a computer? i.e., without mentioning hardware, abstraction, kernal, etc.

contents/shells/bash.md
```diff
@@ -0,0 +1,179 @@
...
+
+[**bash**](https://www.gnu.org/software/bash/manual/html_node/) stands for **Bourne Again SHell**, an enhanced version of the original Unix Shell program. It has become the standard shell of various linux distributions. 
+
+## Why use command-line based shell?
```
damithc: This sounds like there are other types of shells and you are focusing on the _command-line based_ shell. That's not the case, right? shell is by definition command line based. If that is the case, the heading should be `Why shell? Why not GUI?`.

contents/shells/bash.md
```diff
@@ -0,0 +1,179 @@
...
+
+A common question many people ask is: "Why should we learn the command-line based shell when there are tons of Graphical User Interface (GUI) tools out there that appear to be more user-friendly and powerful?". Well, here are some key advantages the command-line shell can give you:
+
+- **Faster to use for remote operations**: A remote command-line shell makes it much easier to perform tasks in a cloud server, such as when a developer is connecting from home. A remote command-line shell usually requires less bandwidth, which results in less lag so that we can do things faster.
+
+- **More stable**: No matter what linux distribution you are running, the command line shell is always available and usage is almost the same. However, GUI tools might be different in different distributions, make it very hard to learn.
```
damithc: _stable_ is not the right word for these qualities.

contents/shells/bash.md
```diff
@@ -0,0 +1,179 @@
...
+
+> Linux is the best-known and most-used open source operating system. As an operating system, Linux is software that sits underneath all of the other software on a computer, receiving requests from those programs and relaying these requests to the computer’s hardware. -- <cite>[What is Linux](https://opensource.com/resources/linux)</cite>
+
+[**bash**](https://www.gnu.org/software/bash/manual/html_node/) stands for **Bourne Again SHell**, an enhanced version of the original Unix Shell program. It has become the standard shell of various linux distributions. 
+
+## Why use command-line based shell?
+
+A common question many people ask is: "Why should we learn the command-line based shell when there are tons of Graphical User Interface (GUI) tools out there that appear to be more user-friendly and powerful?". Well, here are some key advantages the command-line shell can give you:
+
+- **Faster to use for remote operations**: A remote command-line shell makes it much easier to perform tasks in a cloud server, such as when a developer is connecting from home. A remote command-line shell usually requires less bandwidth, which results in less lag so that we can do things faster.
```
damithc: Isn't it faster for both remote and local operations?

contents/shells/bash.md
```diff
@@ -0,0 +1,179 @@
...
+
+The **shell** is the interface through which we pass instructions to the kernel. These instructions will then be executed through the hardware. In this chapter, we will look into **Linux bash shell**.
+
+> Linux is the best-known and most-used open source operating system. As an operating system, Linux is software that sits underneath all of the other software on a computer, receiving requests from those programs and relaying these requests to the computer’s hardware. -- <cite>[What is Linux](https://opensource.com/resources/linux)</cite>
+
+[**bash**](https://www.gnu.org/software/bash/manual/html_node/) stands for **Bourne Again SHell**, an enhanced version of the original Unix Shell program. It has become the standard shell of various linux distributions. 
+
+## Why use command-line based shell?
+
+A common question many people ask is: "Why should we learn the command-line based shell when there are tons of Graphical User Interface (GUI) tools out there that appear to be more user-friendly and powerful?". Well, here are some key advantages the command-line shell can give you:
```
damithc: Shorter version:
>Why type commands in a shell when we can do the same things using GUI applications? Here are some reasons:

contents/shells/bash.md
```diff
@@ -0,0 +1,179 @@
...
+
+- **Faster to use for remote operations**: A remote command-line shell makes it much easier to perform tasks in a cloud server, such as when a developer is connecting from home. A remote command-line shell usually requires less bandwidth, which results in less lag so that we can do things faster.
+
+- **More stable**: No matter what linux distribution you are running, the command line shell is always available and usage is almost the same. However, GUI tools might be different in different distributions, make it very hard to learn.
+
+- **More flexibility for power users**: You will have complete control in the command-line shell. GUIs tends to simplify things, giving the user fewer options. However, the command line always provides the full suite of options, giving you complete control.
+
+- **More powerful in the hands of an expert user**: While you might find command-line difficult to grasp, eventual usage will familiarize yourself such that the command-line is far easier to use than trying to find options in the drop-down menus and alike of a GUI.
```
damithc: These two points seem to overlap.

contents/shells/bash.md
```diff
@@ -0,0 +1,179 @@
...
+
+- **Faster to use for remote operations**: A remote command-line shell makes it much easier to perform tasks in a cloud server, such as when a developer is connecting from home. A remote command-line shell usually requires less bandwidth, which results in less lag so that we can do things faster.
+
+- **More stable**: No matter what linux distribution you are running, the command line shell is always available and usage is almost the same. However, GUI tools might be different in different distributions, make it very hard to learn.
+
+- **More flexibility for power users**: You will have complete control in the command-line shell. GUIs tends to simplify things, giving the user fewer options. However, the command line always provides the full suite of options, giving you complete control.
+
+- **More powerful in the hands of an expert user**: While you might find command-line difficult to grasp, eventual usage will familiarize yourself such that the command-line is far easier to use than trying to find options in the drop-down menus and alike of a GUI. 
+
```
damithc, line 43: To repeat the comment I gave last time, the points mentioned above are not benefits? Why only the content below are given as benefits?

contents/shells/bash.md
```diff
@@ -0,0 +1,179 @@
...
+     The grep utility searches any given input files, selecting lines that
+     match one or more patterns.  By default, a pattern matches an input line
+     if the regular expression (RE) in the pattern matches the input line
+     without its trailing newline.  An empty expression matches every line.
+     Each input line that matches at least one of the patterns is written to
+     the standard output ......
+
+```
+## How to Get Started with linux command-line based shell?
+In our opinion, there is no need to learning linux command-line based shell in particular, because there are tons of linux commands out there, it is impossible to remember and master them all at one time without actual practices. Instead, when it comes to the time that you have to use it, try to use the terminal instead of GUI based tool. Although it might be hard at first, it will benefit you more later. For example, `git`, We highly recommend get started to use command-line based git instead of visualisation tools such as `Source Tree`.
```
damithc: Suggested:
>In our opinion, there is no need to learn the shell in one go. Instead, whenever you use a GUI tool to accomplish a task, try to learn how to do the same using the shell. For example, when using `git`, use it via the shell instead of a GUI tool such as `Source Tree`. That way, you can learn the shell incrementally, over time.

contents/shells/bash.md
```diff
@@ -0,0 +1,179 @@
...
+
+However, if you really wish to learn bash systematically, below are some resources you might find useful.
+
+
+# Where to Go from Here?
```
damithc: Make this list more focused about bash shell. Give exactly what the reader needs to know to get started with the bash shell.

contents/shells/bash.md
```diff
@@ -0,0 +1,179 @@
+<frontmatter>
+  title: Introduction to bash shell
+  footer: footer.md
+  head: head.md
+  siteNav: mainNav.md
+  pageNav: 3
+</frontmatter>
+
+{{ navbar | safe }}
```
damithc: Overall, there seems to be a mix up between features and benefits. Consider this structure:

What is the shell?
* It is a CLI for the OS (is that correct?). Here, we focus on Bash shell.
* Bash supports a Turing-complete language for writing scripts.
* Bash supports combining of scripts via stream redirection.

Why use shell?
* Benefit: it is faster
* Benefit: can automate tasks
* Benefit: it is more powerful than a GUI in the hands of an expert

How to get started?
* Learn as you go
* Some beginner resources: ...
* Use man when you need more info

damithc (changes requested):
Some more comments added. Ready for the next round :-p

junming403:
Hi prof, I have updated the article and reorganized its structure. Ready for the next review.

contents/shells/bash.md
```diff
@@ -0,0 +1,169 @@
...
+Reviewers: [Rahul Rajesh](https://github.com/rrtheonlyone), [Ong Shu Peng](https://github.com/ongspxm), [Jeremy Choo](https://www.github.com/ChooJeremy), [Tan Zhen Yong
+](https://github.com/Xenonym)
+
+## What is the shell?
+
+In general, a computer can be divided into 3 abstraction layers, hardware, kernel and applications. Users cannot control the hardware directly; instructions have to be given through kernel as the kernel is the one that controls the hardware.
+
+<img src="https://cdn-images-1.medium.com/max/1200/1*zEv6mAa2wzHnz4a5uWW4gw.png" alt="shell-kernel" style="width:300px;"/>
+
+<sub>[What (really) happens when you type ls -l in the shell](https://medium.com/meatandmachines/what-really-happens-when-you-type-ls-l-in-the-shell-a8914950fd73)</sub>
```
damithc: At this point the reader doesn't know what this `ls -l` you are referring to.

contents/shells/bash.md
```diff
@@ -0,0 +1,169 @@
...
+
+The **shell** is the interface through which we pass instructions to the kernel. These instructions will then be executed through the hardware. In this article, we will look into **Linux bash shell**.
```
damithc: Give an example of a shell command and the output.

contents/shells/bash.md
```diff
@@ -0,0 +1,169 @@
...
+
+[**bash**](https://www.gnu.org/software/bash/manual/html_node/) stands for **Bourne Again SHell**, an enhanced version of the original Unix Shell program. It has become the standard shell of various linux distributions. 
+
+### Bash script
```
damithc: Script -> Scripts

contents/shells/bash.md
```diff
@@ -0,0 +1,169 @@
...
+
+Bash supports a [**Turning complete**](https://en.wikipedia.org/wiki/Turing_completeness) language for writing scripts, which we usually refers to as **bash script**.
```
damithc: as **bash scripts**.

damithc: Change in other places too.

contents/shells/bash.md
```diff
@@ -0,0 +1,169 @@
...
+
+Here's one example of bash script, suppose you have to check if some given file exist periodically. The following script will ask for your input and output the results. `filename` is a variable used to store user input.
```
damithc: > `filename` is a variable used to store user input.

 it's not like the reader is going to understand the script anyway, given that the target reader is new to bash.

contents/shells/bash.md
```diff
@@ -0,0 +1,169 @@
...
+# waiting for user input.
+read -p "Input a filename : " filename
+
+# check if file exists.
+test -z $filename && echo "You MUST input a filename." && exit 0
+test ! -e $filename && echo "The filename '$filename' DO NOT EXIST" && exit 0
+
+# output result.
+echo "The filename: $filename is a $filetype"
+```
```
damithc, line 65: Give an example output?

damithc: You can also consider using something like asciinema

contents/shells/bash.md
```diff
@@ -0,0 +1,169 @@
...
+
+The above bash script is still just a collection of bash commands. but it still can be done by typing those commands line by line. What makes bash script really exciting is the semantics of `conditionals`, `loops` and `functions`.
```
damithc: The above bash script is still just a collection of bash commands, which means you can get the same results by typing those commands one at a time. What raises bash scripts to a whole new level is the support for `conditionals`, `loops` and `functions`.

contents/shells/bash.md
```diff
@@ -0,0 +1,169 @@
...
+# output permissions for each of the file under the directory.
+filelist=$(ls $dir)
+for filename in $filelist
+do 
+  perm=""
+  test -w "$dir/$filename" && perm="readable"
+  echo "The file $dir/$filename 's permission is $perm" 
+done
+```
+**Functions** in bash script just like functions in normal programming languages, more semantics of bash script can be found [here](https://www.gnu.org/software/bash/manual/html_node/).
```
damithc: **Functions** in bash script are just like functions in normal programming languages. More information about bash scripts can be found [here](https://www.gnu.org/software/bash/manual/html_node/).

contents/shells/bash.md
```diff
@@ -0,0 +1,169 @@
...
+  perm=""
+  test -w "$dir/$filename" && perm="readable"
+  echo "The file $dir/$filename 's permission is $perm" 
+done
+```
+**Functions** in bash script just like functions in normal programming languages, more semantics of bash script can be found [here](https://www.gnu.org/software/bash/manual/html_node/). 
+
+### Combining of scripts via stream redirection
+
+In linux, everything is abstracted as file, including **input**, **output**, and **error** streams. **Redirection** is a feature in Linux such that when executing a command/user program, you can change the standard input/output devices by changing its [file descriptor](https://en.wikipedia.org/wiki/File_descriptor). Each command and user program in linux is an application with input, output and errors. We can redirect the output of one application to the input of another, as such, we combined the two applications together as if we build a pipe between them.
```
damithc: linux -> Linux?

damithc: I think this descriptions has unnecessary details the reader doesn't need to know. This article is not about how Linux works. Can be simplified as follows.

> In Linux, you can _redirect_ the output of one application to the input of another, combining the two applications together as if there is a _pipe_ between them. In fact, Linux can _pipe_ data between programs, files, and input/output devices seamlessly. You can take advantage of these abilities in your shell commands/scripts to perform complex tasks with just a few commands.

contents/shells/bash.md
```diff
@@ -0,0 +1,169 @@
...
+
+<img src="https://www.computerhope.com/jargon/f/file-descriptor-illustration.jpg" alt="file-descriptor" style="width:200px;"/>
+
+<sub>[File descriptor](https://www.computerhope.com/jargon/f/file-descriptor.htm)</sub>
```
damithc: Not needed.

contents/shells/bash.md
```diff
@@ -0,0 +1,169 @@
...
+
+For example, suppose you just finished your coding assignment, the resulting executable `calculate` takes a user input, do the calculation, and gives the output. To run it, you simply types:
```
damithc: >For example, suppose you wrote an executable `calculate` which takes a user input, do a calculation, and gives the output. To run it, you simply type:

contents/shells/bash.md
```diff
@@ -0,0 +1,169 @@
...
+if [ -s errors ]
+then
+     ./verify < result
+else
+     echo "error occured during execution."
+fi
+```
+Another useful command is `|`, which chained linux commands together, as such the output of previous commands is passed as input to the next command.
+
+The following command make use of shell command [ps](http://man7.org/linux/man-pages/man1/ps.1.html) and [grep](http://man7.org/linux/man-pages/man1/grep.1.html). Suppose you want to check the status of process `p` running on your system. Thus you typed in `ps aux`, and all the processes are listed, but there are too many processes that you can't find `p`. As such, you can pass the result of `(ps) aux` to `grep 'p'` to capture process `p`'s status. Thus the command will be look like `ps aux | grep 'p'`.
```
damithc: Explain what `ps aux` and `grep 'p'` does.

contents/shells/bash.md
```diff
@@ -0,0 +1,169 @@
...
+
+Input/output streams can easily be redirected, allowing information to be sent or received from files or other applications(commands and user programs). This can mean test data can be easily supplied or output captured. A more detailed introduction to I/O stream redirection can be found [here](https://www.digitalocean.com/community/tutorials/an-introduction-to-linux-i-o-redirection).
```
damithc: > ~~Input/output streams can easily be redirected, allowing information to be sent or received from files or other applications(commands and user programs). This can mean test data can be easily supplied or output captured.~~ A more detailed introduction to I/O stream redirection can be found [here](https://www.digitalocean.com/community/tutorials/an-introduction-to-linux-i-o-redirection).

contents/shells/bash.md
```diff
@@ -0,0 +1,169 @@
...
+
+- **Less lag in remote connection**: A remote command-line shell usually requires less bandwidth, which results in less lag. It makes it much easier to perform tasks in a cloud server, such as when a developer is connecting from home.
+
+- **Defined Standards**: No matter what linux distribution you are running, the command line shell is always available and usage is almost the same. However, GUI tools might be different in different distributions, make it very hard to learn.
+
+- **More powerful in the hands of an expert user**: You will have complete control of every option in shell. GUIs tends to simplify things, giving the user fewer options. While you might find command-line difficult to grasp, eventual usage will familiarize yourself such that the command-line is far easier to use than trying to find options in the drop-down menus and alike of a GUI. 
+
+- **Task automation**: bash script allows us to automate the frequently performed operations, it is easy to use and can be executed on any Unix-like operating systems without any modifications.
```
damithc: >- **Faster**: Often, performing a task via the shell is faster than doing the same via a GUI application because the shell doesn't get slowed down by other overheads such as GUI rendering. For example, when performing tasks on a remote server using a shell, the lag will be less.
>- **Uniform**: Unlike GUI apps whose availability and usage can vary between Linux distributions, the shell is always available and the usage is almost the same across different Linux distributions. 
>- **More powerful**: GUIs tends to simplify things, giving the user fewer options. With a shell, you will have complete control of every option. While the shell has a steep learning curve, once you are familiar with it, you can do more things more efficiently with a shell compared to GUI applications. For example, you can automate things with a shell, something not easy to do with a GUI application.

contents/shells/bash.md
```diff
@@ -0,0 +1,169 @@
...
+
+## How to get started?
+
+In our opinion, there is no need to learn the shell in one go. Instead, whenever you use a GUI tool to accomplish a task, try to learn how to do the same using the shell. For example, when using git, use it via the shell instead of a GUI tool such as `Source Tree`. That way, you can learn the shell incrementally, over time.
```
damithc: `Source Tree` -> _Source Tree_ (i.e., should not be in code font)

contents/shells/bash.md
```diff
@@ -0,0 +1,169 @@
...
+
+DESCRIPTION
+     The grep utility searches any given input files, selecting lines that
+     match one or more patterns.  By default, a pattern matches an input line
+     if the regular expression (RE) in the pattern matches the input line
+     without its trailing newline.  An empty expression matches every line.
+     Each input line that matches at least one of the patterns is written to
+     the standard output ......
+
+```
```
damithc, line 182: This whole section can be simplified as follows:

> In addition, you can always use the [`man`](https://en.wikipedia.org/wiki/Man_page) command to find more information about a bash command. For example, the command `man grep` will give you the built-in help page about the `grep` command. These help pages are written in a programmer-friendly format and are very comprehensive.

damithc (changes requested):
The structure is better now. 👍 Added some more comments. In general, get rid of info that serve more to illustrate the author's knowledge and not strictly needed by the reader. For example, there is strictly no need to mention that scripting language is turing complete. It is unlikely the reader will think 'so, it can write scripts; but is it turing complete?'. Also, when the article is almost done, get help from a reviewers to fix grammar errors (there are many).

junming403:
Hi prof, I have updated according to review, added some screenshot and link to video demos for the example script. Ready for next review

contents/shells/bash.md
```diff
@@ -0,0 +1,173 @@
...
+<sub>source of image: [What (really) happens when you type ls -l in the shell](https://medium.com/meatandmachines/what-really-happens-when-you-type-ls-l-in-the-shell-a8914950fd73)</sub>
+
+The **shell** is the interface through which we pass instructions to the kernel. These instructions will then be executed through the hardware. In this article, we will look into **Linux bash shell**.
+
+<img src="https://www.howtogeek.com/wp-content/uploads/2017/03/ximg_58c0939c2d487.png.pagespeed.gp+jp+jw+pj+ws+js+rj+rp+rw+ri+cp+md.ic.5lDzHl2vQG.png" alt="shell-kernel" style="width:600px;"/>
+
+<sub>source of image: [The Best Keyboard Shortcuts for Bash](https://www.howtogeek.com/howto/ubuntu/keyboard-shortcuts-for-bash-command-shell-for-ubuntu-debian-suse-redhat-linux-etc/)</sub>
+
+[**bash**](https://www.gnu.org/software/bash/manual/html_node/) stands for **Bourne Again SHell**, an enhanced version of the original Unix Shell program. It has become the standard shell of various linux distributions. 
+
```
damithc, line 43: Before going into explaining bash scripts, explain stand-alone bash commands, with a good example. What is a neat thing you can do with a single bash command, without using any advanced techniques?

contents/shells/bash.md
```diff
@@ -0,0 +1,173 @@
...
+test -z $filename && echo "You MUST input a filename." && exit 0
+test ! -e $filename && echo "The file '$filename' DO NOT EXIST" && exit 0
+test -f $filename && filetype="regular file"
+test -d $filename && filetype="directory"
+
+# output result.
+echo "The file: $filename is a $filetype"
+```
+
+Here is an example output of running the script. You can also find a video demonstration [here](https://asciinema.org/a/gGqI8rPSGTRM7Ndot7B6aLNEF).
```
damithc: Can embed the asciinema in this page?

contents/shells/bash.md
```diff
@@ -0,0 +1,173 @@
...
+
+# check if file exists.
+test -z $filename && echo "You MUST input a filename." && exit 0
+test ! -e $filename && echo "The file '$filename' DO NOT EXIST" && exit 0
+test -f $filename && filetype="regular file"
+test -d $filename && filetype="directory"
+
+# output result.
+echo "The file: $filename is a $filetype"
+```
```
damithc, line 65: The script has a lot of error handling code. You can consider simplifying the script by removing error handling code. The simpler the script is, the more impressed the reader will be. Anyway, I'll leave that up to you.

contents/shells/bash.md
```diff
@@ -0,0 +1,173 @@
...
+Here is an example output of running the script. You can also find a video demonstration [here](https://asciinema.org/a/pNnG9krPxUBmQpJ00nWr8pvxM).
+
+<img src="./check-permission.png" alt="check-permission" style="width:500px;"/>
+
+**Functions** in bash scripts are just like functions in normal programming languages. More information about bash scripts can be found [here](https://www.gnu.org/software/bash/manual/html_node/). 
+
+### Combining of scripts via stream redirection
+
+In Linux, you can redirect the output of one application to the input of another, combining the two applications together as if there is a pipe between them. In fact, Linux can pipe data between programs, files, and input/output devices seamlessly. You can take advantage of these abilities in your shell commands/scripts to perform complex tasks with just a few commands.
```
damithc: In the version I gave, certain words were in italic.

contents/shells/bash.md
```diff
@@ -0,0 +1,173 @@
...
+
+But you are not satisfied with manual testing, instead, you want to test with larger data set file `DataBundle`. And verify the output with another executable `verify`, which takes in the result as input and verify its correctness and output **PASS** or **FAIL**. Here's how we can do it.
+
+```bash
+./calculate < DataBundle > result
+./verify < result
+```
+Notice in this case, `calculate` uses the data in `DataBundle` as input, and output the results to the file `result`. Then, `result` is taken as input to `verify`, and the final verification result is printed on screen.
+
+Furthermore, if the code we write is buggy, it throws exception and error messages are directed to error output, then our result file will be empty! To handle it properly, we can also redirect error output to a file `errors`, and verify it is empty(such that no error occurs during execution) before we execute `verify`.
```
damithc: missing space `empty(such`

contents/shells/bash.md
```diff
@@ -0,0 +1,173 @@
...
+
+- **Faster**: Often, performing a task via the shell is faster than doing the same via a GUI application because the shell doesn't get slowed down by other overheads such as GUI rendering. For example, when performing tasks on a remote server using a shell, the lag will be less.
+- **Uniform**: Unlike GUI apps whose availability and usage can vary between Linux distributions, the shell is always available and the usage is almost the same across different Linux distributions. 
+- **More powerful**: GUIs tends to simplify things, giving the user fewer options. With a shell, you will have complete control of every option. While the shell has a steep learning curve, once you are familiar with it, you can do more things more efficiently with a shell compared to GUI applications. For example, you can automate things with a shell, something not easy to do with a GUI application.
+
+- **Task automation**: bash scripts allows us to automate the frequently performed operations, it is easy to use and can be executed on any Unix-like operating systems without any modifications.
```
damithc: The fourth point is not needed because it is also mentioned at the end of the third point?

contents/shells/bash.md
```diff
@@ -0,0 +1,173 @@
...
+
+DESCRIPTION
+     The grep utility searches any given input files, selecting lines that
+     match one or more patterns.  By default, a pattern matches an input line
+     if the regular expression (RE) in the pattern matches the input line
+     without its trailing newline.  An empty expression matches every line.
+     Each input line that matches at least one of the patterns is written to
+     the standard output ......
+
+```
```
damithc, line 182: If you want to include the man output example, also include a sentence to mention what it is. Alternatively you can omit it because it can be scary to a beginner to see even the help page is hard to understand.

damithc (changes requested):
Nearly ready 👍 . One more round of revisions and then we can ask for some copy-editing help from a peer reviewer.

junming403:
Hi prof, I have updated and ready for review, it seems that asciinema's script is out of date and not supported by chrome, thus I used an image and linked to the video.

contents/shells/bash.md
```diff
@@ -21,17 +21,24 @@ Reviewers: [Rahul Rajesh](https://github.com/rrtheonlyone), [Ong Shu Peng](https
...
 
 <sub>source of image: [The Best Keyboard Shortcuts for Bash](https://www.howtogeek.com/howto/ubuntu/keyboard-shortcuts-for-bash-command-shell-for-ubuntu-debian-suse-redhat-linux-etc/)</sub>
 
-[**bash**](https://www.gnu.org/software/bash/manual/html_node/) stands for **Bourne Again SHell**, an enhanced version of the original Unix Shell program. It has become the standard shell of various linux distributions. 
+Users interact with bash using various commands. There are countless bash commands, majority of them deals with file and directory manipulation, process management and network. Commands are like a list of APIs supported by the operating system, while you can combine them to accomplish something more complicated. 
+
+One example of most widely used command is `ls`, which simply list out all the files under current directory. Here is an video demonstration of running the `ls` command.
+
+<a href="https://asciinema.org/a/06V6C9mg1lIAnJkzrsweojJUP"><img src="https://asciinema.org/a/06V6C9mg1lIAnJkzrsweojJUP.svg" alt="ls" style="width:500px;height:320px;"/></a>
```
damithc: If this is just a link to another page, make it open in a new Window? Alternatively, remove the image and give the link only. Otherwise the reader will click on the play button the play the video and will be annoyed to realize it loads a different page entirely!

contents/shells/bash.md
```diff
@@ -0,0 +1,179 @@
...
+     echo "error occured during execution."
+fi
+```
+Another useful command is `|`, which chained linux commands together, as such the output of previous commands is passed as input to the next command.
+
+The following command make use of shell command [ps](http://man7.org/linux/man-pages/man1/ps.1.html) and [grep](http://man7.org/linux/man-pages/man1/grep.1.html). ps displays information about a selection of the active processes and grep searchs for the pattern in given input. Suppose you want to check the status of process `p` running on your system. Thus you typed in `ps aux`, and all the processes' information are listed, but there are too many processes that you can't find `p`. As such, you can pass the result of `(ps) aux` to `grep 'p'` to capture process `p`'s status. Thus the command will be look like `ps aux | grep 'p'`.
+
+A more detailed introduction to I/O stream redirection can be found [here](https://www.digitalocean.com/community/tutorials/an-introduction-to-linux-i-o-redirection).
+
+## Why use shell?
```
damithc: the shell

damithc (changes requested):
Just a few nits.
@rrtheonlyone can help fix grammar problems? Not urgent.

junming403:
Hi prof, @rrtheonlyone has helped me fixed grammar errors and it is ready for review.

damithc:
Merged after adding a few cosmetic tweaks myself.
Good work @junming403 🎉

damithc:
Thanks to reviewers too

PR #93, opened 2019-03-06
Update Vuejs page

chel-seyy:
Elaborates on features that differentiate VueJS from other frameworks


chel-seyy:
There seems to be an issue with `{{ }}` being used in markdown.

chel-seyy:
@ongspxm can you help me review on this? Thanks!

chel-seyy:
@fzdy1914 can you help me review this?

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -151,18 +153,105 @@ VueJs is a relatively new JavaScript framework as compared to Angular and React.
 2. **Language barriers:**<br/>
 A majority of users of VueJs are the Chinese as VueJs is developed by a Chinese American. He is supportive of the Chinese community and hence a lot of the existing plugins are written in Chinese. There might be some language barriers for an English speaking developer seeking for VueJs resources.
 
-#### Detailed Comparison of VueJs with other JavaScript frameworks can be found from:
+-----
+### Special features of VueJs
+Here are some features that VueJs differ from the other frameworks used:
```
ongspxm: The comparison could have been more obvious, after all, it may seem like different ways of writting it in vue? Maybe adding comment like `//react way` or something?

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -151,18 +153,105 @@ VueJs is a relatively new JavaScript framework as compared to Angular and React.
...
+
+2. **2-way binding**: `v-model` is used to bind the DOM input field to its data variable. This effectively allows your DOM variables to be automatically in sync with your data, regardless of which one is being updated.
+In other words, this reduces the need for you to manually update your data.
+```js
+<input type="checkbox", v-model=isChecked">
+    <label for="checked">Select</label>
+```
+<br>
+
+3. **1-way data flow**: Data can only be passed from parent to child, via `props`.
```
ongspxm: Technically there is this `$emit` feature that does child to parent kind of stuff.

I noticed yiu included it at the bottom, maybe make it more prominent, rather than just a sidenote.

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -151,18 +153,105 @@ VueJs is a relatively new JavaScript framework as compared to Angular and React.
...
+
+```js
+computed: totalCount() {
+    let result = 0
+    this.items.forEach((item) => result += item.count);
+    return result;
+}
+```
+
+Unlike the use of methods, this updating of `totalCount` will only be triggered when the number of `items` in the list or any `item`'s `count` changed.
```
ongspxm: There is a difference between the way you call the computed variables too, like with brackets and without. One will run the actual thing, one will use the cache, which is dependent on the state of the data that is used in the last "calculation". 

Its in the docs somewhere, like forcing the computed to always "compute". Could include for completeness sake

ongspxm:
Sorry, missed this. Saw it when i got the email notifs again

fzdy1914:
> @fzdy1914 can you help me review this?

Sorry, I am actually not experienced with vue.js. So I may not be able to give you help on this.

damithc:
> Sorry, I am actually not experienced with vue.js. So I may not be able to give you help on this.

@fzdy1914 In that case you match the target audience of this article. We recommend articles to be reviewed by those who don't know the topic as well as those who know it.

damithc:
> There seems to be an issue with `{{ }}` being used in markdown.

Yes, you can't use `{{ }}` as they will be processed as nunjucks code by the MarkBind processor. You'll have to try some workarounds. Get MarkBind teams' help if needed.

contents/javascript/Javascript-framework-VueJs.md
chel-seyy: Do you mean watched properties?

ongspxm:
https://vuejs.org/v2/guide/computed.html#Computed-Caching-vs-Methods

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -151,18 +154,141 @@ VueJs is a relatively new JavaScript framework as compared to Angular and React.
 2. **Language barriers:**<br/>
 A majority of users of VueJs are the Chinese as VueJs is developed by a Chinese American. He is supportive of the Chinese community and hence a lot of the existing plugins are written in Chinese. There might be some language barriers for an English speaking developer seeking for VueJs resources.
 
-#### Detailed Comparison of VueJs with other JavaScript frameworks can be found from:
+-----
+### Special features of VueJs
+Here are some features that VueJs differ from the other frameworks used:
+1. **Mutating of data in the DOM**
+
+In Vue, the state of the data can be directly modified. Let's say, there is a message in your app. To change the message, insert this line:
```
emer7: Just some nits, can user `<br>` so that the content of the paragraph follow the tab of the list

```
1. abcd
   asfd
```
instead of
```
1. abcd
asfd
```

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -151,18 +154,141 @@ VueJs is a relatively new JavaScript framework as compared to Angular and React.
...
+```js
+// Vue way
+this.message = 'Hello Space';
+```
+
+In React, such a direct modification of data is not allowed, due to React's need to rerun lifecycle hooks after state is being updated. Data can only be updated using `this.setState` method.
```
emer7: Lifecycle hook may not be understood by all people, consider putting link to it or explain it other way (such as due to React's implementation)

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -151,18 +154,141 @@ VueJs is a relatively new JavaScript framework as compared to Angular and React.
...
+```js
+// React way
+this.setState({ message: 'Hello Space' });
+```
+<br>
+
+2. **2-way binding**: `v-model` is used to bind the DOM input field to its data variable. This effectively allows your DOM variables to be automatically in sync with your data, regardless of which one is being updated.
```
emer7: actually isn't `v-model` in Vue is similar to Angular's?
although yeah 2-way binding is quite a general concept in frontend framework
perhaps can focus on how Vue's is better than Angular's?

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -151,18 +154,141 @@ VueJs is a relatively new JavaScript framework as compared to Angular and React.
...
+In other words, this reduces the need for you to manually update your data.
+```js
+<input type="checkbox", v-model=isChecked">
+    <label for="checked">Select</label>
+```
+<br>
+
+3. **1-way data flow**: Data can only be passed from parent to child, via `props`. Props can be of any data type, including Objects.
```
emer7: Also the props in Vue and React
Perhaps showing how Vue's is better than React (or the difference at least if Vue's not better than React?)

emer7 (commented):
Sorry for the holdup

chel-seyy:
Changed `special features that Vue have` to `main concepts you need in Vue`

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -48,7 +48,7 @@ new Vue ({
 });
 ```
 
-Step-by-step explanation of the code:
+++Step-by-step explanation of the code++
```
damithc: Best not to use underlines as it makes text harder to read, and also, can be confused as a link.

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -78,13 +78,15 @@ In this case, only the `root` component can be accessed in VueJs while the rest
 <b>Step 4:</b> Pass the message to the HTML file using double curly brackets.
 ```HTML
   <div id="root">
-    <h2>{{ message }}</h2>
+    <h2>{ message }</h2>
```
damithc: Is this correct? I thought Vue uses two braces?

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -151,18 +153,147 @@ VueJs is a relatively new JavaScript framework as compared to Angular and React.
 2. **Language barriers:**<br/>
 A majority of users of VueJs are the Chinese as VueJs is developed by a Chinese American. He is supportive of the Chinese community and hence a lot of the existing plugins are written in Chinese. There might be some language barriers for an English speaking developer seeking for VueJs resources.
 
-#### Detailed Comparison of VueJs with other JavaScript frameworks can be found from:
+-----
+### Main concepts needed in VueJs
```
damithc: `Vue Features`?

Also, move this above the `Advantages and Disadvantages of VueJs` section.

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -151,18 +153,147 @@ VueJs is a relatively new JavaScript framework as compared to Angular and React.
...
+Here are some concepts that you need to start writing Vue code:
+1. **Mutating of data in the DOM**<br/>
```
damithc: >Let's say, there is a message in your app.

This sentence is ambiguous. What do you mean by `message`? Is that a variable named `message`?

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -151,18 +153,147 @@ VueJs is a relatively new JavaScript framework as compared to Angular and React.
...
+    ```js
+    // React way
+    this.setState({ message: 'Hello Space' });
+    ```
+
+<br>
+
+2. **2-way binding**: `v-model` is used to bind the DOM input field to its data variable. This effectively allows your DOM variables to be automatically in sync with your data, regardless of which one is being updated.
+In other words, this reduces the need for you to manually update your data.
+    ```js
```
damithc: should be `html`?

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -151,18 +153,147 @@ VueJs is a relatively new JavaScript framework as compared to Angular and React.
...
+    <input type="checkbox", v-model=isChecked">
+        <label for="checked">Select</label>
+    ```
+
+    2-way binding is generally used for input form bindings, where new data is given by users and processed to provide other useful data.
+    It is quite similar to the way Angular is currently doing. However, Angular discourages the use of two-way binding.
```
damithc: Shall we not assume the reader knows Angular and React? Minimize mentions of those two. If you must, you can compare Vue to React and Angular in a separate section.

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -151,18 +153,147 @@ VueJs is a relatively new JavaScript framework as compared to Angular and React.
...
+
+<br>
+
+3. **1-way data flow**: Data can only be passed from parent to child, via `props`. Props can be of any data type, including Objects. On the other hand, if the child wants to pass data to its parent, `$emit` events have to be used instead.<br/>
+In the code below, `to-do list` is the parent and `item` is the child.
+The data in `item` is being passed to `todo-list` for rendering.
```
damithc: I don't think the target reader will be able follow this explanation. Is s/he supposed to know the meaning of parent, child, props, etc.?

damithc (changes requested):
Added a few comments. Only reviewed the first part. 
1. Decide how much knowledge you assume on the part of the reader. It's OK to assume basic HTML, CSS, JS knowledge, but it's not OK to assume the reader knows Vue, React, Angular.
2. State your assumption at the start of the article. You can use a `<box type="info">` for that.
3. Go through your explanations and ensure the target reader can understand the explanation based on the above assumption. Evaluation every keyword against "can I be sure the reader knows the meaning of this?". If the answer is "no", either eliminate the keyword or explain it.

contents/javascript/Javascript-framework-VueJs.md
chel-seyy: Yes, the correct syntax is `{{ message }}`, however Markbind doesn't allow this to be rendered, causing this whole part to disappear altogether.

damithc: Perhaps use `{\{` or something like that and mention that in the actual code it should be replaced with `{{`?

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -13,9 +13,14 @@
...
 
-Author: [Lu Lechuan](https://github.com/LuLechuan)
+**Authors**: [Lu Lechuan](https://github.com/LuLechuan), [Chelsey Ong](https://github.com/chelseyong)<br>
+Reviewers: [Ong Shu Peng](https://github.com/ongspxm), [Gilbert Emerson](https://github.com/emer7)
 
-VueJs (also known as Vue) is an open-source [JavaScript framework](https://en.wikipedia.org/wiki/JavaScript_framework) for building user interfaces. It is designed to improve code quality and maintainablity.
+VueJs (also known as Vue) is an open-source [JavaScript framework](https://en.wikipedia.org/wiki/JavaScript_framework) for building user interfaces. It is designed to improve code quality and maintainability.
+
+<box type="info">
+Reader should have basic knowledge of HTML and JavaScript before proceeding with this article.
```
damithc: -> `This article assumes the reader has a basic knowledge of HTML and JavaScript.`

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -27,18 +32,18 @@ Apart from installing VueJs, you can install [VueJs development tools](https://g
```
damithc: To be consistent with other similar articles, remove the section about installation.

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -48,6 +53,9 @@ new Vue ({
   }
 });
 ```
+<box type="warning">
+Note that `{\{` and `}\}` should not have the slash in your actual code.
```
damithc, line 50: Add a blank line before this line so that MarkBind parses it as Markdown text.
image

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -48,6 +53,9 @@ new Vue ({
...
+</box>
 
 Step-by-step explanation of the code:
```
damithc, line 53: > Import VueJs CDN and the JavaScript file in the main HTML file.

Add a tooltip to `CDN`?

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -79,13 +87,163 @@ In this case, only the `root` component can be accessed in VueJs while the rest
...
 ```HTML
   <div id="root">
-    <h2>{{ message }}</h2>
+    <h2>{\{message}\{</h2>
   </div>
 ```
 
 <b>Step 5:</b> Open the brower and we will see "Hello World" being displayed:
 > <h2>Hello World</h2>
+<br>
```
damithc, line 89: I know this example is from the original version and there is no need for you to fix it. Nevertheless, the reader will wonder "why go into all this trouble to display Hello World?".

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -79,13 +87,163 @@ In this case, only the `root` component can be accessed in VueJs while the rest
...
+
+-----
+### Vue's Features
+Here are some features to get you familiar with Vue:
+1. **Mutating of data in the DOM**<br>
+    In Vue, the state of the data can be directly modified.
+
+    Let's say, there is a variable called `message` in your app. To modify `message`, you can do it as such:
```
damithc: `you can do it as such` -> `you can do the following:`

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -79,13 +87,163 @@ In this case, only the `root` component can be accessed in VueJs while the rest
...
+    this.message = 'Hello Space';
+    ```
+    When `message` is changed, the view will be re-rendered to show the new message. So you can say, DOM is "reacting" to `message`.
+
+    In situations when you want to tie data to the view, do ensure that this data already has an initial value, so that any subsequent updates to this data will enable view to update itself
+    ```js
+    data: {
+        this.message = '';
+    }
+    ```
```
damithc: This last point is not needed because this is not a tutorial.

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -79,13 +87,163 @@ In this case, only the `root` component can be accessed in VueJs while the rest
...
+
+
+<br>
+
+2. **2-way binding**<br>
+    `v-model` is a [Vue directive](https://vuejs.org/v2/api/#v-model) used to bind the DOM input field to its data variable.
+
+    This effectively allows the DOM variables and data to be "in sync", regardless of which one is being updated first.
```
damithc: `data will change, vice versa` -> `data will change, and vice versa`

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -79,13 +87,163 @@ In this case, only the `root` component can be accessed in VueJs while the rest
...
+    ```
+
+    2-way binding is generally used for input form bindings such as checkboxes or drop-downs, where new data is entered by users and then updated in the view.
+
+<br>
+
+3. **1-way data flow**<br>
+    When you have components that are nested within each other, data can only be passed from the outer component to the inner component, via `props`, where `props` are just custom data shared between the components.
+
+    The outer and inner components are also called parent and child, respectively.
```
damithc: Use italics when introducing new terms. e.g.,
>The outer and inner components are also called _parent_ and _child_, respectively.

damithc: In any case, it looks like there is no need to introduce these terms parent and child as they are not needed for the explanation.

contents/javascript/Javascript-framework-VueJs.md
```diff
@@ -79,13 +87,163 @@ In this case, only the `root` component can be accessed in VueJs while the rest
...
+
+    While computed properties are more appropriate to use when you want to update a new state or data,
+    watched properties are generally used to run other functions when this particular data has changed.
+
+    Watched property | Computed property
+    :-------------- | :----------------
+    executed every time page refreshes | uses cached data and executes only when changed
+    allow 1 variable to observed only | allow multiple variables to be observed
+
+    If I switch computed property to watched property from the previous example, it will look like this:
```
damithc: If we switch ...

Change in other places too: I -> We

damithc (changes requested):
Good progress from previous version. Some cosmetic changes suggested. It still feels like there is too much details e.g., 1-way data flow explanation is quite long. Given the target reader doesn't know Vue at all and has only basic HTML/JS knowledge too much details are hard to understand and the reader may not be interested to know them anyway. See if you can shorten the explanation of each feature. The reader needs to know _what_ the feature is and _why_ it helps, but not _how_ to use the feature.

damithc:
Merged and deployed. Good work @chelseyong 
thanks to reviewers too.
PS: I did some cosmetic tweaks before merging

chel-seyy:
Thanks Prof!

PR #98, opened 2019-03-10
Add JUnit 5 to Java section

### LINKED ISSUE #97: Introduction to Testing in Java with JUnit 5
Rationale:
We currently do not have any resources directed at testing with JUnit 5. The SE-EDU Software Engineering for Self-Directed Learners book covers principles of testing (e.g. using stubs, differentiating between unit, integration and system tests), but does not cover how to write unit tests that make use of JUnit 5's powerful API library.
JUnit 5 has a wonderful user guide in place already, but (as I learnt from personal experience) the numerous terminologies would make it difficult for a new learner to Software Engineering to understand and make full use of the API.
Hence, I wish to write a book chapter that leverages on examples (that will be added to AB4 in the near future) and which bridges the theory in SE-EDU book with actual practice. This aims to be a good starting point of reference for future students aiming to write unit tests for a SE project in Java.

Outline:
* What is JUnit 5
* Basic JUnit 5 tests
* Before/After methods and their use cases
* JUnit 5 extension model

sijie123:
Resolves #97 

Reviewers: For your convenience, you can find my chapter at
[https://deploy-preview-98--learningresources.netlify.com/contents/java/junit](https://deploy-preview-98--learningresources.netlify.com/contents/java/junit)


sijie123:
Still a significant work in progress. In particular, I've yet to add screenshots for some examples that I've written.
However, do help me review and give me any early suggestions. Thanks!

contents/java/JUnit5.md
```diff
@@ -0,0 +1,137 @@
...
+{{ navbar | safe }}
+
+<div class="website-content">
+
+# Introduction to Testing in Java with JUnit 5
+
+Author(s): [Lin Si Jie](https://github.com/sijie123)
+
+## What is JUnit 5?
+JUnit is a open-sourced automated testing framework for Java. Using its vast API library, you can easily write tests to ensure correctness in your software engineering project. After setting up JUnit, running tests within popular IDEs like IntelliJ is as easy as clicking a button. JUnit 5 is the latest iteration of the JUnit project, supporting multiple new features in Java (such as Lambdas) and adopting a new Extensions-based model to make writing tests easier, cleaner, and more modular.
```
amad-person: > as easy as clicking a button

Perhaps you can rephrase this, as it sounds like a marketing phrase.

amad-person: > automated testing

You could add a tooltip explaining this term?

contents/java/JUnit5.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Why do automated testing?
+There are many reasons why automated testing is important for any software engineering project. Automated testing
```
amad-person: > Automated testing

I think you should start this on a new line as your bullet points use `Automated testing` as the starting phrase. It will be more convenient to read.

contents/java/JUnit5.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Why use JUnit 5?
+
+### Benefit: JUnit 5 is the most popular testing framework for Java
+According to [Synk](https://snyk.io/blog/jvm-ecosystem-report-2018-tools/), JUnit is the most popular testing framework for software engineers working in Java. Surveying over 10,200 respondents, an impressive 78% of respondents use JUnit as their primary testing framework.
+<pic src="{{baseUrl}}/java/junit5_popularity.png"></pic>
+
+### Benefit: JUnit 5 is tightly integrated with IntelliJ
+JUnit is tightly integrated with IntelliJ, the most popular IDE (Source: [Synk](https://snyk.io/blog/jvm-ecosystem-report-2018-tools/)) that enjoys over 45% market share.
+With IntelliJ, you can:
```
amad-person: As of now, this benefit seems more like a benefit of using `IntelliJ`, rather than `JUnit 5`. Perhaps you can rephrase the argument to make it seem like a benefit of `JUnit 5`?

contents/java/JUnit5.md
```diff
@@ -0,0 +1,137 @@
...
+    String result = "";
+    for (int i = 0; i <= copies; i++) {
+      result += reference + "\n";
+    }
+    return result;
+  }
+}
+```
+
+Let's write a test to check the corectness of our code. By convention, we append `Test` to the name of the tested class.
```
amad-person: > By convention, we append `Test` to the name of the tested class.

This should probably be in an `info` box?

contents/java/JUnit5.md
```diff
@@ -0,0 +1,137 @@
...
+    return result;
+  }
+}
+```
+
+Running the same unit test again, we now have a passing test and we can be <trigger for="modal:somewhatAssured" trigger="click">somewhat assured</trigger> that our repeat function works as intended.
+
+## JUnit APIs
+What else can we test with JUnit? In addition to the `assertEquals` method illustrated above, JUnit allows us to test for many other scenarios. These include
+* assertTrue/assertFalse: generic checking statements for checking validity.
```
amad-person: Using inline code highlighting for `assertTrue`, `assertFalse` etc. would be better as they are `JUnit 5` methods.

amad-person (commented):
Good start! I've left some comments and suggestions 😄

contents/java/JUnit5.md
marvinchin: I was also wondering if you could add a tooltip explaining what this means 🙂 

> Extensions-based model

contents/java/JUnit5.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Benefit: JUnit 5 is tightly integrated with IntelliJ
+JUnit is tightly integrated with IntelliJ, the most popular IDE (Source: [Synk](https://snyk.io/blog/jvm-ecosystem-report-2018-tools/)) that enjoys over 45% market share.
+With IntelliJ, you can:
+* Run tests by simply clicking a button
+* Run a subset of tests, to test specific functionality
+* Analyse <popover effect="fade" content="percentage of code checked by tests" placement="top">code coverage</popover>
+* Compare the difference between expected and actual result
+
+### Benefit: JUnit 5 is an [open source project](https://github.com/junit-team/junit5/)
```
marvinchin: Just wondering how much of a benefit JUnit being OSS is 😅 Would its usefulness be diminished if it was not open source?

contents/java/JUnit5.md
```diff
@@ -0,0 +1,137 @@
...
+  }
+}
+```
+
+That's it! You do not have to write any additional runners or helpers to take advantage of the powerful testing features. The `@Test` tag tells JUnit 5 that the repeatTest method is meant to be a test. JUnit 5 will then automatically execute your test(s) in sequence.
+
+Let's have a look at the test results. Hmm, the test seems to fail when we run it.
+But did you notice the subtle bug in Repeater?
+
+We "accidentally" looped for an extra time! Without the use of unit tests, we would have to manually confirm that the output indeed has 5 lines, an action that is tedious to repeat and prone to human errors.
```
marvinchin: The phrase "looped for an extra time" sounds a little off to me - what do you think of "the loop has an extra iteration"?

contents/java/JUnit5.md
```diff
@@ -0,0 +1,137 @@
...
+
+Now, let's fix the code up.
```
marvinchin: Maybe just "fix the code" 🌵

contents/java/JUnit5.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Why do automated testing?
+There are many reasons why automated testing is important for any software engineering project. Automated testing
+* is the first line of defence against bugs
+* protects against <popover effect="fade" content="changes that unintentionally break existing functionality" placement="top">regression</popover>
+* checks against expected output, preventing errors arising from <popover effect="fade" content="looking manually at output" placement="top">LMAO</popover> or <popover effect="fade" content="looking manually for abnormal output" placement="top">LMFAO</popover>. ([Acronym source](https://stackoverflow.com/questions/10858990/why-use-junit-for-testing))
+* ultimately, giving the developer confidence that their code works correctly.
+
+In fact, testing is so powerful and important that an entire agile development process named Test Driven Development is designed with tests as the centerpiece of software development.
```
marvinchin: You may want to add a tooltip for these as well!

> agile development process

>Test Driven Development

marvinchin (commented):
Looks pretty good! Just a couple of comments and questions 🙂 Looking forward to the rest of the section!

contents/java/JUnit5.md
sijie123: For me personally, they released a new feature just a few weeks ago (temporary directories) that made my job much easier when I converted from JUnit 4 -> JUnit 5 in AB4.

Maybe I should rephrase the benefit of OSS...

sijie123:
Hi @marvinchin @amad-person. Thanks for helping me review my first draft. They were really helpful and I've tried to incorporate as many as I could.
I've completed the chapter, so could I trouble you to help me review my draft again? Thanks!

contents/java/JUnit5.md
```diff
@@ -0,0 +1,243 @@
...
+
+When using JUnit 5 on IntelliJ, you can
+* Run tests by simply clicking a button
+* Run a subset of tests, to test specific functionality
+* Analyse <popover effect="fade" content="percentage of code checked by tests" placement="top">code coverage</popover>
+* Compare the difference between expected and actual result
+
+### Benefit: JUnit 5 is an [open source project](https://github.com/junit-team/junit5/)
+Being an open source project, JUnit 5 actively accepts feedback and contributions from the community. New features are released on cycles averaging every 3-5 weeks. The fast development cycle means that JUnit can constantly keep up with new Java features.
+For example, the TempDir feature was introduced in JUnit 5.4.0 after receiving overwhelming feedback on its usefulness.
```
amad-person: Tooltip for TempDir to explain what it is?

contents/java/JUnit5.md
```diff
@@ -0,0 +1,243 @@
...
+```java
+public class DatabaseExtension {
+  public static Database connectTo(String url) {
+    Database db = connectToDatabase(url);
+    return db;
+  }
+}
+```
+
+Now, any code requiring a database can use simply do
```
amad-person: Now, any code requiring a database can ~use~ simply do

contents/java/JUnit5.md
```diff
@@ -0,0 +1,243 @@
...
+
+## Why do automated testing?
+There are many reasons why automated testing is important for any software engineering project.
+
+<trigger for="modal:automatedTesting2103" trigger="click">Automated testing</trigger>
+* is the first line of defence against bugs
+* protects against <popover effect="fade" content="changes that unintentionally break existing functionality" placement="top">regression</popover>
+* checks against expected output, preventing errors arising from <popover effect="fade" content="looking manually at output" placement="top">LMAO</popover> or <popover effect="fade" content="looking manually for abnormal output" placement="top">LMFAO</popover>. ([Acronym source](https://stackoverflow.com/questions/10858990/why-use-junit-for-testing))
+* ultimately, giving the developer confidence that their code works correctly.
```
amad-person: Automated testing...
ultimately, ~giving~ gives the developer confidence that their code works correctly.

(to be consistent with how the rest of the points are written)

contents/java/JUnit5.md
```diff
@@ -0,0 +1,243 @@
...
+
+In fact, testing is so powerful and important that an entire <trigger for="modal:agileDev" trigger="click">agile development process</trigger> named <trigger for="modal:tdd" trigger="click">Test Driven Development</trigger> is designed with tests as the centerpiece of software development.
+
+## Why use JUnit 5?
+
+### Benefit: JUnit 5 is the most popular testing framework for Java
+According to [Synk](https://snyk.io/blog/jvm-ecosystem-report-2018-tools/), JUnit is the most popular testing framework for software engineers working in Java. Surveying over 10,200 respondents, an impressive 78% of respondents use JUnit as their primary testing framework.
```
amad-person: Synk -> Synk's 2018 report

This is your call though, I'm not sure if you really need to include the year. The rationale was that this article will be read in future years too (the survey results might change at that time).

contents/java/JUnit5.md
```diff
@@ -0,0 +1,243 @@
...
+<pic src="{{baseUrl}}/contents/java/junit5_popularity.png"></pic>
+
+### Benefit: JUnit 5 is tightly integrated with IntelliJ
+JUnit is tightly integrated with IntelliJ, the most popular IDE (Source: [Synk](https://snyk.io/blog/jvm-ecosystem-report-2018-tools/)) for Java.
```
amad-person: Synk -> Synk's 2018 report

Same as comment above.

contents/java/JUnit5.md
```diff
@@ -0,0 +1,243 @@
...
+      result += reference + "\n";
+    }
+    return result;
+  }
+}
+```
+
+Let's write a test to check the correctness of our code. 
+
+:bulb: By convention, we append `Test` to the name of the tested class.
```
amad-person: This could be in a `tip` box.

contents/java/JUnit5.md
```diff
@@ -0,0 +1,243 @@
...
+We "accidentally" caused the loop to have an extra iteration! Without the use of unit tests, we would have to manually confirm that the output indeed has 5 lines, an action that is tedious to repeat and prone to human errors.
+
+Now, let's fix the code.
+
+```java
+public class Repeater {
+  // Returns copies of reference string separated by the '\n' character.
+  public static String repeat(String reference, int copies) {
+    String result = "";
+    for (int i = 0; i < copies; i++) {
```
amad-person: Add a comment above the `for` loop to indicate the code change?

amad-person (commented):
Reviewed!

contents/java/JUnit5.md
```diff
@@ -0,0 +1,243 @@
...
+  head: head.md
+  siteNav: mainNav.md
+  pageNav: 3
+</frontmatter>
+
+{{ navbar | safe }}
+
+<div class="website-content">
+
+# Introduction to Testing in Java with JUnit 5
```
Xenonym: Are the benefits only specific to JUnit *5*? Could this article be called **Introduction to Unit Testing in Java with JUnit**?

contents/java/JUnit5.md
```diff
@@ -0,0 +1,243 @@
...
+
+Author: [Lin Si Jie](https://github.com/sijie123)
+
+Reviewers: [Aadyaa Maddi](https://github.com/amad-person), [Marvin Chin](https://github.com/marvinchin)
+
+## What is JUnit 5?
+JUnit is a open-sourced automated testing framework for Java. Using its vast API library, you can easily write and run tests to ensure correctness in your software engineering project. JUnit 5 is the latest iteration of the JUnit project, supporting multiple new features in Java (such as Lambdas) and adopting a new <popover effect="fade" content="think: OOP for testing" placement="top">Extension-based model</popover> to make writing tests easier, cleaner, and more modular.
```
Xenonym: `open-sourced` -> `open-source`?

contents/java/JUnit5.md
```diff
@@ -0,0 +1,243 @@
...
+```java
+assertThrows(ArithmeticException.class, () -> divide(100, 0));
+```
+
+This expression is a lot easier to understand, and as an added bonus, the code is also much shorter to write!
+
+### Before/After hooks
+
+Not every system can be tested so easily. Sometimes, testing is not as straightforward as running a function.
+For example, we might want to store information in an external database. We would have to first connect to the database before we can run any test.
```
Xenonym: In this scenario, shouldn't the database connection be mocked? Test mocking may be outside the scope of this article, but we probably shouldn't accidentally encourage unit tests that rely on external systems. Perhaps this example can be changed to storing data in a file?

contents/java/JUnit5.md
```diff
@@ -0,0 +1,243 @@
...
+  @Test
+  public static void addUserTest() {
+    db.addUser('NewUser');
+    ...
+  }
+}
+```
+
+The `BeforeEach` API can also be leveraged to ensure that we standardise the testing environment before each test.
+For example, we can re-seed a random variable before each test to ensure that we get consistent tests in functions involving randomisation.
```
Xenonym: This implicitly assumes that readers would understand what it means to seed a random number generator. Given the target audience of SE students, this may not be a safe assumption to make. Perhaps it would be safer to change this example eg. resetting a in-app database to an initial state?

contents/java/JUnit5.md
Xenonym: Maybe you could link to this overwhelming feedback, so that the reader can see for themselves?

contents/java/JUnit5.md
```diff
@@ -0,0 +1,243 @@
...
+  }
+}
+```
+
+<img src="{{baseUrl}}/contents/java/assertPass.PNG" />
+Running the same unit test again, we now have a passing test and we can be <trigger for="modal:somewhatAssured" trigger="click">somewhat assured</trigger> that our repeat function works as intended.
+
+### JUnit APIs
+What else can we test with JUnit? In addition to the `assertEquals` method illustrated above, JUnit allows us to test for many other scenarios. These include
+* `assertTrue`/`assertFalse`: generic checking statements for checking validity.
```
Xenonym: What does `generic` mean here, exactly? This could be confused with Java generics. Perhaps `statements to check if a statement returns true or false`?

Xenonym (commented):
Clear explanation of unit testing and JUnit! Some comments for your consideration :)

contents/java/JUnit5.md
sijie123: Yes. This guide actually uses things specific to JUnit 5, such as the Extensions model.

contents/java/JUnit5.md
sijie123: I'm actually a little conflicted on this. I see where you're coming from, although a file will kinda defeat the point of a BeforeAll initialisation.
I think it might be okay for a unit test to test a connection to a service (eg database), since some part of the code has to be responsible for this functionality anyway.

sijie123:
Hi Prof @damithc, can I get a round of review from you?

contents/java/JUnit5.md
```diff
@@ -0,0 +1,259 @@
...
+  pageNav: 3
+</frontmatter>
+
+{{ navbar | safe }}
+
+<div class="website-content">
+
+# Introduction to Testing in Java with JUnit 5
+
+Author: [Lin Si Jie](https://github.com/sijie123)
```
damithc: Make this bold.

contents/java/JUnit5.md
```diff
@@ -0,0 +1,259 @@
...
+
+Reviewers: [Aadyaa Maddi](https://github.com/amad-person), [Marvin Chin](https://github.com/marvinchin)
+
+## What is JUnit 5?
+JUnit is an open-source automated testing framework for Java. Using its vast API library, you can easily write and run tests to ensure correctness in your software engineering project. JUnit 5 is the latest iteration of the JUnit project, supporting multiple new features in Java (such as Lambdas) and adopting a new <popover effect="fade" content="think: OOP for testing" placement="top">Extension-based model</popover> to make writing tests easier, cleaner, and more modular.
```
damithc: Minimize unnecessary claims, such as _vast_, especially when the claim is subjective. The sentence can stand without it too. Same for _easily_.

damithc: >JUnit 5 is the latest iteration ...

This will make the article incorrect as soon as there is a new version. Instead, just declare that this article is written to be compatible with JUnit 5. For similar reasons, the article title can do without `5` too.

contents/java/JUnit5.md
```diff
@@ -0,0 +1,259 @@
...
+
+## Why do automated testing?
+There are many reasons why automated testing is important for any software engineering project.
```
damithc: Modal should be linked to the first appearance of the keyword?.
image

contents/java/JUnit5.md
```diff
@@ -0,0 +1,259 @@
...
+
+<trigger for="modal:automatedTesting2103" trigger="click">Automated testing</trigger>
+* is the first line of defence against bugs
+* protects against <popover effect="fade" content="changes that unintentionally break existing functionality" placement="top">regression</popover>
+* checks against expected output, preventing errors arising from <popover effect="fade" content="looking manually at output" placement="top">LMAO</popover> or <popover effect="fade" content="looking manually for abnormal output" placement="top">LMFAO</popover>. ([Acronym source](https://stackoverflow.com/questions/10858990/why-use-junit-for-testing))
+* ultimately, gives the developer confidence that their code works correctly.
+
+In fact, testing is so powerful and important that an entire <trigger for="modal:agileDev" trigger="click">agile development process</trigger> named <trigger for="modal:tdd" trigger="click">Test Driven Development</trigger> is designed with tests as the centerpiece of software development.
```
damithc: I think we are straying away from the topic. Automated testing can be an article of its own. In here, just focus on JUnit.

contents/java/JUnit5.md
```diff
@@ -0,0 +1,259 @@
...
+
+## Why use JUnit 5?
```
damithc: Let's drop the 5.

contents/java/JUnit5.md
```diff
@@ -0,0 +1,259 @@
...
+
+### Benefit: JUnit 5 is the most popular testing framework for Java
+According to [Synk's 2018 report](https://snyk.io/blog/jvm-ecosystem-report-2018-tools/), JUnit is the most popular testing framework for software engineers working in Java. Surveying over 10,200 respondents, an impressive 78% of respondents use JUnit as their primary testing framework.
+<pic src="{{baseUrl}}/contents/java/junit5_popularity.png"></pic>
```
damithc: hmm... I don't see why this is a benefit per se.

contents/java/JUnit5.md
```diff
@@ -0,0 +1,259 @@
...
+
+When using JUnit 5 on IntelliJ, you can
+* Run tests by simply clicking a button
+* Run a subset of tests, to test specific functionality
+* Analyse <popover effect="fade" content="percentage of code checked by tests" placement="top">code coverage</popover>
+* Compare the difference between expected and actual result
+
+### Benefit: JUnit 5 is an [open source project](https://github.com/junit-team/junit5/)
+Being an open source project, JUnit 5 actively accepts feedback and contributions from the community. New features are released on cycles averaging every 3-5 weeks. The fast development cycle means that JUnit can constantly keep up with new Java features.
+For example, the <popover effect="fade" content="To test filesystems in a isolated environment" placement="top">Temporary Directory</popover> feature was introduced in JUnit 5.4.0 after receiving <a href="https://github.com/junit-team/junit5/issues/1247">feedback</a> on its usefulness.
```
damithc: This too is not that noteworthy. In fact, most dev tools are open source tools these days.

damithc (changes requested):
The content is good. Consider the following structure to keep the article more focused and future proof:

# JUnit

## What is JUnit?
JUnit is an automated unit testing framework for Java programs. {can link some keywords to modals/sites}

Let's understand how JUnit works by going through a simple example.

{give example here}

## JUnit Features
Here are some noteworthy JUnit features:

### Assertions API

### Before/After Hooks

### Extension Model

## Why JUnit?

### Reason 1: Well-Established
* has been around long
* lively community around its OSS project
* Popular among devs

### Reason 2: Well-Integrated into Dev Tools
e.g., Intellij (what else?)

## ~~Closing Remarks~~
Probably not needed, as it doesn't seem to add anything new

 ## Getting Started

Give a learning path with links to resources.

sijie123:
Thanks for the review Prof. The structure you've proposed makes more sense. I've moved a couple of things around.
Maybe I can get @marvinchin / @amad-person to help me take a quick glance first? Thanks!

contents/java/JUnit.md
```diff
@@ -0,0 +1,258 @@
...
+# JUnit
+
+<b>Author: [Lin Si Jie](https://github.com/sijie123)</b>
+
+Reviewers: [Aadyaa Maddi](https://github.com/amad-person), [Marvin Chin](https://github.com/marvinchin)
+
+## What is JUnit?
+_____
+
+JUnit is an open-source <trigger for="modal:automatedTesting2103" trigger="click">automated testing</trigger> framework for Java. Using its <popover effect="fade" content="Application Programming Interfaces">APIs</popover>, developers can write and run tests to check for correctness in their software engineering projects.
```
marvinchin: I feel like a popover might be better UX here compared to a modal, since it is placed closer to the term itself 🙂 Also, making it a click trigger made me a little uncertain about clicking it, because I thought it was a link elsewhere.

contents/java/JUnit.md
```diff
@@ -0,0 +1,258 @@
...
+      result += reference + "\n";
+    }
+    return result;
+  }
+}
+```
+
+Let's write a test to check the correctness of our code. 
+
+<tip-box>:bulb: By convention, we append `Test` to the name of the tested class.</tip-box>
```
marvinchin: Would putting this box below the example make more sense to the reader? It felt a little odd to be reading the explanation before seeing the example.

contents/java/JUnit.md
```diff
@@ -0,0 +1,258 @@
...
+    for (int i = 0; i < copies; i++) { //Strictly less than copies!
+      result += reference + "\n";
+    }
+    return result;
+  }
+}
+```
+
+<img src="{{baseUrl}}/contents/java/assertPass.PNG" />
+Running the same unit test again, we now have a passing test and we can be <trigger for="modal:somewhatAssured" trigger="click">somewhat assured</trigger> that our repeat function works as intended.
```
marvinchin: Same as before, I feel like popover could be more appropriate here

contents/java/JUnit.md
```diff
@@ -0,0 +1,258 @@
...
+We "accidentally" caused the loop to have an extra iteration! Without the use of unit tests, we would have to manually confirm that the output indeed has 5 lines, an action that is tedious to repeat and prone to human errors.
+
+Now, let's fix the code.
+
+```java
+public class Repeater {
+  // Returns copies of reference string separated by the '\n' character.
+  public static String repeat(String reference, int copies) {
+    String result = "";
+    for (int i = 0; i < copies; i++) { //Strictly less than copies!
```
marvinchin: I think you missed a space between the `//` and `Strictly` 🚀

contents/java/JUnit.md
```diff
@@ -0,0 +1,258 @@
...
+
+<p></p> 
+
+## JUnit Features
+_____
+
+In addition to the basic example in the introduction, JUnit also has more powerful features that further simplifies the process of writing complex tests.
+
+### Assertions API
+In addition to the `assertEquals` method illustrated in the introduction, JUnit allows us to check for many other things. These include
```
marvinchin: Should you have a `:` here?

contents/java/JUnit.md
```diff
@@ -0,0 +1,258 @@
...
+* `assertThrows`/`assertDoesNotThrow`: checks that a piece of code throws the correct type of exception. In other words, JUnit even checks that the application can handle errors (or not)!
+* `assertTimeout`: ensures that a piece of code runs within time constraints - we don't want to keep the user waiting!
+
+These powerful expressions means that we can write more expressive tests.
+Instead of
+```java
+try {
+  int result = divide(100, 0); //Divides 100 by 0
+} catch (ArithmeticException ae) {
+  //This is the expected behaviour.
```
marvinchin: There seems to be a missing space between the `//` and the comment content here as well.

contents/java/JUnit.md
```diff
@@ -0,0 +1,258 @@
...
+
+<p></p> 
+
+## Benefits of JUnit
+
+_____
+
+Here are some key reasons why developers should consider using JUnit:
+
+### Benefit 1: JUnit is well-integrated with dev tools
```
marvinchin: Should we use `development` instead of `dev`?

marvinchin (changes requested):
Looks good! Just a couple of formatting and presentation from me 🙂

sijie123:
Thanks @marvinchin.
I think it's ready for a second round of review from Prof.

contents/java/JUnit.md
```diff
@@ -0,0 +1,226 @@
...
+
+```java
+public class RepeaterTest {
+  @Test
+  public void repeatTest() {
+    assertEquals(repeat("a", 5), "a\na\na\na\na\n");
+  }
+}
+```
+<tip-box>:bulb: By convention, we append `Test` to the name of the tested class.</tip-box>
```
damithc: Also say a few words about the intention of the test? i.e., it verifies that when the input is `("a", 5)`, ...

contents/java/JUnit.md
```diff
@@ -0,0 +1,226 @@
...
+  // Returns copies of reference string separated by the '\n' character.
+  public static String repeat(String reference, int copies) {
+    String result = "";
+    for (int i = 0; i <= copies; i++) {
+      result += reference + "\n";
+    }
+    return result;
+  }
+}
+```
```
damithc, line 37: parameter names can be improved e.g., stringToCopy, numberOfCopies

contents/java/JUnit.md
```diff
@@ -0,0 +1,226 @@
...
+Let's write a test to check the correctness of our code. 
+
+```java
+public class RepeaterTest {
+  @Test
+  public void repeatTest() {
+    assertEquals(repeat("a", 5), "a\na\na\na\na\n");
+  }
+}
+```
```
damithc, line 49: Consider improving the example to make it more meaningful. Sometimes, a nonsensical or/and trivial example is harder to understand than a realistic one.

contents/java/JUnit.md
```diff
@@ -0,0 +1,226 @@
...
+<tip-box>:bulb: By convention, we append `Test` to the name of the tested class.</tip-box>
+
+That's it! We do not have to write any additional runners or helpers to take advantage of the powerful testing features. The `@Test` tag tells JUnit that the repeatTest method is meant to be a test. JUnit will then automatically execute our test(s) in sequence.
```
damithc: I think one needs IDE support to execute JUnit tests if there is no runner class?
Also, JUnit runs tests in random order by default. Not sure if they changed that in JUnit 5.

contents/java/JUnit.md
```diff
@@ -0,0 +1,226 @@
...
+
+Here are some key reasons why developers should consider using JUnit:
+
+### Benefit 1: JUnit is well-integrated with development tools
+
+JUnit is supported by many popular IDEs for developing in Java:
+* IntelliJ IDEA
+* Eclipse
+* NetBeans
+* Visual Studio Code
```
damithc: This sounds like an exhaustive list while it is actually not. You can give them as _examples_ instead.

contents/java/JUnit.md
```diff
@@ -0,0 +1,226 @@
...
+<img src="{{baseUrl}}/contents/java/assertPass.PNG" />
+Running the same unit test again, we now have a passing test and we can be <popover effect="fade" content="We've only done one simple test. While this test works, we cannot be assured that all other cases would work too. How will our code behave if a negative number is passed in?">somewhat assured</popover> that our repeat function works as intended.
+
+<p></p> 
+
+## Benefits of JUnit
+
+_____
+
+Here are some key reasons why developers should consider using JUnit:
```
damithc: Are you giving benefits of JUnit over other Java unit testing tools? In that case (or in any case), mention what those alternatives are?

contents/java/JUnit.md
```diff
@@ -0,0 +1,226 @@
...
+    for (int i = 0; i < copies; i++) { // Strictly less than copies!
+      result += reference + "\n";
+    }
+    return result;
+  }
+}
+```
+
+<img src="{{baseUrl}}/contents/java/assertPass.PNG" />
+Running the same unit test again, we now have a passing test and we can be <popover effect="fade" content="We've only done one simple test. While this test works, we cannot be assured that all other cases would work too. How will our code behave if a negative number is passed in?">somewhat assured</popover> that our repeat function works as intended.
```
damithc: It feels like you are also describing benefits of automated testing while the section is supposed to explain JUnit.

contents/java/JUnit.md
```diff
@@ -0,0 +1,226 @@
...
+
+</div>
+
+<modal id="modal:affectEachOther">
+  <div slot="modal-header" class="modal-title text-center">
+    <h4>How will tests affect each other?</h4>
+  </div>
+  <p>If we do not reset a shared resource (e.g. databse / file), a test that uses the resource permanently changes the resource. As a result, other tests cannot assume anything about the initial state of this resource.</p>
+  <p>For example, test A checks for a correctly set-up, empty database. Test B writes to an initially empty database. Without resetting the database, test A can fail if test B manages to runs first.</p>
+</modal>
```
damithc, line 262: missing `</div>` at the end

contents/java/JUnit.md
```diff
@@ -0,0 +1,226 @@
...
+_____
+
+In addition to the basic example in the introduction, JUnit also has more powerful features that further simplifies the process of writing complex tests.
+
+### Assertions API
+In addition to the `assertEquals` method illustrated in the introduction, JUnit allows us to check for many other things. These include:
+* `assertTrue`/`assertFalse`: checks whether statements return true/false as expected.
+* `assertNull`/`assertNotNull`: checks whether something is null or not, without causing a NullPointerException.
+* `assertArrayEquals`: yes, it loops through an entire array for us!
+* `assertThrows`/`assertDoesNotThrow`: checks that a piece of code throws the correct type of exception. In other words, JUnit even checks that the application can handle errors (or not)!
```
damithc: >In other words, JUnit even checks that the application can handle errors (or not)!

More accurately, 'JUnit can verify the actual error handling behavior against the expected error handling behavior'

contents/java/JUnit.md
```diff
@@ -0,0 +1,226 @@
...
+// If code is here, then the ArithmeticException wasn't thrown.
+throw new AssertionError("Dividing 100 by 0 should throw an exception!");
+```
+
+We can now write
+```java
+assertThrows(ArithmeticException.class, () -> divide(100, 0));
+```
+
+This expression is a lot easier to understand, and as an added bonus, the code is also much shorter to write!
```
damithc, line 145: This segment seems to be written as if you are explaining why JUnit5 is better than JUnit4, which does not match the target reader.

contents/java/JUnit.md
```diff
@@ -0,0 +1,226 @@
...
+  return;
+}
+// If code is here, then the ArithmeticException wasn't thrown.
+throw new AssertionError("Dividing 100 by 0 should throw an exception!");
+```
+
+We can now write
+```java
+assertThrows(ArithmeticException.class, () -> divide(100, 0));
+```
```
damithc, line 143: Avoid lambda expressions unless absolutely necessary.

contents/java/JUnit.md
```diff
@@ -0,0 +1,226 @@
...
+
+```java
+class TestInvolvingDatabase {
+  Database db;
+  @BeforeEach
+  public static void resetDatabaseTable() {
+    db.truncate("testTable");
+  }
+}
+```
```
damithc, line 199: There is no mention of `After` hooks. Also, isn't it possible to have code that runs before/after each class/suite etc. i.e., not just tests?

contents/java/JUnit.md
```diff
@@ -0,0 +1,226 @@
...
+
+### Extension Model
+
+Consider a project where multiple functions in different files each need to connect to a database with a different URL.
+One way would be to write a separate `initializeDatabaseConnection()` method for each test that requires a database connection. However, this violates the <popover effect="fade" content="Don't Repeat Yourself" placement="top">DRY</popover> principle.
```
damithc: No need to drag in DRY if you can explain the same thing in a few words.

contents/java/JUnit.md
```diff
@@ -0,0 +1,226 @@
...
+
+If our project is written in a OOP fashion that encourages abstraction and reuse, why shouldn't our tests be so too?
```
damithc: The two things are not necessarily connected. Also, it drags in OOP into the picture without a good reason.

contents/java/JUnit.md
```diff
@@ -0,0 +1,226 @@
...
+Now, any code requiring a database can simply do
+
+```java
+public class DatabaseTest {
+  @RegisterExtension
+  public static Database db = DatabaseExtension.connectTo('localhost');
+}
+```
+
+to initialise a database connection for all tests.
```
damithc: There is no need to break sentences as above. Instead, you can do this:
>Now, any code requiring a database can simply do something like the following to initialise a database connection for all tests.

initialize (standardize on American English)

contents/java/JUnit.md
```diff
@@ -0,0 +1,226 @@
...
+<b>Author: [Lin Si Jie](https://github.com/sijie123)</b>
+
+Reviewers: [Aadyaa Maddi](https://github.com/amad-person), [Marvin Chin](https://github.com/marvinchin)
+
+## What is JUnit?
+_____
+
+JUnit is an open-source <popover effect="fade" content="An automatic test running and verfication process to determine the correctness of a system programmatically">automated testing</popover> framework for Java. Using its <popover effect="fade" content="Application Programming Interfaces">APIs</popover>, developers can write and run tests to check for correctness in their software engineering projects. 
+
+To get a better idea of what JUnit does, let's see how JUnit works with an <popover effect="fade" content="better known system under test or SUT" placement="top">example:</popover>
```
damithc: Probably no need to throw in this additional term `SUT`?

contents/java/JUnit.md
```diff
@@ -0,0 +1,226 @@
...
+This makes it easy for developers to write and run tests.
+
+### Benefit 2: JUnit is well-established
+
+As compared to other testing tools for Java, JUnit
+* has a longer history
+* has a larger community (As of 2019, JUnit has more questions on StackOverflow as compared to other tools)
+* is more popular among developers (according to [Synk's 2018 report](https://snyk.io/blog/jvm-ecosystem-report-2018-tools/))
+
+This means that it is easier to find answers or get help with JUnit.
```
damithc: As these two benefits don't have a lot of content under them, you can present them as two bullet points instead.

contents/java/JUnit.md
```diff
@@ -0,0 +1,226 @@
...
+  }
+}
+```
+
+<img src="{{baseUrl}}/contents/java/assertPass.PNG" />
+Running the same unit test again, we now have a passing test and we can be <popover effect="fade" content="We've only done one simple test. While this test works, we cannot be assured that all other cases would work too. How will our code behave if a negative number is passed in?">somewhat assured</popover> that our repeat function works as intended.
+
+<p></p> 
+
+## Benefits of JUnit
```
damithc: `Why JUnit?`?

damithc (changes requested):
Good progress since previous version. Added some more suggestions.

contents/java/JUnit.md
sijie123: Hmm.. I can't think of a particularly meaningful example that emphasizes the ease of using JUnit to test for correctness.
I thought of using things like 
* name validation (from AB4), but that one involves regex which I ideally don't want to introduce into a simple example, or
* running a typical Calculator (addition/division) example, but those don't show a lot of complexity and need for an automated testing to catch results that are difficult to eyeball

Maybe I'll write a routine that tests for correctly summing up values in an array? That might be a little easier to understand and can show a similar complexity.

contents/java/JUnit.md
sijie123: When checking for thrown exceptions, developers need to pass in an "Executable". It is designed as a Functional Interface, so developers normally use lambda functions for conciseness.
The assertThrows method is one of those things "that lend themselves well to being used with Java 8 lambdas" (from the official user guide), so I think it'd be good to keep the lambda here...

contents/java/JUnit.md
sijie123: I think I wasn't clear here. I mean to say, you can run any code before/after each test method/class. Will fix in the next iteration.

sijie123:
Thanks for your review @damithc. I've made the changes, and it's ready for a round of review.
This time, I've swapped out the example for another one, hopefully it more clearly illustrates how JUnit works. I'm trying to avoid complicated examples that require knowledge of, for example, regex or string processing, so I thought using a simple for-loop might be the clearest/easiest way to demonstrate bugs that can be caught with unit testing.

contents/java/JUnit.md
```diff
@@ -0,0 +1,234 @@
...
+Reviewers: [Aadyaa Maddi](https://github.com/amad-person), [Marvin Chin](https://github.com/marvinchin)
+
+## What is JUnit?
+_____
+
+JUnit is an open-source <popover effect="fade" content="An automatic test running and verfication process to determine the correctness of a system programmatically">automated testing</popover> framework for Java. Using its <popover effect="fade" content="Application Programming Interfaces">APIs</popover>, developers can write and run tests to check for correctness in their software engineering projects. 
+
+To get a better idea of what JUnit does, let's see how JUnit works with a simple example. 
+
+In this example, we need to sum up an array and return its value.
```
damithc: Can be more concise here:

>**JUnit is an _automated testing framework_ for Java** i.e., it can be used to programmatically verify the actual behavior of Java code against the expected behavior.
>
>Let's use a simple example to see how JUnit is used. Suppose we want to test the method `calculateArraySum(int[] inputArray, int N)` which returns the sum of the first `N` values of `inputArray`.

Tweak code as follows?
inputArray -> values
N -> n

contents/java/JUnit.md
```diff
@@ -0,0 +1,234 @@
...
+      sum += inputArray[i];
+    }
+    return sum;
+  }
+}
+```
+
+<tip-box>:bulb: Here, we attempt to use a for-loop to add up the first N values in the inputArray.</tip-box>
+
+Let's write a test to check the correctness of our code. This test will use a sample array of size 5 that sums to 15, and check that our calculateArraySum returns the correct result.
```
damithc: calculateArraySum  -> `calculateArraySum`

contents/java/JUnit.md
```diff
@@ -0,0 +1,234 @@
...
+  int result = divide(100, 0); // Divides 100 by 0
+} catch (ArithmeticException ae) {
+  // This is the expected behaviour.
+  return;
+}
+// If code is here, then the ArithmeticException wasn't thrown.
+throw new AssertionError("Dividing 100 by 0 should throw an exception!");
+```
+
+We can now write
```
damithc: With JUnit, we can replace the above code with just this line:

contents/java/JUnit.md
```diff
@@ -0,0 +1,234 @@
...
+```java
+public class DatabaseExtension {
+  public static Database connectTo(String url) {
+    Database db = connectToDatabase(url);
+    return db;
+  }
+}
+```
+
+Now, any code requiring a database can simply do
```
damithc: Minimize using a code block in the middle of a sentence.

> Now, any code requiring a database can simply do the following to initialize a database connection for all tests:

contents/java/JUnit.md
```diff
@@ -0,0 +1,234 @@
...
+
+```java
+public class DatabaseTest {
+  @RegisterExtension
+  public static Database db = DatabaseExtension.connectTo('localhost');
+}
+```
+
+to initialize a database connection for all tests.
+
```
damithc: Why is this better than writing it as a normal static method and calling it wherever it is needed?

damithc (changes requested):
Good progress. Just a few more suggestions.

sijie123:
Thanks @damithc for your feedback.
Your last comment: "Why is this better than writing it as a normal static method and calling it wherever it is needed?" is actually very helpful. I realise in my previous example, I couldn't justify why users should use an extension (other than because of stylistic/architectural reasons).
I've since updated with a more powerful example showcasing how extensions help to make tests cleaner by abstracting unnecessary details.

Please have a look at the newest iteration. Thank you!

contents/java/JUnit.md
```diff
@@ -0,0 +1,243 @@
...
+      sum += values[i];
+    }
+    return sum;
+  }
+}
+```
+
+<img src="{{baseUrl}}/contents/java/assertPass.PNG" />
+
+Running the same unit test again, we now have a passing test. Our `calculateArraySum` function works as intended.
```
damithc: Briefly mention how we can add more tests to test the method with different parameters?
Also, mention how automated tests can be useful in catching regressions?

contents/java/JUnit.md
```diff
@@ -0,0 +1,243 @@
...
+
+<p></p> 
+
+## Why JUnit?
+
+_____
```
damithc: Let's not add these horizontal lines for level 2 headings. In other articles, only level 1 headings have such horizontal lines (auto-added).

contents/java/JUnit.md
```diff
@@ -0,0 +1,243 @@
...
+
+There are many testing tools for Java, such as TestNG. Among these testing tools, here are some key reasons why developers should consider using JUnit:
```
damithc: >There are many testing tools for Java, such as TestNG.

In reality, TestNG is the only worthy competitor. It came about to fix some shortcomings of JUnit. I don't think it is actively maintained anymore which means JUnit is the clear winner. In that context, the term _many_ in your sentence may not be accurate.

contents/java/JUnit.md
```diff
@@ -0,0 +1,243 @@
...
+```
+
+This expression is a lot easier to understand, and as an added bonus, the code is also much shorter to write!
+
+### Before/After hooks
+
+Not every system can be tested so easily. Sometimes, testing is not as straightforward as running a function.
+For example, we might want to store information in a database. We would have to first connect to the database before we can run any test.
+
+JUnit exposes an API that allows developers to do this process easily and clearly. It offers the `@BeforeAll` and `@BeforeEach` tags that allow us to run any code before each test, allowing us to easily initialize our testing environments before we run any tests:
```
damithc: Two tags are given but the explanation mentions only one behavior.

damithc (changes requested):
Looks quite good now 👍  Just a few more areas to tweak.

damithc:
Merged and deployed.
Good work @sijie123 
Thanks to reviewers too

PR #114, opened 2019-03-28
Update rails chapter

### LINKED ISSUE #95: Introduction to Ruby on Rails
1. What is Ruby on Rails
2. Why should you use it all
3. What situations to use it in (and which ones you shouldn't)
4. Getting started with a simple app.
5. Further Reading

AyushChatto:
Fixes #95 

Progress:

- [x] Intro
- [x] What is Rails
- [x] Why use Rails
- [x] Drawbacks
- [x] Getting Started


AyushChatto:
@luyangkenneth I'm about half-way through, but I'm trying to stagger my work, so I'd love a review now :)

contents/ruby/Rails.md
```diff
@@ -0,0 +1,53 @@
...
+
+> ... I created Rails for me. To make me smile, first and foremost. Its utility was to many degrees subservient to its ability to make me enjoy my life more. To enrich my daily toil of wrangling requirements and requests for web information systems.
+>
+> ... One of the early productivity mottos of Rails went: “You’re not a beautiful and unique snowflake”. It postulated that by giving up vain individuality, you can leapfrog the toils of mundane decisions, and make faster progress in areas that really matter.
+>
+> *David Heinemeier Hansson,* __The Rails Doctrine__
+
+Many popular and very technically mature websites continue to use Rails. These include: GitHub, Airbnb, Twitter, Hulu, Shopify, and Twitch, among others.
+
+## Getting Started
```
luyangkenneth: As mentioned, I think it's better to shift this section to (near) the end.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,53 @@
...
+
+The official website for [Rails](https://rubyonrails.org) will have all the information needed to initialize your first Rails application.
+
+To install Rails on your computer, you will first need to install [Ruby](https://www.ruby-lang.org/en/). After this, you simply need to run:
+
+```rb
+gem install rails
+```
+gem is a package manager for Ruby that allows you to install Rails. Once you're done with the setup, you can start developing applications right away. While the [Rails Guides](https://guides.rubyonrails.org/) are the best possible place to learn everything about Rails, [Go Rails](https://gorails.com/) also has great screencast tutorials that really hold your hand through the multitude of features Rails has, and can be used by more visual learners.
```
luyangkenneth: > gem is a package manager for Ruby that allows you to install Rails

I think there's a greater opportunity here to follow up with how there's a thriving Ruby community and that there are many well developed Ruby gems out there that you can easily plug in to speed up the development of your app. (Maybe mention a few established gems like `devise`)

---

> While the [Rails Guides](https://guides.rubyonrails.org/) are the best possible place to learn everything about Rails, [Go Rails](https://gorails.com/) also has great screencast tutorials...

This part is probably more suited for the `Further Reading` section.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,53 @@
...
+
+We will now explore some of the unique features of Rails that make it such a powerful tool.
+
+### Optimised for Programmer Happiness
```
luyangkenneth: As this is more of a Ruby concept than a Rails concept (although they are of course very intertwined), maybe frame this in terms of how so much of Rails is really about its Ruby foundation.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Convention over Configuration
+
+### Advanced Topics
```
luyangkenneth: I'm curious what you have in mind for this ;)

luyangkenneth (commented):
At this point, I have some comments on the general direction that you seem to be going in, in terms of the voice and framing of this piece.

From the POV of your intended audience (someone new to Rails), I think there could be a much stronger opening, which you can probably achieve with a more direct introduction. I can foresee someone starting with reading the quotes from DHH and thinking: "Okay, the creator of Rails says this, but it feels vague because I have no context about what it means when it comes to using the framework."

On a broad level, I'd like to see this centered around the fact that Rails is an opinionated web framework, and how that has shaped the way it works — i.e. features that I imagine you will discuss under `Convention over Configuration`. It might be useful to bring in MVC concepts when you're talking about (as one example) file/folder conventions, and what each of the M-V-C is responsible for. I think some short code snippets here will be nice and help with the clarity. `How some things are done in Rails` is probably a good starting point for understanding `What is Rails about and why I should care`. Discussing `Rails' doctrines and opinions` in isolation is probably not as good.

Another suggestion I have is to move `How to get started with Rails` to one of the last few sections, after you have piqued your reader's interest and now that they've gotten a better idea of what Rails is. Since you'll be including links to the official Rails docs, I think it's more useful here to preempt the reader about the different elements of the Rails workflow that they might encounter or want to consider using. For example, you could talk about what `gem` is, what `bundler` is, typical friends for your database and hosting needs (Postgres / Heroku) for starting a small project, and useful CLI commands (like `scaffold` / `generate` / etc.) and how they play their part in getting a Rails app up and running.

What do you think?

damithc:
Take a look at Ronak's article on Angular. After a lot of revisions, it is now settling into a reasonable structure. As this is a similar topic, let's stick to a similar structure so as to avoid a lot of rework later. 

Broadly, these are the three sections I suggest.

1. WHAT: explain RoR main features using a suitable example. Explain features as they are; avoid making judgements on why they are good. The reader should not have any room for disagreement. The objective is for the user to understand what RoR is and what makes RoR what it is (i.e., unique features). This section is likely to be the bulk of the article. Keep in mind that the reader doesn't know Ruby/RoR and might only have a basic Web dev skills.

2. WHY: give your opinions (and cite others' opinions to support yours) on why RoR is better than alternatives. Refer to features explained in the previous section where needed. The reader may disagree with these opinions but that's OK. This section need not be very long.

3. HOW to get started: Give a thoughtfully planned learning path for the reader to get started on RoR. It's ok to give some alternative options in the learning path, but don't simply dump a list of links.

contents/ruby/Rails.md
AyushChatto: Gotcha, it does flow better that way :)

contents/ruby/Rails.md
AyushChatto: > how there's a thriving Ruby community and that there are many well developed Ruby gems

Great Idea :D 

> This part is probably more suited for the `Further Reading` section.

I meant that as the starting point for Rails, but I'll cross that bridge when I get there, since I'll move it further down.

contents/ruby/Rails.md
AyushChatto: Actually, the Ruby concept is Principle of Least Surprise - The Principle of the Bigger Smile is primarily a Rails concept :) 
https://rubyonrails.org/doctrine/#optimize-for-programmer-happiness

contents/ruby/Rails.md
AyushChatto: I'm having a tough time deciding which ones to pick. I'll push them out asap. Thanks for the extensive review! :D

AyushChatto:
> What do you think?

@luyangkenneth Thanks a lot for the review. I agree that that is probably the best way to introduce the topic, so I'm gonna push up the edited version asap :)

contents/ruby/Rails.md
luyangkenneth: Aha, I guess this just goes to show how there's not too much value trying to distinguish between strictly Rails vs strictly Ruby 😛

contents/ruby/Rails.md
luyangkenneth: Now that you have a clearer idea of the structure, this section could very well be optional - maybe just mention some things that are good to consider after one has gotten started with Rails.

AyushChatto:
@damithc Prof do you also recommend following the same order for the three? Because I feel it would be PUNCH-ier if we go WHY, WHAT, HOW - since I convince them it is worth reading on before bombarding them with info. Of course, the purpose of the chapter is different from the lightning talk, so I'm not too sure.

damithc:
> @damithc Prof do you also recommend following the same order for the three? Because I feel it would be PUNCH-ier if we go WHY, WHAT, HOW - since I convince them it is worth reading on before bombarding them with info. Of course, the purpose of the chapter is different from the lightning talk, so I'm not too sure.

Yes, a bit different from a talk. It's hard to understand WHY without knowing WHAT, unless your WHY is mostly superficial things. Unlike a talk, the reader is here by choice (we can assume a certain level of motivation already) and they can skip ahead to WHY if they are looking for more motivation. As this needs to fit with other articles on alternatives, write more like an independent and unbiased observer, rather than an RoR enthusiast.

luyangkenneth (changes requested):
Gonna break this review down into smaller pieces.

Here are some high-level things to start off with:
- Remove `.DS_Store`
- Add `Fixes #<your issue number>` to the PR description so that your issue will be closed upon merging this PR
- Add a link to your chapter into site navigation (the side bar)
- Standardize on american spelling
- When quoting words/phrases, use double quotes instead of single quotes (e.g. `"the Rails way"` instead of `'the Rails way'`). On a similar note, avoid explaining things with quotes unless it's truly appropriate for the situation, because it gives the impression that you're trying to wing your explanation, when a clearer and more grounded alternative phrasing probably exists.

In general, the tone of your writing still sounds very subjective (you even admit to this yourself in one of your paragraphs), which doesn't match the intended nature of these book chapters. I'll work through a few places as an example of what less subjective writing might sound like, which I'd like you to adopt for the rest of the article as you continue revising this.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,157 @@
...
+    // Call to an update function in controller 
+  });
+
+  app.delete('/name/delete/:id', (req, res, next) => {
+    // Call to a delete function in controller
+  });
+
+```
+
+Each one of these routes had to be written manually, and specific routes had to be deliberated upon, so as to not be confusing or break any established conventions the application is already following. Also, if you open the source linked at the top, you might notice that the actual code is far more bloated than this abbreviated version, and involves much more logic. Now, to contrast this with the Rails approach, you first need to declare a resource, namely, an `entry` in the address book. So, in your `routes.rb` file, you would write something like
```
luyangkenneth: The source that you linked to requires the user to scroll to the code snippet at the end of the article to look at what you are trying to point out, and it also doesn't provide that much of a value-add because the article contains a lot of unrelated context (in other words, too much effort to filter signal from noise). I think it's entirely possible to illustrate your point here without linking to a source.

luyangkenneth: > _Also, if you open the source linked at the top, you might notice that the actual code is far more bloated than this abbreviated version, and involves much more logic._

Additionally, this feels like a straw man because in a different framework, you could always separate your routing and controller logic into different files. I think it's better to emphasize how the conventions in Rails help you to focus more attention and energy on building your application logic and less on deciding how to structure your application.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,157 @@
...
+               edit_entry GET    /entries/:id/edit(.:format)                                                              entries#edit
+                    entry GET    /entries/:id(.:format)                                                                   entries#show
+                          PATCH  /entries/:id(.:format)                                                                   entries#update
+                          PUT    /entries/:id(.:format)                                                                   entries#update
+                          DELETE /entries/:id(.:format)                                                                   entries#destroy
+``` 
+
+The word 'automagically' should finally start to make sense at this point! Instead of going through the arduous process of writing out the code dump at the top, and deciding the specific names of the routes, as well as the corresponding functions in the controller, Rails has already configured all the URI patterns with function names in the controller for entries. Now all you need to do is focus on writing the function logic, instead of coding out the names of the individual URIs and routes.
```
luyangkenneth: What a less subjective alternative to this paragraph might look like:
> Resource routing in Rails allows you to quickly declare all of the common routes for a given resource in your application. Instead of defining separate URI patterns for each controller action (which, by convention, maps to a specific CRUD operation in your database), a resourceful route declares them in a single line of code.

And then somewhere around here you can include the link to the relevant Rails guide:
https://guides.rubyonrails.org/routing.html#resource-routing-the-rails-default

contents/ruby/Rails.md
```diff
@@ -0,0 +1,157 @@
...
+
+Rails developers can use ordinal representation to access the same element with the command
+
+```ruby
+elements.second
+```
+
+And while this may not reduce the amount one has to type, most Rails enthusiasts agree that it makes the code much more readable, intuitive, and inherently happiness-inducing.
+
+### Great Community of Developers
```
luyangkenneth: What a less subjective headline might look like:
> Large and Active Community of Developers

contents/ruby/Rails.md
```diff
@@ -0,0 +1,157 @@
...
+
+#### Performance Time
+
+While newer releases have combatted this problem to a large degree, many of the old releases of Rails tend to be slower than their counterparts. Rails is also used to make server-side applications that send static HTML files, with minimal Ruby code to modify the appearance. Since the current trend is towards making proper and responsive client-side, front-end applications, Rails is a comparatively 'old school' approach to web developement. However, there are many Gems that allow you to combine your Rails back-end with a modern React, Angular, or Vue front-end, thus enabling you to create responsive static apps with Rails, but it has to be noted that the default for Rails does not accomodate such applications, thus giving it a worse performance time as compared to a client-side application.
```
luyangkenneth: FWIW it's possible to use Rails in API-only mode (i.e. no views, just routes/controllers/models) and deploy it together with a front-end framework that interfaces with your Rails backend.

luyangkenneth:
Actually, now that I'm re-reading [Prof's comments](https://github.com/se-edu/learningresources/pull/114#issuecomment-478330255), I realize that the best guiding principle for improving this article is really to make your "WHAT" the bulk of the article and "explain features as they are / avoid making judgements on why they are good".

Do you feel that you have enough clarity on how you can continue working on this? I'm happy to pair in person if you'd like!

contents/ruby/Rails.md
AyushChatto: Gotcha - do you think I should do away with the code there, or leave it in? Since my point is that you need to type a lot more too - so maybe I can just remove the source, and say that in order to be at the 'same' place, you need to type these many LoC in each framework?

contents/ruby/Rails.md
AyushChatto: I thought it's okay to be subjective, since this part of the 'Why' and Prof's comment said that I can give my opinions here. I don't mind switching to your version, but that is a lot more business-y and factual. What do you say?

contents/ruby/Rails.md
AyushChatto: I'm fine with `Large and Active Community of Developers` :)

contents/ruby/Rails.md
AyushChatto: Whoops TIL haha. I'll have to read up more on that in that case. Thanks!

AyushChatto:
> I realize that the best guiding principle for improving this article is really to make your "WHAT" the bulk of the article and "explain features as they are / avoid making judgements on why they are good".
> 
> Do you feel that you have enough clarity on how you can continue working on this? I'm happy to pair in person if you'd like!

@luyangkenneth Oh wow that would great :D 
I am a bit lost as to what else I could add to this without making it a tutorial for Rails - and I don't want to undersell it either by rewording the Why's into objective What's. I'll schedule a meeting time with you if you're free and hopefully we can hash something good out :D

damithc:
Still too much selling. As this is a part of a collection of resources, we don't want each topic to try to outsell the others. It will be confusing to readers if the authors (who are supposed to be experts) are making competing claims and leaving the novice reader to make up their own mind. Keep subjective opinions such as 'easy to learn', 'opinionated is good' etc. to a small section. 

Here's an example:


WHAT:
* ...
* RoR is designed as an _opinionated_ framework i.e., it believes there is a 'right way' to do things and expects users to follow it. {this is a fact; you can quote from RoR docs to confirm that it is exactly what RoR is meant to be}.
* ...

WHY:
* ...
* The opinionated approach of RoR (explained above) can make it easy to ..., especially when the use case fits RoR's 'right way'. {note how the word _can_ makes the claim easier to swallow even for a skeptic} It is also useful for beginners who don't have their own opinion about the 'right way'. When the use case is not a good fit on the other hand, ...  {even when giving subjective opinions, you can still try to give a balanced view}

contents/ruby/Rails.md
luyangkenneth: I think for the purpose of this section, you should do away with the Express code, and explain Rails as it is, without comparing with other frameworks.

luyangkenneth:
> _I thought it's okay to be subjective, since this part of the 'Why' and Prof's comment said that I can give my opinions here. I don't mind switching to your version, but that is a lot more business-y and factual. What do you say?_

> _I am a bit lost as to what else I could add to this without making it a tutorial for Rails - and I don't want to undersell it either by rewording the Why's into objective What's._

As Prof mentioned, you can of course give subjective opinions, but only in your "Why" section. This should also be kept to a minimum.

This book chapter is _supposed_ to be a tutorial for Rails (but not an extensive one, just an introduction). It sounds like you have something against this approach, but I don't think it's inherently good or bad. It just means that the goal here is not about selling, and that most of the content should be factual.

damithc:
> This book chapter is _supposed_ to be a tutorial for Rails (but not an extensive one, just an introduction).

Not in the sense that we want to explain how to install rails step by step etc. The reason being, there must be many other tutorials aiming to do exactly that. But the article can give simple examples to explain features. Our purpose is to give a glimpse of the light at the end of the tunnel rather than guide the reader through the tunnel.

contents/ruby/Rails.md
AyushChatto: I've tried to make it a bit more palatable in this version. If you still feel it's not helping the chapter and is an unnecessary digression, I'll get rid of it :)

luyangkenneth:
This is a step in the right direction, which is great! That said, I think there's still a lot more ground to cover - let's arrange a time to chat in person and work through it together.

AyushChatto:
@luyangkenneth I think this should be more in-line with what we discussed today. Would love a review :)

luyangkenneth:
I'll do a proper review tomorrow, but in the meantime this looks great overall! The only nits I have are minor punctuation and phrasing-related things.

AyushChatto:
@luyangkenneth @marvinchin just bumping this for a review. Whenever you guys are free :)

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+
+<div class="website-content">
+
+# Introduction to Ruby on Rails
+
+Authors: [Chattoraj Ayush](https://github.com/AyushChatto)
+
+## Ruby on Rails overview
+
+**Ruby on Rails** is a web application framework written in the Ruby programming language. It is an opinionated framework, that was designed with the intention of making programming web applications easier and quicker by reducing the amount of code you write, instead replacing it with what it considers the "right" way of doing things - what they call the "Rails" way, although it does allow you to follow your own conventions (at the cost of speed and the productivity boost it promises).
```
luyangkenneth: `by reducing the amount of code you write.` <- This sentence should end here. Right now it's a run-on sentence.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+
+# Introduction to Ruby on Rails
+
+Authors: [Chattoraj Ayush](https://github.com/AyushChatto)
+
+## Ruby on Rails overview
+
+**Ruby on Rails** is a web application framework written in the Ruby programming language. It is an opinionated framework, that was designed with the intention of making programming web applications easier and quicker by reducing the amount of code you write, instead replacing it with what it considers the "right" way of doing things - what they call the "Rails" way, although it does allow you to follow your own conventions (at the cost of speed and the productivity boost it promises). 
+
+The Rails philosophy includes two major principles, namely:-
```
luyangkenneth: Replace all the instances of `:-` in your article with `:`

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+1. **Don't Repeat Yourself (DRY)**, 
+and
+2. **Convention over Configuration** 
+
+### The Rails Way 
+
+One of the core tenets of Rails is 'Convention Over Configuration'. This means that much of the decisions regarding the architecture of the app, as well some other key decisions, are made for you by Rails, and the default application developed by Rails will follow this convention - thus saving you the effort needed to come up with the code necessary to facilitate it. This is perhaps most evident in Rails' decision to follow a Model-View-Controller (MVC) architecture by default.
```
luyangkenneth: `'Convention Over Configuration'` -> `"Convention over Configuration"`

contents/ruby/Rails.md
luyangkenneth: > instead replacing it with what it considers the "right" way of doing things - what they call the "Rails" way, although it does allow you to follow your own conventions (at the cost of speed and the productivity boost it promises).

This feels a bit messy. Maybe rephrase to a new sentence that explains what being opinionated means? Example: `The belief is that there is a certain approach - often known as "The Rails Way" - that is ideal for building most types of web applications.`

I also think that the discussion on following your own conventions should be saved for later.

contents/ruby/Rails.md
luyangkenneth: Lots of repeated words here, so it can definitely be more concise. I like the way the paragraph ends, so maybe this part can simply be:
`One of the core tenets of Rails is "Convention over Configuration", which means that many decisions regarding the structure of your Rails application are made for you in advance, so that there is less effort needed to come up with your own set of best practices. This is perhaps most evident in Rails' decision to follow a Model-View-Controller (MVC) architecture by default.`

I took out `saving you the effort needed to come up with the code necessary to facilitate it.` because you've already mentioned `reducing the amount of code you write` earlier.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+
+The structure of apps following an MVC framework is as follows:-
+
+<center>
+<img src="mvc.jpeg">
+
+_Figure 1. Model-View-Controller Framework_ <sup>[source](https://medium.com/@matthewmain/rails-request-response-cycle-819e9cd8fa4e)</sup>
+</center>
+
+Let us follow along with the diagram, and trace the control flow as a Rails app responds to a request by the client. This app is a standard "Contacts" app that stores the names of "users" of a software.
```
luyangkenneth: `This app is a standard "Contacts" app that stores the names of "users" of a software.` -> `In this example, the client is making a query for all users in the database.`

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+
+#### Router
+
+When Rails receives the URL, it first looks up the URI in the `routes.rb` file, which defines, by default, all the routes in the application. Each valid URI is mapped to a function located inside a Controller (the 'C' in MVC), that is then invoked to provide a response to the request. 
+
+In order to achieve this in the "Rails" way, it has a convenient abstraction called a "Resource". A resource refers to a RESTful resource that can be acted upon with the HTTP verbs (GET, POST, PUT, and DELETE), and can be Created, have its details Read, Updated, and Destroyed (CRUD). In order to declare a resource, you would mention it inside your `routes.rb` in the following manner:-
```
luyangkenneth: `In order to achieve this in the "Rails" way, it has a convenient abstraction called a "Resource".` -> `Rails exposes a convenient abstraction called a "Resource".`

By this point in the article I think you've mentioned `"The Rails Way"` / `the "Rails" way` enough times, so I don't think there's much more value in repeating it, plus your explanations will flow more clearly without it.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+<img src="mvc.jpeg">
+
+_Figure 1. Model-View-Controller Framework_ <sup>[source](https://medium.com/@matthewmain/rails-request-response-cycle-819e9cd8fa4e)</sup>
+</center>
+
+Let us follow along with the diagram, and trace the control flow as a Rails app responds to a request by the client. This app is a standard "Contacts" app that stores the names of "users" of a software.
+
+#### Router
+
+When Rails receives the URL, it first looks up the URI in the `routes.rb` file, which defines, by default, all the routes in the application. Each valid URI is mapped to a function located inside a Controller (the 'C' in MVC), that is then invoked to provide a response to the request.
```
luyangkenneth: `(the 'C' in MVC)` - remove all instances of this. You've already expanded the acronym earlier in your article, so it's reasonable to assume that your readers can follow along.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+
+When Rails receives the URL, it first looks up the URI in the `routes.rb` file, which defines, by default, all the routes in the application. Each valid URI is mapped to a function located inside a Controller (the 'C' in MVC), that is then invoked to provide a response to the request. 
+
+In order to achieve this in the "Rails" way, it has a convenient abstraction called a "Resource". A resource refers to a RESTful resource that can be acted upon with the HTTP verbs (GET, POST, PUT, and DELETE), and can be Created, have its details Read, Updated, and Destroyed (CRUD). In order to declare a resource, you would mention it inside your `routes.rb` in the following manner:- 
+
+```ruby
+resources :users
+```
+
+Now, you can check all the routes that you have in your application by running `bin/rails routes` in your Rails CLI. This should include:
```
luyangkenneth: I know there are nuances here but I think `bin/rails routes` -> `rails routes` works better in an introductory context?

luyangkenneth: `in your Rails CLI` -> `in the command line` ?

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+                edit_user GET    /users/:id/edit(.:format)                                                                users#edit
+                     user GET    /users/:id(.:format)                                                                     users#show
+                          PATCH  /users/:id(.:format)                                                                     users#update
+                          PUT    /users/:id(.:format)                                                                     users#update
+                          DELETE /users/:id(.:format)                                                                     users#destroy
+```
+
+As can be seen, the common routes that you would need for the resources have already been mapped to URI's. Rails will then look for a controller called `users_controller.rb`, as is convention, and invoke the name of the function mapped to the particular URI and Verb. More information about Rails routes and how they work can be found [here](https://guides.rubyonrails.org/routing.html#resource-routing-the-rails-default).
```
luyangkenneth: `mapped to URI's` -> `mapped to their respective URIs`

luyangkenneth: `Rails will then ..., as is convention, and ...` -> `By convention, Rails will then ...`

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+
+#### Controller
+
+In the standard Rails convention, once inside the application directory, if you navigate to `app/controllers/`, you should be able to see all the controllers in your project. Rails will look for the controller with the `<name>_controller.rb`, where `<name>` would be replaced by the result under the `Controller` heading in the list of routes above, in this case, "users". It will then look for the function name under the `Action` heading mapped to the URI function inside the Controller, and then call it. This is where the bulk of the business logic of your application would be. So if the person makes a GET request to `/users`, then Rails will open `users_controller.rb`, and call its `index` function.
```
luyangkenneth: @marvinchin As someone new to Rails, is this explanation clear enough for you?

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+
+#### Model
+
+The Model (the M in MVC) is very similar to a Class (à la Object-Oriented Programming), and is a useful abstraction for representing and encapsulating ideas. In Rails, it is also referred to as the "Active Record", as it follows the Active Record pattern. In this pattern, the Models are mapped to tables in the database, and you can query for the fields in the database directly via the model in the native language (i.e. Ruby), instead of making queries the database's language (SQL, etc).Models also have generic functions such as `find`, `all`, `create`, `save`, `update`, `delete` included in them by default, so you don't need to implement them or write long queries for each.
```
luyangkenneth: `ideas` -> `objects in your application`

luyangkenneth: `In Rails, it is also referred to as the "Active Record", as it follows the Active Record pattern.` -> `This is also known as the Active Record pattern in Rails.`

luyangkenneth: `and you can query for the fields in the database directly via the model in the native language (i.e. Ruby), instead of making queries the database's language (SQL, etc).` -> `and you can query a model's fields directly in Ruby, instead of writing your own SQL queries.`

This sentence is also missing a space after the period.

luyangkenneth: `functions` -> `methods` (Let's stick to this Rails/Ruby convention)

luyangkenneth: `implement them or write long queries for each.` -> `implement them yourself.`

contents/ruby/Rails.md
luyangkenneth: I kind of know what you mean by the `Controller` and `Action` headings, but I don't see those words "in the list of routes above".

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+
+In our diagram, the Controller calls the `all` method in the `User` model, which is, by convention in a file called `User.rb`, which is stored along with all other models at `app/models/`. The model then acts as an intermediary between the application and the database, and returns the results of the corresponding query. 
+
+#### View
+
+After the Controller has fetched all the data necessary and applied whatever transformations are needed, it passes in the requisite fields to its view. The View (the V in MVC) is what the client sees, and is a collection of .erb (Embedded RuBy) files, which are basically just HTML files with lines of Ruby code embedded in it to modify its appearance and behaviour. When the controller is generated using the Rails CLI, a corresponding folder of views is made inside `app/views/`, which is used to store all the .erb files, and that is where Rails will look for files to serve up to the client at the end of the request-response cycle.
```
luyangkenneth: (Embedded RuBy) -> (**E**mbedded **R**u**B**y) ? 😄

luyangkenneth: .erb -> `.erb` (there are 2 instances of these in this paragraph)

luyangkenneth: `behaviour` -> `behavior`

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Fast iteration speed for product developement 
+
+As mentioned before, Rails is an opinionated framework that does things the "Rails" way. This includes automating trivial tasks by following certain conventions, which can make the development process faster. For example, in a Rails project that is connected to a database, each class is mapped to a table, as described in the Active Record Pattern, and the table name is just a pluralised version of the class's name - the `User` class becomes the `Users` table, the `Person` class becomes the `People` table, etc. This means that you don't need to spend any time deliberating on the name of the tables, nor do you have to juggle multiple concepts while mapping out the database and application, as Rails will automatically connect the two and spare you from deliberating about the exact name of the table. A similar approach is in seen in the usage of resources, since it also automates much of the process.
```
luyangkenneth: `As mentioned before, Rails is an opinionated framework that does things the "Rails" way. This includes automating...` -> `As an opinionated framework, Rails helps to automate...`

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+
+The automation of these mundane tasks could make development a more enjoyable (more on that in a bit), and faster process. All of these optimizations can make developing an application between 30-40% faster on Rails, while also making it easier to learn, as you do not need to understand HTTP verbs to make a basic Rails application (while mastery of the topic will certainly help). Furthermore, programmers experienced in Rails usually find it easier to start on existing Rails projects, since the conventions followed are usually the same, giving you the same directory structure and route naming conventions everywhere. 
+  
+### Optimized for Programmer Happiness
+
+One of the primary tenets of Ruby was to make it the "Least Surprising Language", where the language was designed to perform exactly as the developer thought would be the most "expected" function. Rails built on that, and was designed to be optimized for Programmer Happiness. This meant re-imagining many of coding's most standard conventions, as well as adding significant bloat, all for the purpose of making the code look "pretty", and inherently "nice" to write (as subjective as that sounds).
```
luyangkenneth: I think

`One of the primary tenets of Ruby was to make it the "Least Surprising Language", where the language was designed to feel intuitive and perform exactly as a developer would expect. In Rails, even more syntactic sugar is added on top of its Ruby foundations, so that working with a Rails application would be a joyful experience.`

is a better set up for the examples that you are including later on. One of the gripes (among a few other issues) I have about this paragraph is that `adding significant bloat` feels out of place with what you're trying to convey.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+elements[1]
+```
+
+Rails developers can use ordinal representation to access the same element with the command
+
+```ruby
+elements.second
+```
+
+And while this may not reduce the amount one has to type, most Rails enthusiasts agree that it makes the code much more readable, intuitive, and inherently happiness-inducing.
```
luyangkenneth: `readable, intuitive, and inherently happiness-inducing.` -> `readable and inherently happiness-inducing.`

if you agree with my suggestion to include "intuitive" above.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Large and Active Community of Developers
+
+The Ruby language has a very passionate and extensive community of developers, and there is a vast array of libraries (called "Gems"), that can be used with your Rails projects. These libraries provide a lot of functionalities, and can be used to add flexibility to Rails projects when required. Furthermore, as a result of its popularity, it is very likely that any problems you encounter with Rails will have a solution that you can find on Google or StackOverflow. If not, the Ruby community is actually very famous for being 'nice', so you're likely to get a response to a new problem very quickly too.
```
luyangkenneth: `add flexibility to Rails projects when required.` (split these into 2 separate paragraphs) `Furthermore, as a result of its popularity`

luyangkenneth: `If not, the Ruby community is actually very famous for being 'nice', so you're likely to get a response to a new problem very quickly too.` -> `For any new problems, you are also likely to receive a response very quickly because the Ruby community is known for being welcoming and helpful.`

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+
+#### Performance Time
+
+While newer releases have combatted this problem to a large degree, many of the old releases of Rails have had a reputation for being very bloated and slow to respond. Applications made using NodeJS, or purely client-side applications with a minimal backend, tend to have a much lower delay between an action and a response. While it should be noted that the vast majority of delays in applications occur due to poor optimization, and not the inherent speed of the framework, Rails still remains more bulky and slower than many of its competitors. 
+
+#### Learning Curve
+
+Rails' tendency to do things "automagically" can also be very confusing to new users, who might be more used to configuring routes and directory structures themselves. Furthermore, it becomes much harder to go from using Rails to a different framework, as the lack of conventions in other frameworks means you will have to learn some new topics to be proficient. You also have to learn the Ruby language, and it might be harder considering that most front-end applications are in JavaScript and its variants, and you will have to juggle multiple languages while you develop your application.
```
luyangkenneth: > Furthermore, it becomes much harder to go from using Rails to a different framework, as the lack of conventions in other frameworks means you will have to learn some new topics to be proficient.

I disagree - I actually think that having experience with Rails conventions makes you write better web applications, because they incorporate so many best practices. The learning curve here is about learning the conventions of Rails. As for other frameworks that require you to write things more explicitly, it's easier to reason about the application as a beginner / by reading the code. Whereas in a Rails application, you won't be able to find the glue code that links everything together.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+
+However, Rails might not be best suited for all use cases, and there are certain areas where other frameworks might be more suitable. For instance:-
+
+### Drawbacks
+
+#### Lack of Flexibility
+
+As demonstrated above, Rails follows very strict conventions, that make life much easier for developers - or so Rails thinks. It might be true that there are some developers who find Rails' convention very stifling, and, as a result, prefer to use frameworks that allow them more flexibility. Rails will still allow you to do as you please, but since the effort of undo-ing the conventions makes the process longer and less enjoyable, it might be better to avoid Rails.
```
luyangkenneth: I would prefer this paragraph to be phrased like:

`Although Rails is highly optimized for most types of web applications, there are inevitably some cases where following an MVC architecture isn't suitable. In those cases, it might be better for developers to use frameworks that allow for more flexibility. While there are ways for you to configure Rails to your needs, it might be less painful to avoid those conventions from the beginning.`

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+Rails' tendency to do things "automagically" can also be very confusing to new users, who might be more used to configuring routes and directory structures themselves. Furthermore, it becomes much harder to go from using Rails to a different framework, as the lack of conventions in other frameworks means you will have to learn some new topics to be proficient. You also have to learn the Ruby language, and it might be harder considering that most front-end applications are in JavaScript and its variants, and you will have to juggle multiple languages while you develop your application. 
+
+## Getting Started
+
+To install Rails on your computer, you will first need to install [Ruby](https://www.ruby-lang.org/en/). After this, you simply need to run:
+
+```rb
+gem install rails
+```
+gem is a package manager for Ruby that allows you to install Rails (as well as all the other Gems in Ruby). Once you're done with the setup, you can start developing applications right away. Some resources you might find helpful are:-
```
luyangkenneth: gem is a package manager -> `gem` is a package manager

luyangkenneth: `Some resources you might find helpful are:-` -> `Here are some resources that you might find helpful:`

because your points below are standalone sentences.

luyangkenneth (changes requested):
Sorry for the delay, I had lots of comments in mind and I only just found a pocket of time to type them all out.

contents/ruby/Rails.md
marvinchin: > It is an opinionated framework, that was designed with the intention of making programming web applications easier and quicker by reducing the amount of code you write, instead replacing it with what it considers the "right" way of doing things ...

I feel that the first comma is unnecessary here, and makes the phrasing look a little weird. Perhaps you could omit it. Also, the tense seems to be inconsistent.

It is an opinionated framework that is designed with the intention of making programming web applications easier and quicker by reducing the amount of code you write.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+
+Authors: [Chattoraj Ayush](https://github.com/AyushChatto)
+
+## Ruby on Rails overview
+
+**Ruby on Rails** is a web application framework written in the Ruby programming language. It is an opinionated framework, that was designed with the intention of making programming web applications easier and quicker by reducing the amount of code you write, instead replacing it with what it considers the "right" way of doing things - what they call the "Rails" way, although it does allow you to follow your own conventions (at the cost of speed and the productivity boost it promises). 
+
+The Rails philosophy includes two major principles, namely:-
+1. **Don't Repeat Yourself (DRY)**, 
+and
```
marvinchin: This might be subjective, but I feel like adding the "and" here makes the spacing of the list a little awkard 😅 

Perhaps you could either add it on to the end of the first line, or omit it altogether.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+2. **Convention over Configuration** 
+
+### The Rails Way 
+
+One of the core tenets of Rails is 'Convention Over Configuration'. This means that much of the decisions regarding the architecture of the app, as well some other key decisions, are made for you by Rails, and the default application developed by Rails will follow this convention - thus saving you the effort needed to come up with the code necessary to facilitate it. This is perhaps most evident in Rails' decision to follow a Model-View-Controller (MVC) architecture by default.
+
+The structure of apps following an MVC framework is as follows:-
```
marvinchin: Again, not sure if this is subjective - but I think this could be rephrased:

The structure of an application that follows the MVC framework is as follows:

contents/ruby/Rails.md
marvinchin: Some words feel out of place here:

Let us follow ~~along with~~ the diagram~~,~~ and trace the control flow as a Rails app responds to a request by the client.

contents/ruby/Rails.md
marvinchin: Would it be worth explaining what URI is? Especially since it is used in the same sentence as URL, it might be confusing to someone new to web development. A tooltip or popover might be a good way to present this 🙂

contents/ruby/Rails.md
marvinchin: > A resource refers to a RESTful resource that can be acted upon with the HTTP verbs (GET, POST, PUT, and DELETE), and can be Created, have its details Read, Updated, and Destroyed (CRUD).

You don't use the CRUD acronym elsewhere in this article so it feels a little strange to have it mentioned here in parenthesis, but I think I get why you might want to let the reader know what this common acronym means.

It also feels a little confusing to talk about both resources in both CRUD and HTTP terms. Perhaps you could express it in a manner that separates the DB operations from their HTTP invocations:

A resource can refers to an object upon which you can perform CRUD (create, read, update, and delete) operations. These operations are often invoked via a HTTP request of the appropriate request verb (POST, GET, PUT, and DELETE respectively).

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+Let us follow along with the diagram, and trace the control flow as a Rails app responds to a request by the client. This app is a standard "Contacts" app that stores the names of "users" of a software.
+
+#### Router
+
+When Rails receives the URL, it first looks up the URI in the `routes.rb` file, which defines, by default, all the routes in the application. Each valid URI is mapped to a function located inside a Controller (the 'C' in MVC), that is then invoked to provide a response to the request. 
+
+In order to achieve this in the "Rails" way, it has a convenient abstraction called a "Resource". A resource refers to a RESTful resource that can be acted upon with the HTTP verbs (GET, POST, PUT, and DELETE), and can be Created, have its details Read, Updated, and Destroyed (CRUD). In order to declare a resource, you would mention it inside your `routes.rb` in the following manner:- 
+
+```ruby
+resources :users
```
marvinchin, line 48: Perhaps you might want to add a description to explain that we are adding the `users` resource? As someone that is not familiar with ruby, what is happening here is a little confusing to me 😅

contents/ruby/Rails.md
marvinchin: I think I get the rough idea, but I think it could be clearer if a sample controller file is available to guide the reader along! It should make the explanation simpler as well.

contents/ruby/Rails.md
marvinchin: > When the controller is generated using the Rails CLI, a corresponding folder of views is made inside `app/views/`, which is used to store all the .erb files, and that is where Rails will look for files to serve up to the client at the end of the request-response cycle. 

This sentence feels a little long. Perhaps you could split them into multiple sentences?

contents/ruby/Rails.md
marvinchin: > A similar approach is also used ~~in seen in the usage of~~ to handle resources, ~~since it~~ which also automates much of the process.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+
+The automation of these mundane tasks could make development a more enjoyable (more on that in a bit), and faster process. All of these optimizations can make developing an application between 30-40% faster on Rails, while also making it easier to learn, as you do not need to understand HTTP verbs to make a basic Rails application (while mastery of the topic will certainly help). Furthermore, programmers experienced in Rails usually find it easier to start on existing Rails projects, since the conventions followed are usually the same, giving you the same directory structure and route naming conventions everywhere.
```
marvinchin: > a more enjoyable (more on that in a bit), and faster process

The comma here feels unnecessary

marvinchin: > 30-40% faster on Rails
Would be great if you could provide a source for these figures!

marvinchin: > All of these optimizations can make developing an application between 30-40% faster on Rails, while also making it easier to learn, as you do not need to understand HTTP verbs to make a basic Rails application (while mastery of the topic will certainly help).

There seems to be two separate ideas in this sentence (faster development velocity, and easier to learn). Perhaps you could split them up into separate sentences?

contents/ruby/Rails.md
marvinchin: Happiness-inducing sounds a little subjective and too strong in my opinion 😅

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+
+However, Rails might not be best suited for all use cases, and there are certain areas where other frameworks might be more suitable. For instance:-
+
+### Drawbacks
+
+#### Lack of Flexibility
+
+As demonstrated above, Rails follows very strict conventions, that make life much easier for developers - or so Rails thinks. It might be true that there are some developers who find Rails' convention very stifling, and, as a result, prefer to use frameworks that allow them more flexibility. Rails will still allow you to do as you please, but since the effort of undo-ing the conventions makes the process longer and less enjoyable, it might be better to avoid Rails.
+
+#### Performance Time
```
marvinchin: Performance Time sounds a little strange. Perhaps you could drop "Time" and leave it as "Performance"?

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+
+While newer releases have combatted this problem to a large degree, many of the old releases of Rails have had a reputation for being very bloated and slow to respond. Applications made using NodeJS, or purely client-side applications with a minimal backend, tend to have a much lower delay between an action and a response. While it should be noted that the vast majority of delays in applications occur due to poor optimization, and not the inherent speed of the framework, Rails still remains more bulky and slower than many of its competitors.
```
marvinchin: combatted -> combated

marvinchin: > Applications made using NodeJS, or purely client-side applications with a minimal backend,

The phrasing seems to suggest that NodeJS specifically contrasts with client-side applications. I think you're using NodeJS as an example of some class of applications, but I'm not sure what exactly it is.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+## Getting Started
+
+To install Rails on your computer, you will first need to install [Ruby](https://www.ruby-lang.org/en/). After this, you simply need to run:
+
+```rb
+gem install rails
+```
+gem is a package manager for Ruby that allows you to install Rails (as well as all the other Gems in Ruby). Once you're done with the setup, you can start developing applications right away. Some resources you might find helpful are:-
+
+1. The [Ruby Guides](https://www.rubyguides.com/) is a must read, as you cannot use Rails without knowing the underlying language.
```
marvinchin: > The [Ruby Guides](https://www.rubyguides.com/) is a must read

should it be "are" instead of "is"?

contents/ruby/Rails.md
```diff
@@ -0,0 +1,154 @@
...
+2. The [Rails Guides](https://guides.rubyonrails.org/) are the definitive guide for Rails and you can be assured that it will always be up to date with the latest releases.
```
marvinchin, line 165: > The [Rails Guides](https://guides.rubyonrails.org/) are the definitive guide for Rails and you can be assured that it will always be up to date with the latest releases.

Similar issues with singular vs plural terms here.

marvinchin (changes requested):
I think Kenneth has touched more on the structure and the broader tone of the article, so I mainly focused on the understandability(?) of the article from the perspective of someone with some background in web dev, but not Ruby or Rails. 🙂 

Overall I think the article is fairly understandable and covers the high level ideas that Rails is based on (namely MVC and some of the conventions in Rails).

I've also pointed out some parts where the phrasing felt a little off, but these are kinda subjective so feel free to disregard them if you feel that they aren't relevant!

contents/ruby/Rails.md
luyangkenneth: Hahaha the happiness part is a common Rails rhetoric 😛

contents/ruby/Rails.md
luyangkenneth: Oh yeah that's true, ideally you can include sample files for the model, controller, and the index view - in a way that corresponds with the diagram you inserted.

contents/ruby/Rails.md
luyangkenneth: `readable and easy to work with` might be a better description here, because those are the underlying traits that make Rails happiness-inducing.

contents/ruby/Rails.md
luyangkenneth: In the rendered article, the `and` appears at the end of the first bullet point, which was okay for me. I would prefer omitting it though.

contents/ruby/Rails.md
luyangkenneth: Maybe have a follow up of `This is equivalent to manually declaring the 7 resourceful routes... something something`

contents/ruby/Rails.md
luyangkenneth: Yes I very much agree with this

contents/ruby/Rails.md
luyangkenneth: Yeah I would prefer not to mention the names of other frameworks/backends explicitly. Just focus on how Rails has a lot of bulk and is written in Ruby which would run more slowly because it's an interpreted language.

contents/ruby/Rails.md
AyushChatto: Mm gotcha

contents/ruby/Rails.md
AyushChatto: Interesting idea, I'll do that :) 

Screenshots perhaps?

contents/ruby/Rails.md
AyushChatto: Thinking of replacing it with
` You may also find it challenging to switch to other frameworks, where you may be required to configure routes and write SQL queries yourself, instead of having the framework do it for you.` Thoughts?

contents/ruby/Rails.md
AyushChatto: I am largely okay with that, but I also want to make the point about naming conventions and directory structure, etc. 

How about :
`
Although Rails is highly optimized for most types of web applications, there are inevitably some cases where following an MVC architecture isn't suitable. In those cases, it might be better for developers to use frameworks that allow for more flexibility. You may also wish to have a different directory structure, or give your routes different names, and find Rails' defaults very stifling. While there are ways for you to configure Rails to your needs, it might be less painful to avoid those conventions from the beginning. 
`

contents/ruby/Rails.md
AyushChatto: Gotcha, I'll remove it :)

contents/ruby/Rails.md
AyushChatto: I went with `The structure of an app written in the MVC framework is as follows:` , to avoid using `follows` twice :)

contents/ruby/Rails.md
AyushChatto: I think both are fine, depending on whether you're referring to the collection of guides as a singular object (Like 'that group _is_ there') or whether it's referring to just a bunch of guides, not a collective. I'll standardize it to plural though :)

contents/ruby/Rails.md
```diff
@@ -0,0 +1,170 @@
...
+
+#### Controller
+In the standard Rails convention, once inside the application directory, if you navigate to `app/controllers/`, you should be able to see all the controllers in your project. Rails will look for the controller with the `<name>_controller.rb`, where `<name>` would be replaced by the result under the `Controller` heading in the list of routes above, in this case, "users". It will then look for the function name under the `Action` heading mapped to the URI function inside the Controller, and then call it. This is where the bulk of the business logic of your application would be. So if the person makes a GET request to `/users`, then Rails will open `users_controller.rb`, and call its `index` method (refer to the code below).
+
+<center>
```
luyangkenneth: Looks like a copy-paste error. This is making the code snippet look weird in the website. Only the caption should be centered.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Why Rails
+
+Many popular and very technically mature websites started off using, and continue to use Rails. These include: GitHub, Airbnb, Twitter, Hulu, Shopify, and Twitch, among others. Some of the key features that makes Rails the framework of choice for these users are:
+
+### Fast iteration speed for product developement 
+
+As an opinionated framework, Rails helps to automate trivial tasks by following certain conventions, which can make the development process faster. For example, in a Rails project that is connected to a database, each class is mapped to a table, as described in the Active Record Pattern, and the table name is just a pluralised version of the class's name - the `User` class becomes the `Users` table, the `Person` class becomes the `People` table, etc. This means that you don't need to spend any time deliberating on the name of the tables, nor do you have to juggle multiple concepts while mapping out the database and application, as Rails will automatically connect the two and spare you from deliberating about the exact name of the table. A similar approach is used to handle resources, which also automates much of the process.
```
luyangkenneth: `pluralised` -> `pluralized`

contents/ruby/Rails.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Drawbacks
+
+#### Lack of Flexibility
+
+Although Rails is highly optimized for most types of web applications, there are inevitably some cases where following an MVC architecture isn't suitable. In those cases, it might be better for developers to use frameworks that allow for more flexibility. You may also wish to have a different directory structure, or give your routes different names, and find Rails' defaults very stifling. While there are ways for you to configure Rails to your needs, it might be less painful to avoid those conventions from the beginning.
+
+#### Performance
+
+While newer releases have combated this problem to a large degree, many of the old releases of Rails have had a reputation for being very bloated and slow to respond. Applications in other frameworks, or purely client-side applications with a minimal backend, tend to have a much lower delay between an action and a response. While it should be noted that the vast majority of delays in applications occur due to poor optimization, and not the inherent speed of the framework, Rails still remains more bulky and slower than many of its competitors.
```
luyangkenneth, line 149: This was pointed out by Marvin previously:
`combated` -> `combatted`

contents/ruby/Rails.md
```diff
@@ -0,0 +1,170 @@
...
+
+Furthermore, as a result of its popularity, it is very likely that any problems you encounter with Rails will have a solution that you can find on Google or StackOverflow. For any new problems, you are also likely to receive a response very quickly, because the Ruby community prides itself for being welcoming and helpful. 
+
+However, Rails might not be best suited for all use cases, and there are certain areas where other frameworks might be more suitable. For instance:
+
+### Drawbacks
+
+#### Lack of Flexibility
+
+Although Rails is highly optimized for most types of web applications, there are inevitably some cases where following an MVC architecture isn't suitable. In those cases, it might be better for developers to use frameworks that allow for more flexibility. You may also wish to have a different directory structure, or give your routes different names, and find Rails' defaults very stifling. While there are ways for you to configure Rails to your needs, it might be less painful to avoid those conventions from the beginning.
```
luyangkenneth: `You may also wish to have a different directory structure, or give your routes different names, and find Rails' defaults very stifling.`
->
`You may also find Rails' defaults very stifling if you wish to have a different directory structure, or give your routes different names.`

contents/ruby/Rails.md
```diff
@@ -0,0 +1,170 @@
...
+elements[1]
+```
+
+Rails developers can use ordinal representation to access the same element with the command
+
+```ruby
+elements.second
+```
+
+And while this may not reduce the amount one has to type, most Rails enthusiasts agree that it makes the code much more readable, easy to work with, and, consequently, happiness-inducing.
```
luyangkenneth: Too many commas:
`and, consequently, happiness-inducing.` -> `and therefore happiness-inducing.`

If you want you can stick to `consequently` but I feel like `therefore` makes the sentence less of a mouthful here.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Large and Active Community of Developers
+
+The Ruby language has a very passionate and extensive community of developers, and there is a vast array of libraries (called "Gems"), that can be used with your Rails projects. These libraries provide a lot of functionalities, and can be used to add flexibility to Rails projects when required. 
+
+Furthermore, as a result of its popularity, it is very likely that any problems you encounter with Rails will have a solution that you can find on Google or StackOverflow. For any new problems, you are also likely to receive a response very quickly, because the Ruby community prides itself for being welcoming and helpful. 
+
+However, Rails might not be best suited for all use cases, and there are certain areas where other frameworks might be more suitable. For instance:
```
luyangkenneth: Move this line to below your `Drawbacks` heading and omit the `However` (maybe also omit the `For instance:`). I think it'll be more at home over there.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,170 @@
...
+
+The automation of these mundane tasks could make development a more enjoyable (more on that in a bit) and faster process. All of these optimizations can make developing an application between 30-40% faster on Rails [source](https://www.forbes.com/sites/quora/2018/04/03/is-ruby-a-dying-language/#10eb4d0b6a3d). It also makes Rails relatively easy to learn, as you do not need to understand HTTP verbs to make a basic Rails application (while mastery of the topic will certainly help). Furthermore, programmers experienced in Rails usually find it easier to start on existing Rails projects, since the conventions followed are usually the same, giving you the same directory structure and route naming conventions everywhere.
```
luyangkenneth: You need to wrap your `[source](https://the.link)` with square brackets.

Screenshot 2019-04-17 at 1 43 34 AM

contents/ruby/Rails.md
```diff
@@ -0,0 +1,170 @@
...
+
+In our diagram, the Controller calls the `all` method in the `User` model, which is, by convention in a file called `User.rb`, which is stored along with all other models at `app/models/`. The model then acts as an intermediary between the application and the database, and returns the results of the corresponding query. 
+
+#### View
+
+After the Controller has fetched all the data necessary and applied whatever transformations are needed, it passes in the requisite fields to its view. The View is what the client sees, and is a collection of `.erb` (**E**mbedded **R**u**B**y) files, which are basically just HTML files with lines of Ruby code embedded in it to modify its appearance and behavior. When the controller is generated using the Rails CLI, a corresponding folder of views is made inside `app/views/`. All the `.erb` files are stored here, and Rails will look for the file mapped to the controllers actions to serve up to the client at the end of the request-response cycle.
```
luyangkenneth: My suggestion for replacing the last sentence:

All the `.erb` files are named after the controller action and stored in folders named after the controller. Following the same example above, Rails will serve up the `app/views/users/index.html.erb` view to the client when executing the `users#index` controller action, thus concluding the request-response cycle.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,170 @@
...
+
+#### Controller
+In the standard Rails convention, once inside the application directory, if you navigate to `app/controllers/`, you should be able to see all the controllers in your project. Rails will look for the controller with the `<name>_controller.rb`, where `<name>` would be replaced by the result under the `Controller` heading in the list of routes above, in this case, "users". It will then look for the function name under the `Action` heading mapped to the URI function inside the Controller, and then call it. This is where the bulk of the business logic of your application would be. So if the person makes a GET request to `/users`, then Rails will open `users_controller.rb`, and call its `index` method (refer to the code below).
```
luyangkenneth: "users" -> `users`

luyangkenneth: > This is where the bulk of the business logic of your application would be.

Based on Rails best practices, the bulk of the business logic should not live in the controller. This sentence also feels like a mini sidetrack. I think it'll be better to remove it so there's no break when describing the control flow (`...and then call it.` immediately followed by `So if the person makes a GET request...`)

luyangkenneth: And maybe the last sentence should instead read as:

For example, if someone makes a GET request to `/users`, Rails will then respond by calling the `users#index` controller action (refer to the code below).

"So if the person" is super conversational... 😛

contents/ruby/Rails.md
```diff
@@ -0,0 +1,170 @@
...
+                    users GET    /users(.:format)                                                                         users#index
+                          POST   /users(.:format)                                                                         users#create
+                 new_user GET    /users/new(.:format)                                                                     users#new
+                edit_user GET    /users/:id/edit(.:format)                                                                users#edit
+                     user GET    /users/:id(.:format)                                                                     users#show
+                          PATCH  /users/:id(.:format)                                                                     users#update
+                          PUT    /users/:id(.:format)                                                                     users#update
+                          DELETE /users/:id(.:format)                                                                     users#destroy
+```
```
luyangkenneth, line 63: I just noticed that there's a large chunk of content that's been "missing" all this while, because you'd have to scroll so much to the right to realize that it's there:

Screenshot 2019-04-17 at 1 29 54 AM

You should reduce the number of spaces between the columns so that they show up nicely in the website.

Also, remove the excess whitespace padding to the left of the `Prefix` column.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,170 @@
...
+
+The Rails philosophy includes two major principles, namely:
+1. **Don't Repeat Yourself (DRY)**
+2. **Convention over Configuration** 
+
+### The Rails Way 
+
+One of the core tenets of Rails is "Convention over Configuration", which means that many decisions regarding the structure of your Rails application are made for you in advance, so that there is less effort needed to come up with your own set of best practices. This is perhaps most evident in Rails' decision to follow a Model-View-Controller (MVC) architecture by default.
+
+The structure of an app written in the MVC framework is as follows:
```
luyangkenneth: Replace this with `Here's how a Rails application is structured:`

Because the filenames etc. in the diagram are specific to Rails. Also, you can't really... write an app in the MVC framework. The Rails framework does _follow an MVC architecture_, though.

luyangkenneth (changes requested):
Best to check the netlify deploy while proofreading your article: https://deploy-preview-114--learningresources.netlify.com/contents/ruby/rails.html

contents/ruby/Rails.md
luyangkenneth: Hmmmm

`and there are certain areas where other frameworks might be more suitable.`
->
`and there are a few areas that you should consider when deciding on your web framework of choice.`

contents/ruby/Rails.md
AyushChatto: Er.. he pointed out `combatted` -> `combated` , and as I understand it, they're both correct. Google docs doesn't flag either as an error, and this link says they're alternate forms. 
https://wikidiff.com/combatted/combated

contents/ruby/Rails.md
AyushChatto: Ah dear I just copied that from my terminal. Okay gotcha :)

contents/ruby/Rails.md
AyushChatto: Whoops meant architecture here. Fixed it :)

contents/ruby/Rails.md
luyangkenneth: OH. Sorry my bad.

contents/ruby/Rails.md
```diff
@@ -129,27 +128,27 @@ Rails developers can use ordinal representation to access the same element with
...
 
 The Ruby language has a very passionate and extensive community of developers, and there is a vast array of libraries (called "Gems"), that can be used with your Rails projects. These libraries provide a lot of functionalities, and can be used to add flexibility to Rails projects when required. 
 
 Furthermore, as a result of its popularity, it is very likely that any problems you encounter with Rails will have a solution that you can find on Google or StackOverflow. For any new problems, you are also likely to receive a response very quickly, because the Ruby community prides itself for being welcoming and helpful. 
 
-However, Rails might not be best suited for all use cases, and there are certain areas where other frameworks might be more suitable. For instance:
+## Drawbacks
 
-### Drawbacks
+Rails might not be best suited for all use cases,and there are a few areas that you should consider when deciding on your web framework of choice.
```
luyangkenneth: needs a space between the comma and the `and`

contents/ruby/Rails.md
```diff
@@ -89,17 +88,17 @@ In our diagram, the Controller calls the `all` method in the `User` model, which
...
+After the Controller has fetched all the data necessary and applied whatever transformations are needed, it passes in the requisite fields to its view. The View is what the client sees, and is a collection of `.erb` (**E**mbedded **R**u**B**y) files, which are basically just HTML files with lines of Ruby code embedded in it to modify its appearance and behavior. When the controller is generated using the Rails CLI, a corresponding folder of views is made inside `app/views/`. All the .erb files are named after the controller action and stored in folders named after the controller. Following the same example above, Rails will serve up the `app/views/users/index.html.erb` view to the client when executing the `users#index` controller action, thus concluding the request-response cycle.
```
luyangkenneth: .erb -> `.erb`

contents/ruby/Rails.md
```diff
@@ -52,23 +52,21 @@ resources :users
...
-
-<center>
+In the standard Rails convention, once inside the application directory, if you navigate to `app/controllers/`, you should be able to see all the controllers in your project. Rails will look for the controller with the `<name>_controller.rb`, where `<name>` would be replaced by the result under the `Controller` heading in the list of routes above, in this case, `users`. It will then look for the function name under the `Action` heading mapped to the URI function inside the Controller, and then call it. For example, if someone makes a GET request to /users, Rails will then respond by calling the users#index controller action (refer to the code below).
```
luyangkenneth: users#index -> `users#index`

luyangkenneth: /users -> `/users`

luyangkenneth (changes requested):

contents/ruby/Rails.md
```diff
@@ -0,0 +1,169 @@
...
+```ruby
+class UsersController < ApplicationController
+  def index
+    ..
+  end
+end
+```
+<center>
+`users_controller.rb` <i>with an </i>`index` <i>method</i> 
+</center>
```
luyangkenneth, line 80: What this currently looks like:

Screenshot 2019-04-17 at 12 04 11 PM

Inserting a line break after the `<center>` tag should solve this.

contents/ruby/Rails.md
```diff
@@ -0,0 +1,169 @@
...
+
+#### Controller
+In the standard Rails convention, once inside the application directory, if you navigate to `app/controllers/`, you should be able to see all the controllers in your project. Rails will look for the controller with the `<name>_controller.rb`, where `<name>` would be replaced by the result under the `Controller` heading in the list of routes above, in this case, `users`. It will then look for the function name under the `Action` heading mapped to the URI function inside the Controller, and then call it. For example, if someone makes a GET request to /users, Rails will then respond by calling the users#index controller action (refer to the code below). 
+
+```ruby
+class UsersController < ApplicationController
+  def index
+    ..
```
luyangkenneth: Instead of `..`, maybe have a comment instead to make it valid ruby syntax?

Perhaps something like `# implementation goes here`

luyangkenneth:
Apart from the nits mentioned, I think this is ready for final review 🙂

contents/ruby/Rails.md
luyangkenneth: I don't think you verified this with the preview. Adding the `<br>` doesn't work. What I meant was:

```html
<center>

`users_controller.rb` <i>with an </i>`index` <i>method</i> 

</center>

**Review by luyangkenneth — APPROVED** [review]
LGTM after the styling is fixed 🎉

https://github.com/se-edu/learningresources/pull/114#discussion_r276284015 in case it gets lost in the comment threads

File: contents/ruby/Rails.md [hunk 51]
```diff
@@ -0,0 +1,170 @@
...
+
+In our diagram, the Controller calls the `all` method in the `User` model, which is, by convention in a file called `User.rb`, which is stored along with all other models at `app/models/`. The model then acts as an intermediary between the application and the database, and returns the results of the corresponding query. 
+
+#### View
+
+After the Controller has fetched all the data necessary and applied whatever transformations are needed, it passes in the requisite fields to its view. The View is what the client sees, and is a collection of `.erb` (**E**mbedded **R**u**B**y) files, which are basically just HTML files with lines of Ruby code embedded in it to modify its appearance and behavior. When the controller is generated using the Rails CLI, a corresponding folder of views is made inside `app/views/`. All the `.erb` files are named after the controller action and stored in folders named after the controller. Following the same example above, Rails will serve up the `app/views/users/index.html.erb` view to the client when executing the `users#index` controller action, thus concluding the request-response cycle.
```
damithc: `When the controller is generated using the Rails CLI`: this generating of controllers using a CLI takes the reader by surprise. There was no hint of such a thing before?

contents/ruby/Rails.md
```diff
@@ -0,0 +1,170 @@
...
+
+Authors: [Chattoraj Ayush](https://github.com/AyushChatto)
+
+## Ruby on Rails overview
+
+**Ruby on Rails** is a web application framework written in the Ruby programming language. It was designed with the intention of making programming web applications easier and quicker by reducing the amount of code you write. Rails achieves this by being an "opinionated framework", which means that the developers believe that there is a certain approach - often called "The Rails Way" - that is ideal for building most types of web applications. By adhering to this approach, Rails allows you to focus primarily on the logic, and fills in much of the boilerplate code for you. 
+
+The Rails philosophy includes two major principles, namely:
+1. **Don't Repeat Yourself (DRY)**
+2. **Convention over Configuration**
```
damithc: These two are mentioned as major principles here. But later, only one of them is mentioned. Either give both some more air time or remove these three lines (preferred).

contents/ruby/Rails.md
```diff
@@ -0,0 +1,170 @@
...
+  pageNav: 3
+</frontmatter>
+
+{{ navbar | safe }}
+
+<div class="website-content">
+
+# Introduction to Ruby on Rails
+
+Authors: [Chattoraj Ayush](https://github.com/AyushChatto)
```
damithc: Bold this line and add reviewers below it

damithc (changes requested):
Quite good. Near-LGTM already! 👍 
Just a few minor suggestions

damithc:
Merged and deployed.
Good work @AyushChatto 
Thanks to reviewers too, especially @luyangkenneth

PR #118, opened 2019-04-11
Add Book Chapter: Static Typing with TypeScript

### LINKED ISSUE #117: Static Typing with TypeScript

luyangkenneth:
Fixes #117

Here's the permalink to the preview page: https://deploy-preview-118--learningresources.netlify.com/contents/javascript/staticTypingInJavascript.html

---

Structure:
- What is Static Typing?
- Adding Static Typing to JavaScript
- TypeScript Basics
- Getting Started with TypeScript (Additional Resources)

### DIFF
--- a/contents/javascript/staticTypingWithTypescript.md
+++ b/contents/javascript/staticTypingWithTypescript.md
@@ -26,7 +26,7 @@ To help identify these problems without having to run any code, many developers 
 
 - Static type checking **catches type errors earlier in the development cycle**. These errors can cause downtime in your application if undetected, and they are often expensive to triage and fix. Static type checking provides an automatic way to verify the type safety and correctness of your application during the development stage, ensuring that type errors are eliminated before your code is deployed to production. While the correctness of your application logic can also be determined via test suites and manual testing during development, those are often costlier and less reliable than static type checking when it comes to catching type errors.
 
-- Static type checking can **make developers more productive**. When paired with the appropriate IDE support, it enables features such as auto-completion and offers instant feedback by incrementally rechecking your code as you type.
+- Static type checking can **make developers more productive**. With the additional type information made available to IDEs, features such as auto-completion, code hinting, incremental error checking, and automatic refactoring become more powerful.
 
 - Static type checking **improves collaboration on a large codebase**. Statically typed code allows developers to refactor with greater confidence, knowing that API boundaries are enforced by the type checker. If you change the signature of a method, you can quickly discover the other parts of the codebase that need to be changed, as they would be surfaced as errors by the type checker. Explicitly declared types are also a form of documentation, which makes it easier to understand code written by other developers.
 
@@ -39,7 +39,7 @@ In the JavaScript community, Flow and TypeScript have emerged as the two main op
 
 - **[TypeScript](https://www.typescriptlang.org/)** is a statically typed superset of JavaScript that compiles to plain JavaScript. It is an open-source programming language developed by Microsoft.
 
-While Flow and TypeScript integrate into your development workflow in slightly different ways, they have the same goals and share many similarities in terms of syntax. As the industry seems to be shifting towards TypeScript as the top choice <sup>[source](https://dev.to/nickytonline/is-2019-the-year-of-typescript-18p2)</sup>, in this article I will introduce how TypeScript works and how you can get started with TypeScript.
+While Flow and TypeScript integrate into your development workflow in slightly different ways, they have the same goals and share many similarities in terms of syntax. As the industry seems to be shifting towards TypeScript as the top choice <sup>[source](https://dev.to/nickytonline/is-2019-the-year-of-typescript-18p2)</sup>, this article will introduce how TypeScript works and how you can get started with TypeScript.
 
 
 ## TypeScript Basics
--- a/contents/javascript/staticTypingWithTypescript.md
+++ b/contents/javascript/staticTypingWithTypescript.md
@@ -1,0 +1,54 @@
+<frontmatter>
+  title: Static Typing with TypeScript
+  header: pagetop.md
+  footer: footer.md
+  head: head.md
+  siteNav: mainNav.md
+  pageNav: 3
+</frontmatter>
+
+<div class="website-content">
+
+{{ booktitle | safe }}
+
+# Static Typing with TypeScript
+
+**Author(s): [Lu Yang Kenneth](https://github.com/luyangkenneth)**
+
+Reviewer(s): WIP FILL THIS OUT LATER
+
+
+## What is Static Typing?
+
+**Statically typed languages** like Java, Go, and C++ are able to catch type-related errors at compile time, while in **dynamically typed languages** like JavaScript, such errors are not as easily discoverable because the types of variables are only known at runtime.
+
+To help identify these problems without having to run any code, many developers have adopted technologies like Flow and TypeScript, which enable static type checking for codebases written in JavaScript.
+
+Equivalent type checking tools exist for other dynamically typed languages like Python and Ruby as well, which suggests that this issue is not specific to the JavaScript community. The decision to move towards a statically typed codebase is usually motivated by the following benefits:
+
+- Static type checking **reduces type-related errors during the development process**. These errors can be expensive to fix if they leak into production and cause downtime in your application, especially for teams that value stability over short-term productivity. A type checker provides an automatic way to verify the type safety and correctness of your application _before_ it goes live.
+
+- Static type checking can **make developers more productive**. When paired with the appropriate IDE support, it enables features such as auto-completion and offers instant feedback by incrementally rechecking your code as you type.
+
+- Static type checking **improves collaboration on a large codebase**. Explicitly declared types allow developers to refactor with greater confidence, knowing that API boundaries are enforced by the type checker. If you change a method signature or the shape of an object, you can quickly discover the other parts of the codebase that also need to be changed, as they would be surfaced as errors by the type checker.
+
+
+### What's the difference between Flow and TypeScript?
+
+**[Flow](https://flow.org/)** is a static type checker for JavaScript. It is an open-source tool developed by Facebook.
+
+**[TypeScript](https://www.typescriptlang.org/)** is a statically typed superset of JavaScript that compiles to plain JavaScript. It is an open-source programming language developed by Microsoft.
+
+While Flow and TypeScript integrate into your development workflow in slightly different ways, they have the same goals and share many similarities in terms of syntax. As the industry seems to be shifting towards TypeScript as the top choice <sup>[source](https://dev.to/nickytonline/is-2019-the-year-of-typescript-18p2)</sup>, in this article I will introduce how TypeScript works and how you can get started with TypeScript.
+
+
+## TypeScript Basics
+
+WIP
+
+
+## Getting Started with TypeScript
+
+WIP
+
+</div>
--- a/contents/javascript/staticTypingWithTypescript.md
+++ b/contents/javascript/staticTypingWithTypescript.md
@@ -15,29 +15,29 @@
 
 **Author(s): [Lu Yang Kenneth](https://github.com/luyangkenneth)**
 
-Reviewer(s): WIP FILL THIS OUT LATER
+Reviewer(s): [Marvin Chin](https://github.com/marvinchin)
 
 
 ## What is Static Typing?
 
-**Statically typed languages** like Java, Go, and C++ are able to catch type-related errors at compile time, while in **dynamically typed languages** like JavaScript, such errors are not as easily discoverable because the types of variables are only known at runtime.
+Statically typed languages like Java, Go, and C++ are able to catch type-related errors at compile time. However, in dynamically typed languages like Python, Ruby, and JavaScript, such errors are not as easily discoverable because the types of variables are only known at runtime.
 
-To help identify these problems without having to run any code, many developers have adopted technologies like Flow and TypeScript, which enable static type checking for codebases written in JavaScript.
+To help identify these problems without having to run any code, many developers have adopted technologies to enable static type checking for codebases written in dynamically typed languages. The decision to move towards a statically typed codebase is usually motivated by the following benefits:
 
-Equivalent type checking tools exist for other dynamically typed languages like Python and Ruby as well, which suggests that this issue is not specific to the JavaScript community. The decision to move towards a statically typed codebase is usually motivated by the following benefits:
-
-- Static type checking **reduces type-related errors during the development process**. These errors can be expensive to fix if they leak into production and cause downtime in your application, especially for teams that value stability over short-term productivity. A type checker provides an automatic way to verify the type safety and correctness of your application _before_ it goes live.
+- Static type checking **catches type errors earlier in the development cycle**. These errors can cause downtime in your application if undetected, and they are often expensive to triage and fix. Static type checking provides an automatic way to verify the type safety and correctness of your application during the development stage, ensuring that type errors are eliminated before your code is deployed to production. While the correctness of your application logic can also be determined via test suites and manual testing during development, those are often costlier and less reliable than static type checking when it comes to catching type errors.
 
 - Static type checking can **make developers more productive**. When paired with the appropriate IDE support, it enables features such as auto-completion and offers instant feedback by incrementally rechecking your code as you type.
 
-- Static type checking **improves collaboration on a large codebase**. Explicitly declared types allow developers to refactor with greater confidence, knowing that API boundaries are enforced by the type checker. If you change a method signature or the shape of an object, you can quickly discover the other parts of the codebase that also need to be changed, as they would be surfaced as errors by the type checker.
+- Static type checking **improves collaboration on a large codebase**. Statically typed code allows developers to refactor with greater confidence, knowing that API boundaries are enforced by the type checker. If you change a method signature or the shape of an object, you can quickly discover the other parts of the codebase that need to be changed, as they would be surfaced as errors by the type checker. Explicitly declared types are also a form of documentation, which makes it easier to understand code written by other developers.
 
 
-### What's the difference between Flow and TypeScript?
+## Adding Static Typing to JavaScript
 
-**[Flow](https://flow.org/)** is a static type checker for JavaScript. It is an open-source tool developed by Facebook.
+In the JavaScript community, Flow and TypeScript have emerged as the two main options for enabling static type checking:
 
-**[TypeScript](https://www.typescriptlang.org/)** is a statically typed superset of JavaScript that compiles to plain JavaScript. It is an open-source programming language developed by Microsoft.
+- **[Flow](https://flow.org/)** is a static type checker for JavaScript. It is an open-source tool developed by Facebook.
+
+- **[TypeScript](https://www.typescriptlang.org/)** is a statically typed superset of JavaScript that compiles to plain JavaScript. It is an open-source programming language developed by Microsoft.
 
 While Flow and TypeScript integrate into your development workflow in slightly different ways, they have the same goals and share many similarities in terms of syntax. As the industry seems to be shifting towards TypeScript as the top choice <sup>[source](https://dev.to/nickytonline/is-2019-the-year-of-typescript-18p2)</sup>, in this article I will introduce how TypeScript works and how you can get started with TypeScript.
 
--- a/contents/javascript/staticTypingWithTypescript.md
+++ b/contents/javascript/staticTypingWithTypescript.md
@@ -150,6 +150,31 @@ function sayHello(name: string, favoriteColor?: string) {
 
 ## Getting Started with TypeScript
 
-WIP
+### Installation
+
+The command-line TypeScript compiler can be installed as a Node.js package via `npm`:
+```bash
+npm install -g typescript
+```
+
+After installing TypeScript, you will have access to the `tsc` command for compiling TypeScript files:
+```bash
+tsc my_file.ts
+```
+
+This gets compiled into a `my_file.js` file, which you can execute in a browser or on Node.js.
+
+
+### Additional Resources
+
+- The [TypeScript Playground](https://www.typescriptlang.org/play/) is a convenient way to experiment with different scenarios in TypeScript (such as the examples in this article) without any setup required.
+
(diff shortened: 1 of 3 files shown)


luyangkenneth:
@marvinchin Can you help me to review the Intro section? 😄

contents/javascript/staticTypingWithTypescript.md
```diff
@@ -0,0 +1,54 @@
...
+# Static Typing with TypeScript
+
+**Author(s): [Lu Yang Kenneth](https://github.com/luyangkenneth)**
+
+Reviewer(s): WIP FILL THIS OUT LATER
+
+
+## What is Static Typing?
+
+**Statically typed languages** like Java, Go, and C++ are able to catch type-related errors at compile time, while in **dynamically typed languages** like JavaScript, such errors are not as easily discoverable because the types of variables are only known at runtime.
```
marvinchin: I feel like bolding these phrases don't really add much impact. You might want to consider adding tooltips/popovers to explain what exactly statically or dynamically typed languages are. Perhaps having some code snippets from a statically typed language vs dynamically typed language would help to highlight the difference.

contents/javascript/staticTypingWithTypescript.md
```diff
@@ -0,0 +1,54 @@
...
+
+To help identify these problems without having to run any code, many developers have adopted technologies like Flow and TypeScript, which enable static type checking for codebases written in JavaScript.
+
+Equivalent type checking tools exist for other dynamically typed languages like Python and Ruby as well, which suggests that this issue is not specific to the JavaScript community. The decision to move towards a statically typed codebase is usually motivated by the following benefits:
```
marvinchin: This discussion involving other languages feel a little out of scope for the article. I feel that the motivations for static typing can stand alone without introducing static typing in other languages.

contents/javascript/staticTypingWithTypescript.md
```diff
@@ -0,0 +1,54 @@
...
+
+- Static type checking **reduces type-related errors during the development process**. These errors can be expensive to fix if they leak into production and cause downtime in your application, especially for teams that value stability over short-term productivity. A type checker provides an automatic way to verify the type safety and correctness of your application _before_ it goes live.
+
+- Static type checking can **make developers more productive**. When paired with the appropriate IDE support, it enables features such as auto-completion and offers instant feedback by incrementally rechecking your code as you type.
+
+- Static type checking **improves collaboration on a large codebase**. Explicitly declared types allow developers to refactor with greater confidence, knowing that API boundaries are enforced by the type checker. If you change a method signature or the shape of an object, you can quickly discover the other parts of the codebase that also need to be changed, as they would be surfaced as errors by the type checker.
+
+
+### What's the difference between Flow and TypeScript?
```
marvinchin: I'm not sure if discussing the difference between Flow and TypeScript is appropriate here, especially since there is no context as to what Flow is yet. It may be confusing for someone new to this subject to introduce it so early. Perhaps you could consider placing this at the end instead as a section on alternatives to TypeScript?

contents/javascript/staticTypingWithTypescript.md
```diff
@@ -0,0 +1,54 @@
...
+
+- Static type checking **reduces type-related errors during the development process**. These errors can be expensive to fix if they leak into production and cause downtime in your application, especially for teams that value stability over short-term productivity. A type checker provides an automatic way to verify the type safety and correctness of your application _before_ it goes live.
+
+- Static type checking can **make developers more productive**. When paired with the appropriate IDE support, it enables features such as auto-completion and offers instant feedback by incrementally rechecking your code as you type.
+
+- Static type checking **improves collaboration on a large codebase**. Explicitly declared types allow developers to refactor with greater confidence, knowing that API boundaries are enforced by the type checker. If you change a method signature or the shape of an object, you can quickly discover the other parts of the codebase that also need to be changed, as they would be surfaced as errors by the type checker.
```
marvinchin: I think it's also worth mentioning that it acts as a kind of "contract" between developers - they can know for certain what types of inputs any method takes in, and what it returns, which helps to improve understanding of code and collaboration.

contents/javascript/staticTypingWithTypescript.md
```diff
@@ -0,0 +1,54 @@
...
+
+## What is Static Typing?
+
+**Statically typed languages** like Java, Go, and C++ are able to catch type-related errors at compile time, while in **dynamically typed languages** like JavaScript, such errors are not as easily discoverable because the types of variables are only known at runtime.
+
+To help identify these problems without having to run any code, many developers have adopted technologies like Flow and TypeScript, which enable static type checking for codebases written in JavaScript.
+
+Equivalent type checking tools exist for other dynamically typed languages like Python and Ruby as well, which suggests that this issue is not specific to the JavaScript community. The decision to move towards a statically typed codebase is usually motivated by the following benefits:
+
+- Static type checking **reduces type-related errors during the development process**. These errors can be expensive to fix if they leak into production and cause downtime in your application, especially for teams that value stability over short-term productivity. A type checker provides an automatic way to verify the type safety and correctness of your application _before_ it goes live.
```
marvinchin: I think there is a gap between writing code (where TS catches errors) and going live, so perhaps you might be able to be more specific when discussing when exactly type errors can be eliminated.

Also, correctness can also be determined via tests and manual testing during the development stage as well, but those are much costlier and less reliable than static tools. Perhaps that is something you might want to discuss here as well! 🙂

contents/javascript/staticTypingWithTypescript.md
```diff
@@ -0,0 +1,54 @@
...
+**Author(s): [Lu Yang Kenneth](https://github.com/luyangkenneth)**
+
+Reviewer(s): WIP FILL THIS OUT LATER
+
+
+## What is Static Typing?
+
+**Statically typed languages** like Java, Go, and C++ are able to catch type-related errors at compile time, while in **dynamically typed languages** like JavaScript, such errors are not as easily discoverable because the types of variables are only known at runtime.
+
+To help identify these problems without having to run any code, many developers have adopted technologies like Flow and TypeScript, which enable static type checking for codebases written in JavaScript.
```
marvinchin: I think you can discuss some of the pain points of not being able to automatically or reliably catch type errors here/ There may be some overlap with the points you discuss further below, though!

marvinchin (commented):
I think the introduction does a good job in highlighting what static typing is, and why developers might want to consider typescript. Looking forward to the rest of the article!

contents/javascript/staticTypingWithTypescript.md
luyangkenneth: I think it's a reasonable assumption that readers know what statically / dynamically typed languages are (on a basic level), so I meant for this part to just set the stage for the rest of the introduction. I don't have the intention to go into too much explanation here.

The bolding wasn't meant to have any impact actually haha. It was added as an afterthought because I thought the page might look more aesthetic with some bolded words 😛

contents/javascript/staticTypingWithTypescript.md
luyangkenneth: I agree with your second point. There were a few moments where I was contemplating whether to include something because it's not strictly required but might provide useful context anyway - since these book chapters are for exposure purposes. What dyou think?

contents/javascript/staticTypingWithTypescript.md
luyangkenneth: > especially since there is no context as to what Flow is yet

I did mention it in a paragraph on top: `...technologies like Flow and TypeScript, which enable static type checking for codebases written in JavaScript.`

Hmm I wanted to use this an opportunity to state what TypeScript is upfront (a programming language; not quite the same as a type checking tool) in the same breath as Flow so that there's a super clear distinction about their high-level differences. And then use this also as a segue to `As the industry seems to be shifting towards TypeScript as the top choice, in this article I will introduce how TypeScript works and how you can get started with TypeScript.`

Placing this at the end could of course still work, but that would mean duplicating `TypeScript is a statically typed superset of JavaScript that compiles to plain JavaScript.` again after the article has already gone through TypeScript Basics.

---

Maybe what I'll do is to restructure this to:
- What is Static Typing? (no mention of Flow / TypeScript; all discussion here is generic)
- Adding Static Typing to JavaScript ("There are two popular ways to do this! Here are their differences in a few sentences! Now I will talk more about TypeScript!")

^ Does this sound better to you?

contents/javascript/staticTypingWithTypescript.md
luyangkenneth: Nice

contents/javascript/staticTypingWithTypescript.md
luyangkenneth: Oooh yeah good to make those other related things explicit

contents/javascript/staticTypingWithTypescript.md
luyangkenneth: Yep I'll rework this so that the bullet points can stand on their own

contents/javascript/staticTypingWithTypescript.md
luyangkenneth: (see below comments) I think this will feel less out of place after the rework - stay tuned!

contents/javascript/staticTypingWithTypescript.md
marvinchin: Yes, I think that makes it a little clearer why Flow is being introduced. However, I'm still not sure how valuable introducing Flow may be - you might also just add a note on it in the further reading section. But if you think it's useful, I think the structure you proposed makes sense.

luyangkenneth:
@marvinchin Added some new sections :D

contents/javascript/staticTypingWithTypescript.md
```diff
@@ -0,0 +1,157 @@
...
+**[TypeScript](https://www.typescriptlang.org/)** is a statically typed superset of JavaScript that compiles to plain JavaScript. It is an open-source programming language developed by Microsoft.
+
+While Flow and TypeScript integrate into your development workflow in slightly different ways, they have the same goals and share many similarities in terms of syntax. As the industry seems to be shifting towards TypeScript as the top choice <sup>[source](https://dev.to/nickytonline/is-2019-the-year-of-typescript-18p2)</sup>, in this article I will introduce how TypeScript works and how you can get started with TypeScript.
+
+
+## TypeScript Basics
+
+### Compilation to JavaScript
+
+As a superset of JavaScript, TypeScript is based on the same programming syntax as JavaScript, and valid JavaScript code is also valid TypeScript code. Besides static type checking, TypeScript offers support for the latest JavaScript features, including those from ECMAScript 2015 (ES6) and beyond.
```
marvinchin: > Besides static type checking, TypeScript offers support for the latest JavaScript features, including those from ECMAScript 2015 (ES6) and beyond.

This somehow invokes some kind of comparison between static type checking and compatibility with >ES6 syntax to me. Perhaps consider this?
Besides static type checking, ... -> In addition, ...

contents/javascript/staticTypingWithTypescript.md
```diff
@@ -0,0 +1,157 @@
...
+
+While TypeScript can be configured to target different flavors of JavaScript, it is transpiled down to ES3 by default to maintain compatibility with all JavaScript engines. Because of this, you can use features like classes, generators, and async functions without worrying about whether your code is supported by a particular web browser or version of Node.js.
```
marvinchin: Perhaps you can add a tooltip to explain what transpilation is, since it may be a new concept to some readers!

contents/javascript/staticTypingWithTypescript.md
```diff
@@ -0,0 +1,157 @@
...
+A key feature of TypeScript is the ability to add **type annotations** to your JavaScript variables and functions. This is useful for recording the intended contract of a variable or function.
+
+Here's an example of declaring a variable with the `string` type and assigning it a `string` value:
+```typescript
+let myString: string = "abc";
+```
+
+If we try to assign it a `number` value later on, the TypeScript compiler will raise an error:
+```typescript
+myString = 123; // error: 123 is not assignable to type 'string'
```
marvinchin: Do we want to include a screenshot of the compiler/IDE displaying the error, to make it a little more "realistic" to the reader?

contents/javascript/staticTypingWithTypescript.md
```diff
@@ -0,0 +1,157 @@
...
+const myList = [1, "hello", false];
+printItemsToConsole(myList);
+```
+
+TypeScript also supports other variable types such as arrays and `void`, as seen in the example above. Check out [this article](https://www.typescriptlang.org/docs/handbook/basic-types.html) from the official TypeScript website for more details on those other types.
+
+
+### Custom Types
+
+In JavaScript, values are often passed around in the form of objects. Using TypeScript can help enforce the shape of these objects. Let's use the `type` keyword to define our own `Person` type to describe an object that has the `firstName` and `lastName` properties:
```
marvinchin: Perhaps you can add an explanation for what "shape" means here.

contents/javascript/staticTypingWithTypescript.md
```diff
@@ -0,0 +1,157 @@
...
+// valid
+myBoolean = false;
+
+// error: "hello world" is not assignable to type 'boolean'
+myBoolean = "hello world";
+```
+
+Sometimes, we don't know what the type of a variable might be. In those cases, we would use the `any` type to opt out of compile-time type checking. For example, `myList` is an array that contains elements with a mix of different types:
+```typescript
+function printItemsToConsole(list: any[]): void {
```
marvinchin: would the method name `logItems` be shorter? 😛

contents/javascript/staticTypingWithTypescript.md
```diff
@@ -0,0 +1,157 @@
...
+**[Flow](https://flow.org/)** is a static type checker for JavaScript. It is an open-source tool developed by Facebook.
+
+**[TypeScript](https://www.typescriptlang.org/)** is a statically typed superset of JavaScript that compiles to plain JavaScript. It is an open-source programming language developed by Microsoft.
+
+While Flow and TypeScript integrate into your development workflow in slightly different ways, they have the same goals and share many similarities in terms of syntax. As the industry seems to be shifting towards TypeScript as the top choice <sup>[source](https://dev.to/nickytonline/is-2019-the-year-of-typescript-18p2)</sup>, in this article I will introduce how TypeScript works and how you can get started with TypeScript.
+
+
+## TypeScript Basics
+
+### Compilation to JavaScript
```
marvinchin: Just some thoughts - I wonder how important this is for a reader who wants to know what typescript is. while it is important, I feel it may fall outside the "core" functionality of typescript, which is type checking, and hence it may make more sense to place it after the section on type checking.

contents/javascript/staticTypingWithTypescript.md
```diff
@@ -0,0 +1,157 @@
...
+
+As a superset of JavaScript, TypeScript is based on the same programming syntax as JavaScript, and valid JavaScript code is also valid TypeScript code. Besides static type checking, TypeScript offers support for the latest JavaScript features, including those from ECMAScript 2015 (ES6) and beyond.
+
+While TypeScript can be configured to target different flavors of JavaScript, it is transpiled down to ES3 by default to maintain compatibility with all JavaScript engines. Because of this, you can use features like classes, generators, and async functions without worrying about whether your code is supported by a particular web browser or version of Node.js.
+
+
+### Type Annotations and Type Inference
+
+A key feature of TypeScript is the ability to add **type annotations** to your JavaScript variables and functions. This is useful for recording the intended contract of a variable or function.
```
marvinchin: Intended contract feels a little informal and possibly unspecific here. Would "expected type" be more appropriate here?

marvinchin (changes requested):
Good additions to show the reader how using typescript looks like, and highlights the key functionalities of typescript 🙂 Just have a couple of questions regarding the organisation and phrasing of the article.

contents/javascript/staticTypingWithTypescript.md
luyangkenneth: I intended this to be a slightly more detailed introduction to what TypeScript's relationship with JavaScript actually is, which I think is important to understand from the beginning - I can imagine lacking clarity about this as a new reader. Putting it here is also to extend on the `statically typed superset of JavaScript that compiles to plain JavaScript` one-liner in the previous section.

contents/javascript/staticTypingWithTypescript.md
luyangkenneth: Because these examples are focusing on the types - but we still need some kind of implementation to help contextualize the example - I wanted to have implementation that was as simple as possible, to not distract the reader from the main focus. So I think I want to stick to a method name that is glaringly explicit (as opposed to abstract, which I might favor in a real-world codebase).

contents/javascript/staticTypingWithTypescript.md
luyangkenneth: It's kinda like providing that little bit of theoretical knowledge before diving into "how do we actually use this" for the rest of the article.

contents/javascript/staticTypingWithTypescript.md
luyangkenneth: Haha I definitely considered this. There a few reasons why I decided against it eventually.

1) Readers won't be able to copy-paste the code snippets if they wanna play with the examples themselves - I'm going to introduce https://www.typescriptlang.org/play/ in the last section.

2) If I hover over the error, it will block the rest of the code. Of course, I can work around this by adding more line breaks above the errored line, but it'll be uglier. I also want to simplify the error messages, so using comments would be better than letting readers see the actual IDE errors. This is especially true in the case of objects.

Screenshot 2019-04-20 at 1 50 31 PM

3) Screenshots are much harder to update later on if someone else wants make any changes / additions to the article. Also, a new screenshot might have different syntax highlighting, different font, etc. which makes the article look less consistent overall.

contents/javascript/staticTypingWithTypescript.md
luyangkenneth: Good point 👍

luyangkenneth:
@marvinchin Just made some updates - if these are ok, I just have my last section remaining! (It's on the way)

marvinchin (commented):
It looks good to me so far!

luyangkenneth:
Ready for final review!

contents/javascript/staticTypingWithTypescript.md
```diff
@@ -0,0 +1,180 @@
...
+
+To help identify these problems without having to run any code, many developers have adopted technologies to enable static type checking for codebases written in dynamically typed languages. The decision to move towards a statically typed codebase is usually motivated by the following benefits:
+
+- Static type checking **catches type errors earlier in the development cycle**. These errors can cause downtime in your application if undetected, and they are often expensive to triage and fix. Static type checking provides an automatic way to verify the type safety and correctness of your application during the development stage, ensuring that type errors are eliminated before your code is deployed to production. While the correctness of your application logic can also be determined via test suites and manual testing during development, those are often costlier and less reliable than static type checking when it comes to catching type errors.
+
+- Static type checking can **make developers more productive**. When paired with the appropriate IDE support, it enables features such as auto-completion and offers instant feedback by incrementally rechecking your code as you type.
```
damithc: `auto completion of type information and incrementally rechecking for type errors as you type`?

damithc: Anyway, how does static typing help here? IDEs have auto-completion and incremental rechecking even for dynamic languages.

contents/javascript/staticTypingWithTypescript.md
```diff
@@ -0,0 +1,180 @@
...
+
+- Static type checking **improves collaboration on a large codebase**. Statically typed code allows developers to refactor with greater confidence, knowing that API boundaries are enforced by the type checker. If you change the signature of a method, you can quickly discover the other parts of the codebase that need to be changed, as they would be surfaced as errors by the type checker. Explicitly declared types are also a form of documentation, which makes it easier to understand code written by other developers.
+
```
damithc: I guess the point here is with the addition of type information, IDE features such as auto-completion, error checking, and refactoring can be more powerful because the IDE can do more automation with the help of the additional information available in the code.

contents/javascript/staticTypingWithTypescript.md
```diff
@@ -0,0 +1,180 @@
...
+## Adding Static Typing to JavaScript
+
+In the JavaScript community, Flow and TypeScript have emerged as the two main options for enabling static type checking:
+
+- **[Flow](https://flow.org/)** is a static type checker for JavaScript. It is an open-source tool developed by Facebook.
+
+- **[TypeScript](https://www.typescriptlang.org/)** is a statically typed superset of JavaScript that compiles to plain JavaScript. It is an open-source programming language developed by Microsoft.
+
+While Flow and TypeScript integrate into your development workflow in slightly different ways, they have the same goals and share many similarities in terms of syntax. As the industry seems to be shifting towards TypeScript as the top choice <sup>[source](https://dev.to/nickytonline/is-2019-the-year-of-typescript-18p2)</sup>, in this article I will introduce how TypeScript works and how you can get started with TypeScript.
```
damithc: `in this article I will introduce` -> `this article introduces` (avoid _I_, because more authors can join later)

contents/javascript/staticTypingWithTypescript.md
```diff
@@ -0,0 +1,180 @@
...
+  head: head.md
+  siteNav: mainNav.md
+  pageNav: 3
+</frontmatter>
+
+<div class="website-content">
+
+{{ booktitle | safe }}
+
+# Static Typing with TypeScript
```
damithc: ATM it is not clear who the target audience is.
How about we make this `Static Typing in JavaScript` and add a couple of lines about Flow? That way, the target audience is clearly JS programmers who are looking to add static typing to their code. Also, it will be similar to the article on JavaScript modules.

damithc (changes requested):
Mostly good 👍  Added a few comments.

contents/javascript/staticTypingWithTypescript.md
luyangkenneth: Yeah exactly - there is only that much an IDE can infer from code written in a dynamic language.

contents/javascript/staticTypingWithTypescript.md
luyangkenneth: I'm not sure about adding more lines about Flow - almost all of the basics in Flow and TypeScript are similar so there's a lot of overlap there, unlike in the JavaScript modules article.

I'm happy to change the title to `Static Typing in JavaScript` - what do you think?

damithc: If you think the suggested heading fits the current content without adding more on Flow, go ahead.

luyangkenneth: Ok! Done 🙂

damithc:
Merged and deployed.
Good work @luyangkenneth 
Thanks to the reviewer too.

PR #188, opened 2020-04-05
Update Python document to reflect that Python 2 is discontinued

### LINKED ISSUE #165: Python document may need update since Python 2 has reached end-of-life
As seen in this [announcement](https://www.python.org/doc/sunset-python-2/) from the Python Software Foundation, Python 2 reached the end of its life on 1 January 2020.

Here is an excerpt from the announcement:
> As of January 1st, 2020 no new bug reports, fixes, or changes will be made to Python 2, and Python 2 is no longer supported.

In the textbook, the Python document contains a section titled "Python 2 vs Python 3".
Since Python 2 has reached end of life, this section may need an update.

Specifically, we could:
- Remove the comparison between Python 2 and Python 3 (it is not very relevant *for beginners* as Python 3 should now be the default choice for new learners)
- Mention that Python 2 has reached end of life, for completeness

anubh-v:
Fixes #165

### DIFF
--- a/contents/python/introduction-to-python.md
+++ b/contents/python/introduction-to-python.md
@@ -18,7 +18,6 @@ Authors: [Samson Tan Min Rong](https://www.linkedin.com/in/samsontmr/), [Phang C
 <box id="article-toc">
 
 * [Getting started with Python‎](#getting-started-with-python)
-* [Python 2 vs Python 3‎](#python-2-vs-python-3)
 * [Virtual Environment‎](#virtual-environment)
 * [Common Data Structures‎](#common-data-structures)
 * [Indexing and Slicing‎](#indexing-and-slicing)
@@ -27,6 +26,7 @@ Authors: [Samson Tan Min Rong](https://www.linkedin.com/in/samsontmr/), [Phang C
 * [Python for Data Science‎](#python-for-data-science)
   * [Tutorials‎](#tutorials)
 * [Gotchas‎](#gotchas)
+* [A note on Python 2](#a-note-on-python-2)
 </box>
 
 Python is a simple yet powerful and versatile language. Conceived in the late 80s, it is now widely used across many fields of computer science and software engineering. While not as speedy as compiled languages like C or Java, Python's emphasis on readability, and resulting ease of maintenance, often outweighs the advantages conferred by compiled languages. This especially true in applications where execution speed is non-critical.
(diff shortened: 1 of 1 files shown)


iskandarzulkarnaien (changes requested):
Am not very keen on removing the section comparing Python 2 and Python 3. It seems to be quite useful and points the reader to helpful sources.

I'm not very familiar with Python myself but I would think that at some point, a Python programmer will have to deal with older systems built on Python 2, so I personally feel that that section is still quite relevant.

Perhaps you could shift your addition to the top of the Python 2 vs Python 3 section? Its good to mention that it has been officially discontinued. @damithc what is your opinion on this?

Also: slight nit, but could you change this:
>
> it has been 10 years since Python 3 debuted

to this:
>
> it has been more than a decade since Python 3 debuted

otherwise the number will keep getting more and more inaccurate as the years pass

damithc:
> Perhaps you could shift your addition to the top of the Python 2 vs Python 3 section? Its good to mention that it has been officially discontinued. @damithc what is your opinion on this?

It is true that one still does come across Python 2 examples code when searching for Python solutions. The `print 'x'` is a classic give away. I don't mind keeping a bit of info about the difference between the two versions, if not the full thing. We can always point to another resource for a more an in-depth treatment. Also, may be it can be at the bottom of the article rather than in the middle?

Anyway, not totally opposed to removing it either. In that case, add a link to a resource dealing with the difference.

anubh-v:
@iskandarzulkarnaien , thanks for the feedback
You have a point that an awareness of the changes between Python 2 and 3 is still helpful.
Perhaps I can keep the useful links from the section `Python 2 vs Python 3` ?

I do feel that this section should be at the bottom, though.

anubh-v:
Based on the feedback, I will still remove the section `Python 2 vs Python 3`, but I will keep some of its useful information in the section `A note on Python 2` (such as links to resources that explain the difference)

iskandarzulkarnaien:
> 
> 
> Based on the feedback, I will still remove the section `Python 2 vs Python 3`, but I will keep some of its useful information in the section `A note on Python 2` (such as links to resources that explain the difference)

Sounds good to me! The important thing is that we at least briefly touch upon and we point readers to other resources that will help them understand the differences

contents/python/introduction-to-python.md
```diff
@@ -116,4 +108,11 @@ Learning how to avoid such pitfalls is one huge step towards being a productive
 
 * [Top 10 Common Mistakes of Python Programmers](https://www.toptal.com/python/top-10-mistakes-that-python-programmers-make)
 * [Common Python Gotchas](https://sopython.com/wiki/Common_Gotchas_In_Python)
+
+## A Note on Python 2
+When reading about Python online, you may encounter references to *Python 2*. Python 2 is an older, **discontinued** version of the Python language.
+The current major version of Python, Python 3, is backwards incompatible with Python 2. Read this [article](https://www.geeksforgeeks.org/important-differences-between-python-2-x-and-python-3-x-with-examples/) to learn about some of the differences between Python 2 and 3.
```
iskandarzulkarnaien:
```suggestion
The current major version of Python, Python 3, is backwards incompatible with Python 2. You may wish to read this [article](https://www.geeksforgeeks.org/important-differences-between-python-2-x-and-python-3-x-with-examples/) to learn more about some of the differences between Python 2 and 3.
```
Just a small nit, I think "Read this" sounds too much like an instruction rather than a suggestion. Feel free to change it to something other than "You may wish to read", that's the best phrasing I can think of right now.

iskandarzulkarnaien (changes requested):

contents/python/introduction-to-python.md
crphang: backward incompatible

contents/python/introduction-to-python.md
```diff
@@ -116,4 +108,11 @@ Learning how to avoid such pitfalls is one huge step towards being a productive
...
+
+The Python Software Foundation has [stopped maintaining Python 2](https://www.python.org/doc/sunset-python-2/), which means new projects are unlikely to use it.
```
crphang, line 116: unlikely to use it *going forward*

I think this makes it more complete. There are still many systems using Python 2 that are yet to be migrated.

crphang (changes requested):
LGTM in general. Minor nits.

contents/python/introduction-to-python.md
anubh-v: Thanks, yes "You may wish to read" sounds better to me. I have changed it.

However, for your second suggestion: I think the sentence should still be "to learn about some of the differences", rather than "to learn **more** about some of the differences".
Since we are not covering anything about the differences in this article, we should not use the phrase "to learn more about" ?

anubh-v: Thanks, I have changed this

contents/python/introduction-to-python.md
anubh-v: Thanks very much for reviewing this PR.
I feel that the meaning is clear, considering that the sentence refers to `new projects`.
What do you think?

contents/python/introduction-to-python.md
```diff
@@ -116,4 +108,11 @@ Learning how to avoid such pitfalls is one huge step towards being a productive
 
 * [Top 10 Common Mistakes of Python Programmers](https://www.toptal.com/python/top-10-mistakes-that-python-programmers-make)
 * [Common Python Gotchas](https://sopython.com/wiki/Common_Gotchas_In_Python)
+
+## A Note on Python 2
```
Tejas2805: Sometimes readers don't read all the sections and just focus on sections they want to read. Better to name it - "Python 2 discontinued" or something along those lines? So that even if the reader doesn't read the section, by seeing the heading, he at least kind of gets an idea.

Tejas2805 (commented):
Left a comment. Everything else, LGTM!

contents/python/introduction-to-python.md
anubh-v: That's a good suggestion. Will change

anubh-v: I've changed the title to `Python 2: a Discontinued Version`

anubh-v: @Tejas2805, thanks a lot for taking the time to review this. 
Do you think the new title is suitable?

anubh-v:
@crphang @iskandarzulkarnaien thanks very much for taking the time to review this.
I've responded to your comments. Do you think any further changes are needed?

crphang (approved):
LGTM. You might be interested to pick this one up as well: https://github.com/se-edu/learningresources/issues/192

iskandarzulkarnaien (approved):
LGTM! Good work @anubh-v !

PR #195, opened 2020-05-09
Add C# Asynchronous Programming Async/Await Chapter

### LINKED ISSUE #194: Asynchronous Programming in C# using async/await
Proposed outline:
- Introduction
  - What is asynchronous programming?
- Simple example of async/await + Basic Syntax
- Control flow of async/await
- Exception Handling
- Disadvantages of async/await
- Resources
  - Further Reading

xantho09:
Closes #194


damithc (approved):
Structure seems to be fine. Did not spot any content problems when I did quick read. Let's wait one more day to see if there are any other reviews.

contents/csharp/CSAsyncAwait.md
```diff
@@ -0,0 +1,322 @@
...
+    * [Potential Pitfall: Deadlock](#potential-pitfall-deadlock)
+    * [Additional Overhead](#additional-overhead)
+* [Resources](#resources)
+* [Further Reading](#further-reading)
+
+</box>
+
+## What Is Asynchronous Programming?
+
+In **synchronous programming**, code is executed one step at a time, waiting for each step to be finished before proceeding to the next step. For example, if the code looks something like this:
```
j-lum, line 38: Might want to clarify whether the steps are dependent on each other/need to be return their results in a specified order for this example.

Might also want to clarify that async can result in steps finishing in a different order.

contents/csharp/CSAsyncAwait.md
```diff
@@ -0,0 +1,322 @@
...
+In general, this would be fine. However, if Task 1 happens to take a long time to finish, then a lot of time is spent in the execution simply waiting. We can make better use of our time by executing the code like this:
+
+1. Start running Task 1.
+2. Start running Task 2 _(While waiting for Task 1 to finish)_.
+3. Start running Task 3 _(While waiting for Task 1 to finish)_.
+4. Wait for Task 1 to finish.
+5. Use the result from Task 1 to start Task 4.
+6. Wait for Tasks 2, 3 and 4 to finish.
+
+This is the main idea behind asynchronous programming: while waiting for tasks to complete in the background, perform other tasks instead of just waiting there and doing nothing. This can result in better utilization of resources at hand.
```
j-lum: Well technically the current task is being executed. Is there a layman's way to explain execution contexts or just mention threading?

contents/csharp/CSAsyncAwait.md
```diff
@@ -0,0 +1,322 @@
...
+
+</box>
+
+## Disadvantages Of Async/Await
+
+While the use of `async` and `await` can provide the wonderful benefits of responsiveness while keeping the code readable, there are some disadvantages that come with their use.
+
+### Potential Pitfall: Deadlock
+
+It takes some time to understand the correct usage of `async` and `await`. If you're not careful, you could end up with a deadlock situation. Consider the following example:
```
j-lum: I kind of got the idea after reading the article provided. Do you think having a similar diagram from the control flow section could help illustrate the idea here?

j-lum (commented):
Content-wise looks mostly OK to me. My background in async is mainly from scala so I can't really comment too much on the C# code but they look simple and sufficient.

contents/csharp/CSAsyncAwait.md
xantho09: Admittedly, the introduction was the hardest part to write. It was tricky to keep the explanation simple while making sure it gets the idea across.

I think mentioning threading is probably the simplest solution. Something along the lines of:
- You have 3 additional threads, A, B and C.
- Start Task 1 on Thread A.
- Start Task 2 on Thread B.
And so on.

contents/csharp/CSAsyncAwait.md
xantho09: That might be a good idea.

I think I should probably briefly mention (possibly in the `async` section)
- Why the completed task needs to wait for the UI thread specifically (and not just any thread in the thread pool)
- How the execution resumes _at the first opportunity_ the thread is available. (Since that availability never comes, the execution can't resume)

contents/csharp/CSAsyncAwait.md
xantho09: How does it look now?

I'm not too satisfied with the sentence that follows _"This is the main idea behind asynchronous programming"_; I feel like there's a better way to phrase it.

xantho09:
@damithc It doesn't seem like @j-lum is responding to the comments he left, but I've addressed the points he made. Reviewers' names have been added. 

It should be ready for merging.

damithc:
@xantho09 Will merge soon. But first, can you remove my name from the reviewers list, to be consistent with other pages?

xantho09:
Done.

Please and thank you.

